PR #44, opened 2022-01-29
NV6145, NV6249 and NV6038: pathWalker updates.

jayhuang-suse:
(1) To prevent pathWalkers from running into an indefinite state. 

(2) For debugging purpose, the enforcer will pass the container identifier to the pathWalker, too.


jayhuang-suse:
Added:
NV6145: Ability to disable specific functionality to reduce resource utilization

New environmental variable to disable the secret scans on container: "ENF_NO_SECRET_SCANS=1"

jayhuang-suse:
NV6145: Add more debug parameters to enforcer side support log.

When the env variable "ENF_MONITOR_TRACE=1" is set, the extra pathWalker debug traces will be included in the enforcer logs.

PR #104, opened 2022-03-07
NV6172: a new group is unable to create after rolling update on oc49 or rke2 setup

jayhuang-suse:
The rolling update procedure will have a KV data fetching from the old setup, which had a bug about the group name generations. Thus, the workloads with the wrong group names will be accepted by the new controllers using the new group naming method. 

To prevent the miss-handling cases, the profile data of the obsolete groups will be moved into the new group. Then, the obsolete (profile) groups will be deleted.

This patch will be limited to a namespace: "kube-system". The observed cases from oc4.9 were all "task" types (short-lived, temporary pods). Thus, their old occurrences will have no change by the patch but the future tasks will be collected into a new common group holder.


becitsthere:
Close the PR for now to evaluate the impact of the issue.

PR #120, opened 2022-03-15
OpenShift group info support

jeffhuang4704:
Controller now will get associated group information from login-in OpenShift user.
The OpenShift API is [apis/user.openshift.io/v1/users/~], details please refer to 
https://docs.openshift.com/container-platform/4.6/rest_api/user_and_group_apis/user-user-openshift-io-v1.html

It then fetch corresponding roles for each group role and combine them.


jeffhuang4704:
This is for NVSHAS-5977 - 
support clusterrolebindings or rolebindings with group info in IDP for Openshift

PR #177, opened 2022-04-12
NV6280 and NV6289: zero-drift, unexpected Process.Profile.Violation in the oc setup

jayhuang-suse:
Zero-drift baseline profile: k8s probe enhancement:

(1) Consolidate duplicate Probe commands and allow multiple different probe command sections together.

(2) Improve shell script probe detection.

(3) Allow system pods with all privileges to use the binaries in the node.

(4) Fix a wrong detection on the pod's upper folders.

Tested on the oc 4.8 clusters.


chunyenlin (approved):
Controller part changes are good

PR #290, opened 2022-06-14
NVSHAS-6422: fixed containsAny admission control operation ignoring multiple values for same key

alopez-suse:
Previously the "containsAny" operation in admission control would incorrectly resolve to false when given a criteria that contained multiple key/value pairs (formatted as "key=value"). It was due to the map implementation for the lower level functions only allowing one value per key. This change alters that map to allow multiple potential values per key to be checked.

### DIFF
--- a/controller/cache/admission.go
+++ b/controller/cache/admission.go
@@ -773,22 +773,22 @@ func doesMapContain(key, value string, propMap map[string]string) bool {
 	return false
 }
 
-// does propKey=propValue matches kvMap(criteria)
-func doesPropMatchCrt(propKey, propValue string, kvMap map[string]string) bool {
-	if value, exist := kvMap[propKey]; exist {
-		// kvMap contains propKey
-		if value != "" {
-			// does kvMap contain propKey=propValue?
-			return share.EqualMatch(value, propValue)
+func doesPropMatchCrt(propKey, propValue string, kvMap map[string][]string) bool {
+	if values, exist := kvMap[propKey]; exist {
+		for _, value := range values {
+			if value == "" && len(values) == 1 {
+				return true
+			} else if share.EqualMatch(value, propValue) {
+				return true
+			}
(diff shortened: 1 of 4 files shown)


controller/cache/admission.go
```diff
@@ -773,22 +773,22 @@ func doesMapContain(key, value string, propMap map[string]string) bool {
...
-func doesPropMatchCrt(propKey, propValue string, kvMap map[string]string) bool {
-	if value, exist := kvMap[propKey]; exist {
-		// kvMap contains propKey
-		if value != "" {
-			// does kvMap contain propKey=propValue?
-			return share.EqualMatch(value, propValue)
+func doesPropMatchCrt(propKey, propValue string, kvMap map[string][]string) bool {
+	if values, exist := kvMap[propKey]; exist {
+		for _, value := range values {
+			if value == "" && len(values) == 1 {
```
chunyenlin, line 779: Is this line correct?
Let's say: 
propKey="DESCRIPTION"
propValue="iperfserver"
kvMap={"DESCRIPTION": {"", ""}}
This function will return false. 
But is it expected to return false or true in this case?

alopez-suse: During the creation of an admission control rule from the NeuVector web console duplicate key/values get consolidated (for example `"DESCRIPTION,DESCRIPTION"` in the value field will get transformed into `"DESCRIPTION"`, which then gets translated to `kvMap={"DESCRIPTION": {""}}` here).

However, I am unsure whether this behavior exists within the front-end code, or within the controller code, so I must verify.

If it exists solely within the front-end code this function `doesPropMatchCrt` should then be updated to handle multiple empty values per key. If however that duplicate deletion exists within the controller code, this function should never be reached with multiple empty values per key.

chunyenlin: ok

PR #292, opened 2022-06-15
NVSHAS-6422: fixed containsAny admission control operation ignoring multiple values for same key

alopez-suse:
Previously the "containsAny" operation in admission control would incorrectly resolve to false when given a criteria that contained multiple key/value pairs (formatted as "key=value"). It was due to the map implementation for the lower level functions only allowing one value per key. This change alters that map to allow multiple potential values per key to be checked and adds a related test case.

### DIFF
--- a/controller/cache/admission.go
+++ b/controller/cache/admission.go
@@ -773,55 +773,60 @@ func doesMapContain(key, value string, propMap map[string]string) bool {
 	return false
 }
 
-// does propKey=propValue matches kvMap(criteria)
-func doesPropMatchCrt(propKey, propValue string, kvMap map[string]string) bool {
-	if value, exist := kvMap[propKey]; exist {
+
+func doesCrtContainProp(propKey, propValue string, kvMap map[string][]string) bool {
+	if crtValues, exist := kvMap[propKey]; exist {
 		// kvMap contains propKey
-		if value != "" {
-			// does kvMap contain propKey=propValue?
-			return share.EqualMatch(value, propValue)
+		for _, crtValue := range crtValues {
+			// crtValue of "" means user only cares whether propKey exists in criteria
+			// otherwise check that propValue equals crtValue
+			if crtValue == "" || share.EqualMatch(crtValue, propValue) {
+				return true
+			}
 		}
-		return true
 	}
 	return false
 }
 
 func isMapCriterionMet(crt *share.CLUSAdmRuleCriterion, propMap map[string]string) (bool, bool) {
 	if len(propMap) > 0 {
-		kvMap := make(map[string]string, len(crt.ValueSlice)) // the key=value configured in criteria
+		kvMap := make(map[string][]string) // the key=value configured in criteria
 		for _, crtValue := range crt.ValueSlice {
 			var key, value string
 			if i := strings.Index(crtValue, "="); i >= 0 {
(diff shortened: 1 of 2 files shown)


controller/cache/admission.go
```diff
@@ -773,22 +773,21 @@ func doesMapContain(key, value string, propMap map[string]string) bool {
...
 }
 
-// does propKey=propValue matches kvMap(criteria)
-func doesPropMatchCrt(propKey, propValue string, kvMap map[string]string) bool {
-	if value, exist := kvMap[propKey]; exist {
-		// kvMap contains propKey
-		if value != "" {
-			// does kvMap contain propKey=propValue?
-			return share.EqualMatch(value, propValue)
+func doesPropMatchCrt(propKey, propValue string, kvMap map[string][]string) bool {
```
chunyenlin, line 776: Is this function correct?
Let's say:
propKey="DESCRIPTION"
propValue="iperfserver"
kvMap={"DESCRIPTION": {"abc", "iperfserver"}}
This function will return false.
But is it expected to return false or true in this case?

alopez-suse: I see.

This function is only used in the following two places:
https://github.com/alopez-suse/neuvector/blob/372f93db1f2aaeaed73a3321f8441449d0f084a4/controller/cache/admission.go#L825
https://github.com/alopez-suse/neuvector/blob/372f93db1f2aaeaed73a3321f8441449d0f084a4/controller/cache/admission.go#L908

My change mimics what the previous logic was. Previously there could only be one value per key, so each new value would override the last, and so it would have returned false in that case. However, while this isn't introducing any new issues, when considering that this function is meant to represent `CriteriaOpContainsOtherThan` I think returning `true` is the desired behavior in that case.

Essentially this function should only return `false` when the `kvMap` variable doesn't contain anything matching `propKey` or `propValue`.

alopez-suse: Implemented changes to have `CriteriaOpContainsOtherThan` behave as discussed, changed the helper function name to be a bit clearer as to what its doing, and included a relevant test case.

controller/cache/admission.go
```diff
@@ -773,22 +773,28 @@ func doesMapContain(key, value string, propMap map[string]string) bool {
...
 
-// does propKey=propValue matches kvMap(criteria)
-func doesPropMatchCrt(propKey, propValue string, kvMap map[string]string) bool {
-	if value, exist := kvMap[propKey]; exist {
-		// kvMap contains propKey
-		if value != "" {
-			// does kvMap contain propKey=propValue?
-			return share.EqualMatch(value, propValue)
+
+func doesCrtContainProp(propKey, propValue string, kvMap map[string][]string) bool {
```
chunyenlin, line 777: Is this line correct?
Let's say:
propKey="DESCRIPTION"
propValue=""
kvMap={"DESCRIPTION": {"", "abc", "iperfserverb"}}
This function will return false.
But is it expected to return false or true in this case?

alopez-suse: Will the `propValue` argument ever be an empty value? It is sourced from the `nvsysadmission.AdmResObject.Labels` and `nvsysadmission.ScannedImageSummary.Labels` fields.

alopez-suse: After discussion, added behavior to return true for empty propValues when propKey is matched.

PR #332, opened 2022-07-22
Update: Update Go version

b1gb4by:
The version of Go was not updated.


garyduan:
Thanks for submitting this PR. We have tried to upgrade Golang to a newer version, however, we run into this issue with the built-in certificate.

https://jfrog.com/knowledge-base/general-what-should-i-do-if-i-get-an-x509-certificate-relies-on-legacy-common-name-field-error/

Because scanner is released daily, there will be a compatibility issue between the scanner and the existing controllers for the gRPC connection. We are still thinking how to mitigate this problem.

PR #388, opened 2022-08-25
NVSHAS-6654 fix >= and <= operators in module admission control rule

alopez-suse:
The issue was the way I was "finding" the operator.

When a user enters a criterion value such as `myKey<=myExample`, the code previously resolved the operator as `=`. These changes establish hard checks for those possibilities, and may be worth extracting into its own type upon further expansion.


alopez-suse:
New changes also correct some formatting errors from before.

williamlin-suse (approved):

PR #445, opened 2022-10-03
NVSHAS-6703 For 5.0.0, with Antrea CNI, traffic between pods within the same node see source as unmanaged workload

gfsuse:


becitsthere:
Already cherry picked.

PR #449, opened 2022-10-10
NVSHAS-6767: delete bench and scan keys after a pod is deleted and rolling upgrade

jayhuang-suse:
The rolling upgrades might carry over not-exiting workload bench reports and states. This implementation will have a 2-strikes out mechanism to purge those not-existing KV entries with continued "set flag" events. The first check is set at the beginning of the controllers and marks the flag for 30 minutes periodically.

For example, there is a not-existing workload "1234" in which we find a KV key associated with "bench/1234/set_id".

(1) 1st run:  the "1234" is not the current known workload list. It was marked/flagged as a "missing id list" entry: "1234"
      30 minutes later .....
(2) 2nd run: the "1234" is still NOT in the current known workload list. We delete the KV tree associated with the workload id "1234". Otherwise, it exists and we remove it from the "missing id list".

Targeting to: 
 (1) bench reports: bench/<id>/<BenchType>
 (2) bench scan state: scan/state/bench/workload/<id>

It could expand to purge other KV entries, if any, with not-existing working workloads later.

To enhance the performance, I also add the GetKeys and transactional "DeleteTree" capabilities.
 
By the way, it has not guaranteed to tell by the controller, itself,  from a rolling upgrade. Thus, we can not use the one-time purge approach at the beginning of the controllers.


jayhuang-suse:
At controllers: prune invalid workloads from below KV trees.
   "scan/state/bench/<id>"
   "bench/<id>"
   "scan/data/report/workload/<id>"

At enforcers: remove the old workload tree once connect with the cluster. Let the enforcer refill the records of the running workloads.
"object/workload/<host-id>/<id>"

jayhuang-suse:
Add "scan/state/report/workload/\<id\>".

Issue #338, opened 2022-07-26
Neuvector can't scan JFrog container Registry (OpenSource). 


os-learner:
**Environment**
Platform: OpenShift 4.8.39
Kubernetes/Platform Version(s): v1.21.8+ed4d8fd

**Describe the bug**
I have setup our JFrog Container Registry on the OpenShift platform. I have provided credentials to our JFrog Container Registry.
When I click on "Test Setting" and "Start Test", In the step_content "manifests", neuvector reads all the manifests from repository.catalog but when reading each manifest individually failing. 
I'm attaching the partial output of the errors to the issue.

**To Reproduce**
1. Setup JFrog Container Registry in Neuvector.
2. Click on "Test Setting"
3. Click on "Start Test"
4. Scroll down to see the error.

**Expected behavior**
The images should be queued and scanned. 
Instead of looking for manifests in https://<jfrog jcr domain>/v2/<jfrog repository name>/<application name>/manifests/<version>, it should be looking at https://<jfrog jcr domain>/v2/<jfrog repository name>/<application name>/<version>/manifest.json.

**Screenshots**
N/A

**UI Issue**
Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome

**Additional context**
[neuvector-jfrog-scan-error.txt](https://github.com/neuvector/neuvector/files/9191681/neuvector-jfrog-scan-error.txt)

lindhe:
I think I'm seeing the same thing – the tests seem to work but all scans are failing. I'm using Harbor as the registry.

![Screenshot of failing tests](https://user-images.githubusercontent.com/7773090/196407710-76403227-922a-4b23-af19-b64b8925ffef.png)

PR #477, opened 2022-10-25
NVSHAS-6866: controller panic when scanning a fed registry

williamlin-suse:


lindhe:
Could this be related to  #338?

PR #491, opened 2022-11-04
NVSHAS-6910: Manual trigger of Updater job cases NV to create a group for the container.

jayhuang-suse:
The case is duplicated by following similar command:

"kubectl create job --from cronjob/neuvector-updater-pod manual-run-jay.says.do.5 -n neuvector"

The syntax will create a user-defined POD name: "manual-run-jay.says.do.5"
Our current method can not identify that they were one of our pod names. 

(1) enhance the identification method to cover this case.
(2) "cri-o driver": add image labels into container labels.


jayhuang-suse:
Messing up branches. Need to re-submit a PR again.

PR #538, opened 2022-11-21
[Update] update CPU stat calculation by CPU kernal count

youzhiye:
cDelta / sDelta is range between 0 and 100%.
But on a 4-core system, cpu usage can be anywhere between 0 and 400%, so it has to be multiplied by the number of cores.

refer 'docekr stats' :https://github.com/moby/moby/blob/eb131c5383db8cac633919f82abad86c99bffbe5/cli/command/container/stats_helpers.go#L175-L188 
https://github.com/moby/moby/issues/13626

### DIFF
--- a/share/system/stats.go
+++ b/share/system/stats.go
@@ -1,12 +1,13 @@
 package system
 
 import (
-	"time"
-
 	"github.com/neuvector/neuvector/share"
+	"runtime"
+	"time"
 )
 
 func calculateCPU(prevCPU, prevCPUSystem uint64, cpu, cpuSystem uint64) float64 {
+	cpuNum := runtime.NumCPU()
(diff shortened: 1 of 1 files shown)


becitsthere:
The intent is to get the total CPU usage.

youzhiye:
In https://github.com/neuvector/prometheus-exporter, The cpu stat is nv_enforcer_cpu，I think the calculation by CPU kernal count more reasonable, not total CPU usage.
For Operations, they care more about the actual CPU usage.
refer:https://github.com/neuvector/prometheus-exporter/blob/master/nv_exporter.py#L164-L171

becitsthere:
If in the Grafana, it is preferred to show per core cpu usage, it is better to build the widget to show that. In most of use cases, our users care about the total CPU usage.

I will close the PR.

PR #574, opened 2022-12-08
NVSHAS-7054 [agent] Added a line of code to update proc user(name) when it is getting evaluated.

jaimeyu-suse:
The issue was that when you do a `docker exec --user <user> ...` the nv agent would pick up the host' username instead of the container's username (they share the same uid). I figured out that when we're doing the evaluation, that the process' filepaths look correct but getting the wrong user from `/etc/passwd`. But that wasn't true since the path was correct - it was because the username was inherited by the parent by default and never updated again.


jaimeyu-suse:
Will reopen once i figure out unit testing

jayhuang-suse (commented):
A better approach to reduce i-o access. Good.

PR #583, opened 2022-12-16
NVSHAS-7445: show empty object in workload if app container is stopped

jayhuang-suse:
Docker runtime driver: avoid using docker's native sandbox identifier.


becitsthere:
6992 is a low priority one. Fixes for 6992 are reverted, so this is not needed.

PR #587, opened 2022-12-22
NVSHAS-7471 concurrent map read/write fix

jaimeyu-suse:
We hit an error with patchRuntimeUser() with error: fatal error: concurrent map read and map write

So we need to add a lock around it at process.go:1825 to prevent the issue.


jayhuang-suse (commented):
Good.

PR #602, opened 2023-01-09
NVSHAS-7501  Made two changes so we can kill existing running process so they can be killed when we transition to Protect mode.

jaimeyu-suse:
NVSHAS-7501 

One, I changed it so func (p *Probe) evaluateApp() will not set the keep alive flag to true by default. This blanket default will cause a violating process to continue to run even tho the logs will say it was "denied".
Second, I added a check in procProfileEval() so it will ignore the reported flags when we are in Protect mode. I noticed that when we transition from monitor -> protect, we don't track the state change and reset the flags. So even tho the logs say that the process is in violation, we do not attempt to kill it. So my change is to ignore the flags when we are in Protect mode.

The reason is that regardless of the reporting flags, when we're in protect mode, we should always be trying to kill violating processes and not base the decision on the reported flag.

### DIFF
--- a/agent/probe/process.go
+++ b/agent/probe/process.go
@@ -2418,6 +2418,15 @@ func (p *Probe) procProfileEval(id string, proc *procInternal, bKeepAlive bool) 
 				default:
 					pp.Action = share.PolicyActionAllow
 				}
+			} else {
+				// NVSHAS-7501 - I think we have to assume false on keep alive.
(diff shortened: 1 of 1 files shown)


jaimeyu-suse:
For https://jira.suse.com/browse/NVSHAS-7501

becitsthere:
Superseded by #610

PR #606, opened 2023-01-11
NVSHAS-7454 fix forwarded dlp export and lift forwarded user/role restrictions

alopez-suse:


williamlin-suse (changes requested):
transaction id handling for redirected dlp import request is missing

williamlin-suse (approved):

PR #614, opened 2023-01-13
NVSHAS-7532: Can not re-create deleted learned group

jayhuang-suse:
The deleted learned service groups fail to delete the dispatcher entry. Thus, the re-created group can not be formed.


jayhuang-suse:
Updated by PR#629 and PR#631.

PR #621, opened 2023-01-19
Nvshas 7505 - Fixing : Process execution denied on pods even after the specific process is allowed on process profile with ZeroDrift enabled

jaimeyu-suse:
Problem: Process execution denied on pods even after the specific process is allowed on process profile with ZeroDrift enabled

Fix: I went thru the code and it looks like evaluateApp doesn't check the whitelist so it continues to block processes if they're new to the image. (that's why the work-around was to restart the container so the binaries would look like they're not new to the container). I put in a fix to check the whitelist before doing anymore checks on the process -- if its on the whitelist, it should reduce overhead and just skip to allow the process to continue.


jayhuang-suse (commented):
Need some changes. Please follow the slack messages. Thanks.

jaimeyu-suse:
Closing, looks like it is behaving as it should.

PR #622, opened 2023-01-19
https://jira.suse.com/browse/NVSHAS-7534

jaimeyu-suse:
Adding a if-else to disable keep alive if we determine a process is not exempted from being killed during runtime.


becitsthere:
Please correct the commit message and PR title

PR #645, opened 2023-02-06
Remove hardcoded exception for 'ps' process

dqsully:
I found this while looking through the source code. Although there are other security features enabled by default that should protect against any vulnerabilities caused by this hardcoded exception, it's still scary to disable process execution protections for any processes named `ps`.

As far as I can tell, `p.procProfileEval()` is _the_ function which determines if a detected process should be running or not, implements zero-drift process protection, process learning, process incident reports, and even process kills. However, file access protections, which can also kill processes and generate reports, are implemented in a separate file without this hardcoded exception for `ps`.

### DIFF
--- a/agent/probe/process.go
+++ b/agent/probe/process.go
@@ -1839,24 +1839,22 @@ func (p *Probe) evaluateApplication(proc *procInternal, id string, bKeepAlive bo
 	}
 
 	risky := proc.action == share.PolicyActionCheckApp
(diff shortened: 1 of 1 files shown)


becitsthere:
Replaced by PR #695

jaimeyu-suse:
I had to open another PR here https://github.com/neuvector/neuvector/pull/717

PR #717, opened 2023-03-28
Nvshas 7649 - Applying checks on process "ps" to determine exception

jaimeyu-suse:
Note, Replacing https://github.com/neuvector/neuvector/pull/645

The original code gave "ps" a blanket exception which could allow things to fall thru. So in this PR, I closed the hole with checks to make sure the parent is runtime process for exceptions. Docker/kube will sometimes check on a container using `ps` in the background and this would throw false positives incidents.


jayhuang-suse (commented):
ok

PR #744, opened 2023-04-18
Nvshas 7495 - Fixing issue with containers stats not populating

jaimeyu-suse:
I finally isolated the issue to a cgroup v1 legacy code.

We had this in the code for the v2 path
`if pid != 0 {
		cpath = filepath.Join("/proc", strconv.Itoa(pid), "root", cpath)
}`
Which is not necessary for cgroup v2 kubenetes. The /proc/<pid>/cgroup file actually dumps out the correct location and our tokenizer picked it up just fine. We just were adding the process' /proc location but actually should be /sys/fs/cgroup. 

Found issue with it on docker and applied a fix for it as well

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -5,12 +5,12 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"reflect"
 	"runtime"
 	"strings"
 	"sync/atomic"
 	"syscall"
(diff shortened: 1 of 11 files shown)


jaimeyu-suse:
It does not work with docker and I have a fix for it as well.

PR #798, opened 2023-05-06
NVSHAS-7880: unexpected Process.Profile.Violation incidents on ps process

jayhuang-suse:
The is a patch for NVSHAS-7649.


jaimeyu-suse:
Looks good to me.

PR #806, opened 2023-05-11
7894-invalid secretkey format checking

jeffhuang4704:


jeffhuang4704:
close first to check unit test failure

PR #810, opened 2023-05-12
7894 check apikey secret format

jeffhuang4704:


jeffhuang4704:
Per discussion, we will adjust the creation flow of API key.  Close this PR.

PR #848, opened 2023-06-02
NVSHAS-6217: update scan result correctly

alopez-suse:
The verifiers will now be included within the scan result response, and can be seen by the registry scan and admission control.

### DIFF
--- a/controller/scan/interface.go
+++ b/controller/scan/interface.go
@@ -559,6 +559,11 @@ func (m *scanMethod) GetRegistryImageReport(name, id string, vpf scanUtils.VPFIn
 			for i, m := range c.modules {
 				rrpt.Modules[i] = scanUtils.ScanModule2REST(m)
 			}
+
+			rrpt.Verifiers = make([]string, len(c.verifiers))
+			for i, v := range c.verifiers {
+				rrpt.Verifiers[i] = v
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


controller/scan/interface.go
```diff
@@ -358,6 +358,8 @@ func (m *scanMethod) StoreRepoScanResult(result *share.ScanResult) error {
 		ScanResult: *result,
 	}
 
+	report.Verifiers = append(report.Verifiers, "fish")
```
williamlin-suse, line 361: ?

controller/scan/interface.go
```diff
@@ -559,6 +561,12 @@ func (m *scanMethod) GetRegistryImageReport(name, id string, vpf scanUtils.VPFIn
 			for i, m := range c.modules {
 				rrpt.Modules[i] = scanUtils.ScanModule2REST(m)
 			}
+
+			rrpt.Verifiers = make([]string, len(c.verifiers))
+			for i, v := range c.verifiers {
+				rrpt.Verifiers[i] = v
+			}
+			rrpt.Verifiers = append(rrpt.Verifiers, "bird")
```
williamlin-suse, line 569: ?

controller/scan/interface.go
alopez-suse: lol (x1)

controller/scan/interface.go
alopez-suse: lol (x2)

PR #862, opened 2023-06-08
NVSHAS-7984: request correct manifest schema for signature images

alopez-suse:
Some registries (quay.io), instead of returning an error like we expect (for example "Accept header does not support OCI manifests") will return a morphed version of the requested manifest, meant to be compatible with whatever schema headers the request was made with.

This is meant to ensure that requests for signature image manifests always include the correct media type (application/vnd.oci.image.manifest.v1+json).

### DIFF
--- a/share/scan/registry/manifest.go
+++ b/share/scan/registry/manifest.go
@@ -34,7 +34,14 @@ type ManifestInfo struct {
 	Labels         map[string]string
 }
 
-func (r *Registry) ManifestRequest(ctx context.Context, repository, reference string, schema int) (string, []byte, error) {
+type ManifestRequestType int
+
+const (
+	ManifestRequest_Default ManifestRequestType = iota
+	ManifestRequest_CosignSignature
+)
+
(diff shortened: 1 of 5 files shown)


share/scan/registry/manifest.go
```diff
@@ -46,6 +46,11 @@ func (r *Registry) ManifestRequest(ctx context.Context, repository, reference st
 	retry := 0
 	withOCIManifest := false
 	withOCIIndex := false
+
```
williamlin-suse, line 49: Please add a parameter to ManifestRequest(), like ManifestRequest(....., forSigImage bool) so that forSigImage also needs to be true for line 51 to be executer

alopez-suse: Implemented as `ManifestRequestType`.

alopez-suse:
This will require a follow-up PR to scanner.

alopez-suse:
Closing this PR as the same changes are included in the following PR: https://github.com/neuvector/neuvector/pull/869

alopez-suse:
Closing this PR as the same changes are included in the following PR: https://github.com/neuvector/neuvector/pull/869

PR #869, opened 2023-06-09
NVSHAS-7957 & NVSHAS-7984

alopez-suse:
For 7984:
Some registries (quay.io), instead of returning an error like we expect (for example "Accept header does not support OCI manifests") will return a morphed version of the requested manifest, meant to be compatible with whatever schema headers the request was made with.

This is meant to ensure that requests for signature image manifests always include the correct media type (application/vnd.oci.image.manifest.v1+json).

For 7957:
If an image is determined to be a signature image, it is not scanned. For an image to be considered a signature image, all of its layers must be of the media type `application/vnd.dev.cosign.simplesigning.v1+json`, which is the cosign specific mediatype for signatures.

The `QuaySpecialCase` logic is meant to handle the odd issue that arises with how Quay handles manifest requests without the correct OCI headers, instead of returning an error (which is how we usually determine a retry with an OCI header) it returns a v1 compatible manifest instead. Since our scanning logic does not understand what each image is meant to look like (we only know the repo and tag), we cannot determine whether this compatibility response is actually the correct response for a given repo and tag.

In the quay special case (which is triggered when the registry is Quay.io, and the tag ends with `.sig`), we request the manifest as if the image is a signature image, since that will force a v2 manifest to be returned (if it is available). If a v2 manifest is not returned, or if the returned v2 manifest does not only contain layers with the media type `application/vnd.dev.cosign.simplesigning.v1+json`, we will then know the current image is not a signature image.

### DIFF
--- a/share/scan/registry.go
+++ b/share/scan/registry.go
@@ -56,10 +56,10 @@ type SignatureData struct {
 	Payloads map[string]string `json:"Payloads"`
 }
 
-func (rc *RegClient) GetImageInfo(ctx context.Context, name, tag string) (*ImageInfo, share.ScanErrorCode) {
+func (rc *RegClient) GetImageInfo(ctx context.Context, name, tag string, manifestReqType registry.ManifestRequestType) (*ImageInfo, share.ScanErrorCode) {
 	var imageInfo ImageInfo
 
-	dg, body, err := rc.ManifestRequest(ctx, name, tag, 2)
+	dg, body, err := rc.ManifestRequest(ctx, name, tag, 2, manifestReqType)
 	if err == nil {
 		// check if response is manifest list
 		var ml manifestList.DeserializedManifestList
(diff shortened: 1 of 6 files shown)


share/scan/registry.go
```diff
@@ -56,11 +62,75 @@ type SignatureData struct {
 	Payloads map[string]string `json:"Payloads"`
 }
 
-func (rc *RegClient) GetImageInfo(ctx context.Context, name, tag string) (*ImageInfo, share.ScanErrorCode) {
+func IsPotentialCosignSignatureTag(tag string) bool {
+	if len(tag) < len(CosignSignatureTagSuffix) {
+		return false
+	}
+	last4CharactersOfTag := tag[len(tag)-4:]
+	return last4CharactersOfTag == CosignSignatureTagSuffix
```
williamlin-suse, line 70: should we also check prefix "sha256-" or "sha" for the tag?

PR #870, opened 2023-06-09
NVSHAS-6217: update signature verification results proto

alopez-suse:


alopez-suse:
Will require a follow-up PR to scanner.

share/scan.proto
```diff
@@ -133,6 +133,8 @@ message ScanResult {
     ScanProvider Provider = 23;
     int64 Size = 24;
     repeated string Verifiers = 25;
+    ScanErrorCode SignatureVerificationError = 26;
```
williamlin-suse, line 136: Can we change lin 135~137 to something like
message SigstoreResult {
    repeated string Verifiers = 1;
    ScanErrorCode SignatureVerificationError = 2;
    string SignatureVerificationTimestamp = 3;
}

message ScanResult {
    ......
    int64 Size = 24;
    SigstoreResult SigResult = 25;
}

alopez-suse:
merging from different branch

PR #874, opened 2023-06-13
NVSHAS-8038: CustomCheck is not working for pods on kubernetes cluster

jayhuang-suse:
The service name was not constructed correctly that causes the mismatched service name and failure to select workload targets.


jayhuang-suse:
Unified the input parameters of the createWorkload(). Its svc name must be the full-service name, like "iperf.namespace_name".

PR #882, opened 2023-06-16
NVSHAS-8045: group signature verification grpc fields in new message

alopez-suse:
Signature related fields are in their own message. Other internal objects have those new fields included following whatever pattern they had before.

Error field only implemented in the RPC message since our logic for breaking out/separating signature verification errors from other errors has not been designed/discussed.


alopez-suse:
Requires follow-up scanner PR.

controller/api/apis.go
```diff
@@ -2165,7 +2165,12 @@ type RESTScanReport struct {
 	Envs      []string             `json:"envs,omitempty"`
 	Labels    map[string]string    `json:"labels,omitempty"`
 	Cmds      []string             `json:"cmds,omitempty"`
-	Verifiers []string             `json:"verifiers,omitempty"`
+	SignatureInfo *RESTScanSignatureInfo `json:"signature_data"`
+}
+
+type RESTScanSignatureInfo struct {
```
williamlin-suse, line 2171: json tag for fields in this type is missing

controller/api/apis.go
```diff
@@ -2165,7 +2165,12 @@ type RESTScanReport struct {
 	Envs      []string             `json:"envs,omitempty"`
 	Labels    map[string]string    `json:"labels,omitempty"`
 	Cmds      []string             `json:"cmds,omitempty"`
-	Verifiers []string             `json:"verifiers,omitempty"`
+	SignatureInfo *RESTScanSignatureInfo `json:"signature_data"`
```
williamlin-suse, line 2168: it should be
SignatureInfo *RESTScanSignatureInfo json:"signature_data,omitempty"

PR #888, opened 2023-06-21
NVSHAS-7418 - Changed getUserName() so it updates the container's username list

jaimeyu-suse:
Previously, there is an edge case where deleting and reusing uids will cause nv to not update its cache.

If you create a user and then execute an violation. The user map will be updated. 
Delete the user and them add another user and reuse the deleted user's uid. Before the fix, the name would show the old one and not the new one.
Now cause a violation and the username will be correct.

### DIFF
--- a/agent/probe/process.go
+++ b/agent/probe/process.go
@@ -1394,6 +1394,21 @@ func (p *Probe) handleProcUIDChange(pid, ruid, euid int) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


jayhuang-suse (changes requested):
please consider updating the user map when it hits the violations as we discussed

jayhuang-suse (commented):
ok

jayhuang-suse (approved):

PR #898, opened 2023-06-28
NVSHAS-7990 Adding catches when retrieving stats throw errors. 

jaimeyu-suse:
On some setups, pod holders don't populate the cgroup directory which causes nv to throw a lot of error log entries. This patch should silence the errors coming from pod holders which was flooding the logs because the stats gatherer runs on a timer. 

Otherwise, the code behaves as before and stats are gathered in best effort.


jayhuang-suse (commented):
ok

becitsthere:
Consider to fix at the root cause. The logic should not try to read the pod's stats.

PR #902, opened 2023-06-29
NVSHAS-8108 - Adding locks because getUpdatedUsername 

jaimeyu-suse:
Found a bug in pipeline tests that found that this was updating a map and sendProcessIncident is used as a go routine. This caused a concurrent map write and the solution is to add a lock to make it safe


jayhuang-suse (commented):
Good

PR #905, opened 2023-07-04
NVSHAS-7990 - Disabling stats gathering for pod holders

jaimeyu-suse:
The change here was to check if the container's parent namespace is "" which would mean it is a pod holder. If so, we should skip stats gathering altogether to save resources.

May also solve NVSHAS-7925 because we should not be reporting values for pod holder.


becitsthere:
Consider to fix at the root cause. The logic should not try to read the pod's stats.

PR #934, opened 2023-07-28
NVSHAS-7554 - Added a catch to block identical filter paths from bein…

jaimeyu-suse:
…g stored through the REST API.

The REST api will make sure the directory path for filters do not have any collisions. The bug we hit was that if a customer was allowed to add both `/tmp/` and `/tmp` but only one filter would be used. This caused confusion in the expected behaviour.
If the filter has a trailing /, it is meant to be a filter for directories. If it is ommited, it is meant to be a filter for filepaths. When the two are used in conjection, it is confusing because we can only store one filter and the behaviour may not be what the user expects.

Added comments to parseFileFilter so its clear what are the behaviours if you add a trailing /.
I cleaned up getCoreFile() because it had duplicated code.

### DIFF
--- a/controller/rest/file_monitor.go
+++ b/controller/rest/file_monitor.go
@@ -28,17 +28,28 @@ var fileAccessOptionList []string = []string{
 }
 var fileAccessOptionSet utils.Set = utils.NewSetFromSliceKind(fileAccessOptionList)
 
+// parseFileFilter - takes a filter name and  returns the directory, the regex expression, and bool if successful
+// It is also important to note that this function returns:
+// * the directory path the filter is to operate on
+// * the regular expression to append the path (eg, adding .*)
+// * boolean to signify if an error occurred parsing the filter.
 // only support simple wildcard
 // 1. /dir/xxx
 // 2. /dir/xxx.*
 // 3. /dir/*.xxx
 // 4. /dir/*/abc/*
 // not support [] () regexp
+// NVSHAS-7554 notes - If a trailing `/` is found in the path, then we will treat the filter to work on _directories_.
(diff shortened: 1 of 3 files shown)


controller/rest/file_monitor.go
```diff
@@ -227,7 +238,7 @@ func handlerFileMonitorConfig(w http.ResponseWriter, r *http.Request, ps httprou
 			}
 
 			for i, cfilter := range profConf.Filters {
-				if cfilter.Filter == filter.Filter {
+				if (cfilter.Filter == filter.Filter) || (cfilter.Path == base) {
```
jaimeyu-suse, line 241: Note that `Filter` is not the path, this is the pseudo regex expression. eg: /tmp/.*

While the additional check if Path is the same as base will throw an error if the path is the same. The reason is that the monitoring logic can only store one filter per directory.

share/fsmon/monitor.go
```diff
@@ -575,8 +575,7 @@ func (w *FileWatch) getCoreFile(cid string, pid int, profile *share.CLUSFileMoni
 	dirList := make(map[string]*osutil.FileInfoExt)
 	singleFiles := make([]*osutil.FileInfoExt, 0)
 
-	// get files and dirs from all filters
-	for _, filter := range profile.Filters {
+	addFilterFn := func(filter share.CLUSFileMonitorFilter){
```
jaimeyu-suse, line 578: This is cosmetic to reduce the copy paste code. No changes in the logic was done and will make it easier to make changes in the future.

controller/rest/file_monitor.go
jayhuang-suse: Please add some test cases and make sure this logic does help.

PR #935, opened 2023-08-01
NVSHAS-6572: Parse image config spec referred in manifest

becitsthere:


share/scan/registry.go
```diff
@@ -70,52 +70,79 @@ func IsQuayRegistry(rc *RegClient) bool {
...
-func BuildV2ImageInfo(dg string, body []byte) (imageInfo ImageInfo, parsedSchemaVersion int, err error) {
+func (rc *RegClient) copyV2LayerAndHistory(imageInfo *ImageInfo, manV2 *manifestV2.Manifest, ccmi *registry.ManifestInfo) bool {
+	allLayersAreCosignPayloads := true
+
+	// In the history list from container image config spec, only the layer that has no empty_layer flag
+	// has a digest in the manifest layer list.
+	// The following section bring the layer list in imageInfo to the same size as history (cmd)
+	j := len(manV2.Layers) - 1
+	for i := 0; i < len(ccmi.Cmds); i++ {
+		if (ccmi != nil && ccmi.EmptyLayers[i]) || j < 0 {
```
williamlin-suse, line 81: ccmi should be always non-nil when it reaches line 81 or it would panic in line 80

becitsthere: ok

share/scan/registry.go
```diff
@@ -70,52 +70,79 @@ func IsQuayRegistry(rc *RegClient) bool {
...
 
-func BuildV2ImageInfo(dg string, body []byte) (imageInfo ImageInfo, parsedSchemaVersion int, err error) {
+func (rc *RegClient) copyV2LayerAndHistory(imageInfo *ImageInfo, manV2 *manifestV2.Manifest, ccmi *registry.ManifestInfo) bool {
+	allLayersAreCosignPayloads := true
+
+	// In the history list from container image config spec, only the layer that has no empty_layer flag
+	// has a digest in the manifest layer list.
+	// The following section bring the layer list in imageInfo to the same size as history (cmd)
+	j := len(manV2.Layers) - 1
+	for i := 0; i < len(ccmi.Cmds); i++ {
```
williamlin-suse, line 80: it looks like ccmi could be nil when copyV2LayerAndHistory() is called

share/scan/registry.go
```diff
@@ -70,52 +70,79 @@ func IsQuayRegistry(rc *RegClient) bool {
 	return strings.EqualFold(rc.URL[:len(quayRegistryURL)], quayRegistryURL)
 }
 
-func BuildV2ImageInfo(dg string, body []byte) (imageInfo ImageInfo, parsedSchemaVersion int, err error) {
+func (rc *RegClient) copyV2LayerAndHistory(imageInfo *ImageInfo, manV2 *manifestV2.Manifest, ccmi *registry.ManifestInfo) bool {
+	allLayersAreCosignPayloads := true
+
```
williamlin-suse, line 75: Should imageInfo.Layers be reset to empty first after entering this function?
Or imageInfo.Layers is always empty when copyV2LayerAndHistory() is called?

share/scan/registry.go
becitsthere: OK

share/scan/registry.go
becitsthere: It is initiated in GetImageInfo(). It is always empty when entering here.

becitsthere: Maybe we should, because there are two places where buildV2ImageInfo() is called.

PR #946, opened 2023-08-10
NVSHAS-8170 v2: make auto-profile default and adjustable memory threshold

jayhuang-suse:
(1) Make the auto-profile as default setting( no required flag)
(2) Adjustable memory threshold.

For example, without a k8s resource limits:
(1) AUTO_PROFILE_COLLECT=0, no snapshot.
(2) AUTO_PROFILE_COLLECT=1 (default), snapshots above 60% of the ctl[6GB <= 3.6GB] and enf[3GB <= 1.8GB]
(3) AUTO_PROFILE_COLLECT=300 (300MB), snapshots above 60% of 300MB <= 180MB
(4) AUTO_PROFILE_COLLECT=-100 (invalid), no snapshot.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -275,7 +275,7 @@ func main() {
 	disable_auto_benchmark := flag.Bool("no_auto_benchmark", false, "disable auto benchmark")
 	disable_system_protection := flag.Bool("no_sys_protect", false, "disable system protections")
 	policy_puller := flag.Int("policy_puller", 0, "set policy pulling period")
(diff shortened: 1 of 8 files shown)


jayhuang-suse:
"data.json "

```
	type snapshotData struct {
		RecordedAt        time.Time
		MemoryLimit       uint64
		WorkingMemory     uint64
		MemPercentage     int
		ProfileLimit      uint64    <== 1: follow memory limit; others: memory threshold bytes
		ProfilePercentage int     <== percentage
		Cid               string
		Lsof              []string
		Ps                []string
	}
```

PR #948, opened 2023-08-11
NVSHAS 7925 - Fixes to cgroup logic

jaimeyu-suse:
Also fixes NVSHAS-7990 (probably other cgroup problems).

7990 reported error in accessing the cgroup container metrics. When I debugged the containers, the metrics were located in `/host/cgroup` and not in `/sys/fs/cgroup`. The `/sys/fs/cgroup` path was local to the container and not of the host so none of the paths resolved. Many places were hardcoded to use the enforcer's container's paths instead of the host mounted ones and this could cause a mismatch in information. 

In 7925 - The issue was that the metrics didn't make sense in the UI. The "deployment' tab would show some vague metrics that I couldn't correlate and the container tab would sometimes show the correct values. 

The issue here was the pod holders (the container that start the /pause task) would start up and the engine would seem to grab the wrong container ID. As well, the containers looked like it would fallback silently which would assume the enforcer's metrics. 

This PR covers these two problems and possible future ones.
1. At startup, I updated the cgroup version decision making tree to make it testable.
2. NewSystemtools() does not change behaviour, it retains its original behaviour. It has been re-organized so its easier to test and with the start up fixes. 
3. When a container is added, it will follow a different flow than before. It is simplified now
4. We now use the systemtool's proc path (which ideally will be /host/proc) to get a PID's information
5. Changed how we parse the cgroup file in the proc directory. Before it assumed a single type of system but I added, hopefully, more robust handlers. 
6. It will check for the subsystem in the cgroup file. If the subsystem isn't in the file, it means the subsystem is not active (either host doesn't have it turned on or the distribution does not support it)
7. It will clean the paths in the cgroup file and fix the mapping if needed. I found issues with docker k8s where the paths were relative but it made no sense. I haven't found documentation why this is the case yet. 
8. It will apply the correct v2 and v1 qualifiers to the cgroup path and then use the discovered cgroup path at startup. Ideally it will be /host/cgroup which should be a mirror of the host' /sys/fs/cgroup.
9. cgroup v2 hybrid is not supported. The one setup that I have (ubuntu 18.04) that does support cgroup v2 hybrid doesn't support all the subsystems in v2 mode. So rather than have more logic to deal with hybrid mode, I made it so we just assume cgroup v2 pure or v1 "pure". I did add logic to deal with it but it almost always fallbacks to v1 mode so its probably easier to just avoid it. 

The previous fixes were trying to fix why /sys/fs/cgroup was not resolving correctly which was a red herring. IMO the root cause was actually because we were using the container's cgroup directory and not the host's. 

I added a bunch of unit tests to cover the the different types of cgroup files that may exist. It is not exhaustive and will require more expamples.

PR was tested against

Ubuntu 18.04 Docker all in one:  
```
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Runtimes: runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3dce8eb055cbb6872793272b4f20ed16117344f8
 runc version: v1.1.7-0-g860f061
 init version: de40ad0
```

Ubuntu 23.04 Docker k8s (this one reliably reproduces the bug for 7925)
```
 Cgroup Driver: systemd
 Cgroup Version: 2
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3dce8eb055cbb6872793272b4f20ed16117344f8
 runc version: v1.1.7-0-g860f061
```

AWS EKS  (this one had the 7990 bug)
```k8s=1.26.6-eks-a5565ad ```

Currently being tested on another AWS cluster (I don't have the details yet on it).

### DIFF
--- a/agent/timer.go
+++ b/agent/timer.go
@@ -224,9 +224,20 @@ func updateAgentStats(cpuSystem uint64) {
 
 func updateContainerStats(cpuSystem uint64) {
 	for _, c := range gInfo.activeContainers {
-		mem, _ := global.SYS.GetContainerMemoryUsage(c.cgroupMemory)
-		cpu, _ := global.SYS.GetContainerCPUUsage(c.cgroupCPUAcct)
-		system.UpdateStats(&c.stats, mem, cpu, cpuSystem)
+		var uc, um uint64
+
+		if mem, err := global.SYS.GetContainerMemoryUsage(c.cgroupMemory); err != nil {
+			log.WithFields(log.Fields{"Container": c, "error": err.Error()}).Error("Could not get memory stats")
+		} else {
+			um = mem
+		}
+		if cpu, err := global.SYS.GetContainerCPUUsage(c.cgroupCPUAcct); err != nil {
+			log.WithFields(log.Fields{"Container": c, "error": err.Error()}).Error("Could Not get CPU stats")
+		} else {
+			uc = cpu
+		}
+
+		system.UpdateStats(&c.stats, um, uc, cpuSystem)
 	}
 }
 
--- a/share/system/cgroup_linux.go
+++ b/share/system/cgroup_linux.go
@@ -332,6 +332,42 @@ func getCgroupPath_cgroup_v2(pid int) (string, error) {
 
 }
 
+// GetContainerCgroupPath - Gets a PID's cgroup path for a subsystem
+// PIDs can share containers so we can derive the pid's cgroup path by parsing proc/<pid>/cgroup
+func (s *SystemTools) GetContainerCgroupPath(pid int, subsystem string) (string, error) {
+	subsystemPath := ""
+	mpath, err  := s.getCgroupMetricsPath(pid, subsystem)
+	if err != nil {
+		// Just return the fallback, callers don't care about the error!
+		return mpath, nil
+	}
+
+	// Join to the system's path for host's cgroup
+	subsystemPath = s.JoinToCgroupPath(mpath, subsystem)
+
+	return subsystemPath, nil
+}
+
+// JoinToCgroupPath - Join to the system's path for host's cgroup
+// For cgroup v1, the subsystems are directories in the root and then the namespaces are within them
+// For cgroup v2, a flat directory structure is used and the namespace's directory now holds all the subsystems files
+func (s *SystemTools) JoinToCgroupPath(path string, subsystem string) string {
+	subsystemPath := ""
+	switch s.cgroupVersion {
+	case cgroup_v2:
+		subsystemPath = filepath.Join(s.cgroupDir, path)
+	// unsupported - Found issues in ubuntu 18.04 and missing subsystem in the unified directories
+	//case cgroup_v2_hybrid:
+	//	subsystemPath = filepath.Join(s.cgroupDir, "/unified", mpath)
+	case cgroup_v1:
+		fallthrough
+	default:
+		subsystemPath = filepath.Join(s.cgroupDir, subsystem, path)
+	}
+	return subsystemPath
+}
+
+
 
 // getStatsPathFromCgroupFile - Opens the proc/<pid>/cgroup file and parses it for the subsystem and paths
 func getStatsPathFromCgroupFile(f io.Reader, subsystem string) (string, error) {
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: agent/timer.go [hunk 1]
```diff
@@ -224,9 +224,20 @@ func updateAgentStats(cpuSystem uint64) {
 
 func updateContainerStats(cpuSystem uint64) {
 	for _, c := range gInfo.activeContainers {
-		mem, _ := global.SYS.GetContainerMemoryUsage(c.cgroupMemory)
-		cpu, _ := global.SYS.GetContainerCPUUsage(c.cgroupCPUAcct)
-		system.UpdateStats(&c.stats, mem, cpu, cpuSystem)
+		var uc, um uint64
+
+		if mem, err := global.SYS.GetContainerMemoryUsage(c.cgroupMemory); err != nil {
```
jaimeyu-suse, line 229: Errors were getting silenced and making it difficult to debug in field.

share/system/cgroup_linux.go
```diff
@@ -332,76 +332,147 @@ func getCgroupPath_cgroup_v2(pid int) (string, error) {
...
-	/*
-		mnt, err := FindCgroupMountpoint(subsystem)
-		if err != nil {
-			return "", nil
-		}
-	*/
+	subsystemPath := ""
+	mpath, err  := s.getCgroupMetricsPath(pid, subsystem)
+	if err != nil {
+		// Just return the fallback, callers don't care about the error!
```
jaimeyu-suse, line 341: Not returning an error here because previous behaviour is to ignore the errors. 

The callers only care about having something that works so worst case, we're going to send up the fallback path.

share/system/cgroup_linux.go
```diff
@@ -332,76 +332,147 @@ func getCgroupPath_cgroup_v2(pid int) (string, error) {
...
+		return mpath, nil
+	}
 
-	// Alternative: it might not be necessary to obtain cgroup path as before
-	var path string
+	// Join to the system's path for host's cgroup
+	subsystemPath = s.JoinToCgroupPath(mpath, subsystem)
```
jaimeyu-suse, line 346: This is the biggest difference in logic. 

Previously, we assumed /proc/<pid>/root/sys/fs/cgroup will always exists. 

The new logic does away with this assumption and instead follows the convention that the container runtime will store the cgroup data in /sys/fs/cgroup (on the host, in the container we expect this to be mounted to /host/cgroup).

share/system/cgroup_linux.go
```diff
@@ -332,76 +332,147 @@ func getCgroupPath_cgroup_v2(pid int) (string, error) {
...
+func (s *SystemTools) getCgroupMetricsPath(pid int, subsystem string) (string, error) {
+	var path string
+
+	if pid == 0 { // self
+		path = "/proc/self/cgroup"
+	} else {
+		path = filepath.Join(s.procDir, strconv.Itoa(pid), "cgroup")
 	}
 
-	// However, the k8s POD does not have those subsystem folders
```
jaimeyu-suse, line 360: In the original code, this logic would never get executed, we would always be in v1 or v2 mode.

share/system/cgroup_linux.go
```diff
@@ -332,76 +332,147 @@ func getCgroupPath_cgroup_v2(pid int) (string, error) {
...
+
+		tokens := strings.Split(scanner.Text(), ":")
+
+		if len(tokens) > 2 {
+			// https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-default_cgroup_hierarchies
+			// https://docs.fedoraproject.org/en-US/quick-docs/understanding-and-administering-systemd/
+			// cgroup path is defined by <index>:<subsystem(s)>:<path>
+			// example: 2:cpuset,cpu,cpuacct,memory,net_cls,net_prio,hugetlb:/kubepods/besteffort/pod1fe19bf5-e8ef-11e8-900c-52daee5a874d/da31e536c8d61304a6d5998d163d12400a7a9a1003e1d86369e8fadb022fc17d
+
+			// The subsystem is the second item in the line
```
jaimeyu-suse, line 386: Logic should be simplified here because instead of all the fallback logic, I simplified it to use a map and we just update the map with the subsystem and paths. 

The subsystem map can be cached with the container if we care enough about it.

share/system/cgroup_linux.go
```diff
@@ -332,76 +332,147 @@ func getCgroupPath_cgroup_v2(pid int) (string, error) {
...
+	return defaultHostCgroup, fmt.Errorf("[%s] subsystem not found in tokens: %+v", subsystem, subsystemMap)
+}
+
+// CgroupRelativePathFix - Applies fixes for cgroup files with certain types of paths
+func CgroupRelativePathFix(path string) (string) {
+	if strings.Contains(path, "/..") {
+		// I can't find the documentation why the cgroup file will hold relative paths. For now, I'm applying these
+		// fixes as we encounter them. Documentation on this behaviour would make this code more robust.
+		// Example: 0::/../../kubepods-besteffort-poddee9029c_408f_4466_811d_43eea3042395.slice/docker-a737350ff4843bb79debc4e2dc98f0b1b11d40f814ea4303d9167dd70c314b95.scope
+
```
jaimeyu-suse, line 421: For NVSHAS 7925, I saw some weird behaviour here in the cgroup files that was causing issues with the previous parser. 

The relatives paths don't make sense to me and I can't find any documentation why docker does this.

share/system/cgroup_linux.go
```diff
@@ -332,76 +332,147 @@ func getCgroupPath_cgroup_v2(pid int) (string, error) {
...
+	// shouldn't reuse path, doesn't make sense
+	cpath, err := getStatsPathFromCgroupFile(f, subsystem)
+	if err != nil {
+		log.WithFields(log.Fields{
+			"pid": pid,
+			"path": cpath,
+			"s.cgroupDir": s.cgroupDir,
+			"s.procDir": s.procDir,
+			"s.cgroupVersion": s.cgroupVersion,}).
+			Error("Cgroup file could not be parsed, fallback used")
```
jaimeyu-suse, line 460: For debugging the field, the error will dump enough information to help someone to jump on the node and then inspect the container's mounts and determine what went wrong.

PR #956, opened 2023-08-17
AArch64 support

stevecapperarm:
Hello,
This series enables AArch64 support in Neuvector.

Specifically it removes all the pre-compiled/embedded third party dependencies, plumbs in the distro-supplied libraries and fixes a few bugs that caused issues on AArch64. Notably the golang code has been modified to read the DMI information from /sys/firmware/dmi/tables/DMI rather than /dev/mem (which caused SIGBUS issues on AArch64 and won't work for platforms booted with Secure Boot enabled).

Note that we are pushing a few pull requests to different Neuvector repos. For testing purposes, all these should be considered together.

Cheers, Steve


garyduan:
We will rework this and merge with the current build process.

PR #959, opened 2023-08-18
NVVSHAS-7962/7980: allow image scanning on sigstore updates

alopez-suse:
Previously, when a user updated a sigstore signature image or changed their sigstore verifiers configuration, the controller would not recognize these as updates that require the image to be rescanned.

This PR fixes that by checking for changes in the sigstore signature image digest and a new "last changed" timestamp for the sigstore configuration.

This PR also allows for the controller to request a "signatures only" scan from the scanner, where only the signature data will be re-validated in the scanner. The response to such a request will only contain valid info for the `SignatureInfo` field from the ScanResult object. The new signatures are then added to the persisted scan result attached to the image summary in the cluster registry object.

### DIFF
--- a/controller/cache/group.go
+++ b/controller/cache/group.go
@@ -552,22 +552,24 @@ func groupConfigUpdate(nType cluster.ClusterNotifyType, key string, value []byte
 			evhdls.Trigger(EV_GROUP_DELETE, name, cache)
 		}
 
-		err1 := clusHelper.DeleteProcessProfile(name)
-		err2 := clusHelper.DeleteFileMonitor(name)
-		if cache != nil && cache.group != nil {
-			if isLeader() && cache.group.CfgType == share.FederalCfg && (err1 == nil || err2 == nil) {
-				fedRole := fedMembershipCache.FedRole
-				if fedRole != api.FedRoleNone {
-					// it's not demote/leave/kicked from fed
-					clusHelper.UpdateFedRulesRevision([]string{share.FedProcessProfilesType, share.FedFileMonitorProfilesType})
-				}
-			}
-			if cache.group.Kind == share.GroupKindContainer {
-				clusHelper.DeleteDlpGroup(name)
-				clusHelper.DeleteWafGroup(name)
+		var err error
+		txn := cluster.Transact()
+		clusHelper.DeleteProcessProfileTxn(txn, name)
+		clusHelper.DeleteFileMonitorTxn(txn, name)
+		if cache != nil && cache.group != nil && cache.group.Kind == share.GroupKindContainer {
+			clusHelper.DeleteDlpGroup(txn, name)
+			clusHelper.DeleteWafGroup(txn, name)
+		}
+		clusHelper.DeleteCustomCheckConfig(txn, name)
+		_, err = txn.Apply()
+		txn.Close()
+
+		if isLeader() && fedMembershipCache.FedRole != api.FedRoleNone && err == nil {
+			if cache != nil && cache.group != nil && cache.group.CfgType == share.FederalCfg {
+				// it's not demote/leave/kicked from fed
+				clusHelper.UpdateFedRulesRevision([]string{share.FedProcessProfilesType, share.FedFileMonitorProfilesType})
 			}
 		}
-		clusHelper.DeleteCustomCheckConfig(name)
 	}
 }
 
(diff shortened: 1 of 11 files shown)


share/scan/registry.go
```diff
@@ -137,6 +138,9 @@ func (rc *RegClient) buildV2ImageInfo(imageInfo *ImageInfo, ctx context.Context,
 }
 
 func (rc *RegClient) GetImageInfo(ctx context.Context, name, tag string, manifestReqType registry.ManifestRequestType) (*ImageInfo, share.ScanErrorCode) {
+	if manifestReqType == registry.ManifestRequest_CosignSignature {
+		log.WithFields(log.Fields{"name": name, "tag": tag}).Error("retrieving signature information")
```
williamlin-suse, line 142: why this is an Error log?

controller/scan/registry.go
```diff
@@ -1472,16 +1485,21 @@ func (rs *Registry) scheduleScanImages(
 					sum.Status = api.ScanStatusScheduled
 					clusHelper.PutRegistryImageSummary(rs.config.Name, sum.ImageID, sum)
 				}
+
+				if signatureInfoChanged && sum.Status != api.ScanStatusScheduled {
+					skip = false
```
williamlin-suse, line 1490: When an image's digest doesn't change but nv sigstore settings change(assuming not scheduled yet), the image will be added to the scan task queue.
However, image digest is still the same which means vulnerability scan is not necessary if the image is never scanned(i.e only image signature verification is needed).
Does scanner have this mechanism to do image signature verification only?

share/scan/registry.go
```diff
@@ -278,6 +282,15 @@ func (rc *RegClient) GetImageInfo(ctx context.Context, name, tag string, manifes
 
 	imageInfo.RawManifest = body
 
+	if manifestReqType != registry.ManifestRequest_CosignSignature {
+		signatureTag := GetCosignSignatureTagFromDigest(imageInfo.Digest)
+		signatureImageInfo, errCode := rc.GetImageInfo(ctx, name, signatureTag, registry.ManifestRequest_CosignSignature)
+		if errCode != share.ScanErrorCode_ScanErrNone && errCode != share.ScanErrorCode_ScanErrImageNotFound {
+			return nil, errCode
```
williamlin-suse, line 289: Does it mean any error for getting signature image info will cause the image not scanned for vulnerability?
I don't think we want that behavior.

controller/scan/registry.go
```diff
@@ -879,51 +879,114 @@ func (rs *Registry) checkAndPutImageResult(sctx *scanContext, id string, result
 		// not canceled, continue
 	}
 
+	if result.ScanTypesRequested == nil {
```
williamlin-suse, line 882: When result.ScanTypesRequested is nil, does it mean scanner is older version that supports vulnerability scan only?
If so, should 'Vulnerability' filed be set to true when result.ScanTypesRequested is nil ?

controller/scan/registry.go
```diff
@@ -879,51 +879,114 @@ func (rs *Registry) checkAndPutImageResult(sctx *scanContext, id string, result
...
+		if result.ScanTypesRequested.Vulnerability {
+			sum.Size = result.Size
+		}
 
-		if sum.Status == api.ScanStatusFinished {
-			sum.ScanFlags |= share.ScanFlagCVE
-			if len(result.Layers) != 0 {
-				sum.ScanFlags |= share.ScanFlagLayers
+		// handle status for vuln scan
+		if result.ScanTypesRequested.Vulnerability {
```
williamlin-suse, line 890: in https://github.com/neuvector/scanner/pull/85, when is result.Error set to share.ScanErrorCode_ScanErrSignatureScanError ?
(line 905 in this file checks about share.ScanErrorCode_ScanErrSignatureScanError)

controller/scan/registry.go
```diff
@@ -879,42 +879,108 @@ func (rs *Registry) checkAndPutImageResult(sctx *scanContext, id string, result
...
+			sum.Author = result.Author
+			sum.Size = result.Size
+			sum.Result = result.Error // this represents the vuln scan error
+			if result.Error == share.ScanErrorCode_ScanErrNone {
+				sum.Status = api.ScanStatusFinished
+			} else if result.Error == share.ScanErrorCode_ScanErrNotSupport {
+				sum.Status = api.ScanStatusFinished
+			} else if retAction == scheduler.TaskActionRequeue {
+				sum.Status = api.ScanStatusScheduled
+			} else if result.Error == share.ScanErrorCode_ScanErrSignatureScanError {
```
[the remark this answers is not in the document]
alopez-suse: This check is unnecessary and should be removed, signature errors are not handled with `sum.Status` anymore.

However, I've updated the scanner code to return `SignatureScanError` as it is more appropriate than the other error codes I was using.

controller/scan/registry.go
alopez-suse: Ok. This has been implemented. Scanner has also been updated so that the newer versions never return a nil `req.ScanTypesRequested` so that this behavior is consistent.

PR #967, opened 2023-08-24
NVSHAS 7790 and 7925 - Fixes to CPU/Memory stat files not found

jaimeyu-suse:
Changed updateContainerStats() to deal with errors.
    There is a check if the container is a pod holder as those errors are
    valid under certain OS.
    Added error handling when we first register a container in
    fillContainerProperties(). If the cgroup files fail to load, we blank
    out the path and never process it again. This should silence the errors
    in the logs with at least one message instead of a flood of them.

### DIFF
--- a/agent/timer.go
+++ b/agent/timer.go
@@ -226,13 +226,19 @@ func updateContainerStats(cpuSystem uint64) {
 	for _, c := range gInfo.activeContainers {
 		var mem, cpu uint64 = 0, 0
 		var err error
-		mem, err = global.SYS.GetContainerMemoryUsage(c.cgroupMemory)
-		if err != nil {
-			log.WithFields(log.Fields{"pid": c.pid, "id": c.id, "error": err}).Warning("Memory stats not found")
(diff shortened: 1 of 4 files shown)


agent/timer.go
```diff
@@ -224,8 +224,22 @@ func updateAgentStats(cpuSystem uint64) {
 
 func updateContainerStats(cpuSystem uint64) {
 	for _, c := range gInfo.activeContainers {
-		mem, _ := global.SYS.GetContainerMemoryUsage(c.cgroupMemory)
-		cpu, _ := global.SYS.GetContainerCPUUsage(c.cgroupCPUAcct)
+		var mem, cpu uint64 = 0, 0
+		var err error
+		if c.cgroupMemory != "" {
+			mem, err = global.SYS.GetContainerMemoryUsage(c.cgroupMemory)
+			if err != nil && c.parentNS != "" {
```
jaimeyu-suse, line 231: Not sure I want to keep the pod holder check. This should fail at fillContainerInfo so we could probably drop this.

PR #969, opened 2023-08-24
NVSHAS 7925 and 7990 - Cgroup logging errors

jaimeyu-suse:
getCgroupPathReaderV2 no longer returns a default path, it returns ""
instead.
cgroup_test modified to deal with the change.
fillContainerProperties() will test the path at start up to check if it
exist. If it doesn't work, then we force the path to be "" so we don't
try to get stats.
updateContainerStats() checks for a non empty cgroup path. It also does
error handling and messaging here.
getMemoryStats() returns an error on file not found so we no longer
silence errors.

The changes for getCgroupPathReaderV2() is because in 7925, we would fallback to the node's stats which was very confusing. So we're forcing it to be 0 so its clearer that we should check the logs for errors.


agent/engine.go
```diff
@@ -103,17 +103,17 @@ type localSystemInfo struct {
 	localSubnetMap map[string]share.CLUSSubnet
 	// all subnets that all containers in the cluster connect to, including
 	// local subnets and subnets populated from controller
-	internalSubnets map[string]share.CLUSSubnet
```
jaimeyu-suse, line 106: No functional changes. Don't know why my IDE decided to reformat this.

PR #970, opened 2023-08-24
NVSHAS 7990 & 7925 - cgroup error log fixes

jaimeyu-suse:
Changed getMemoryStats() to return an error for no files found.

Changed updateContainerStats() to deal with errors. There is a check if the container is a pod holder as those errors are valid under certain OS.
Added error handling when we first register a container in fillContainerProperties(). If the cgroup files fail to load, we blank out the path and never process it again. This should silence the errors in the logs with at least one message.

### DIFF
--- a/share/system/cgroup_linux.go
+++ b/share/system/cgroup_linux.go
@@ -383,8 +383,10 @@ func (s *SystemTools) GetContainerCgroupPath(pid int, subsystem string) (string,
 			return "", err
 		}
 		if cpathv2 == "" {
-			return "", fmt.Errorf("cgroup file does not have valid cgroup paths")
+			// path wasn't parsed, so we'll default to v1 style for best effort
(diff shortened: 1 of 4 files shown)


jayhuang-suse (approved):

share/system/cgroup_linux.go
```diff
@@ -595,6 +627,9 @@ func (s *SystemTools) getMemoryStats(path string, mStats *CgroupMemoryStats, bFu
 
 // https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs2/cpu.go
 func (s *SystemTools) statCpu(path string, stats *CpuStats) error {
+	if path == "" {
```
jaimeyu-suse, line 630: Keeping this and getMemoryStats() for the other functions that call this but aren't aware that the path can be empty.

jayhuang-suse (approved):

PR #1018, opened 2023-09-21
NVSHAS-7947 - Process violation is reported as Host.Suspicious instead Container.Suspicious

jaimeyu-suse:
I believe this is a race condition where the container map being updated at the same time we are evaluating a new process. The map data is not available so it assumes it is a host process.

I made a change in evaluateApplication() so when it detects a violation, it will make a last ditch attempt to confirm the process is not a host process by checking:
1. if pidContainerMap has the pid and has an updated information about the container
2. Make another attempt at GetContainerIDByPID() to determine the container ID from the PID.

I also added some extra information to recordIncident() because the log entry was not useful in debugging as is. Either we make it more useful or we remove it.

### DIFF
--- a/agent/probe/process.go
+++ b/agent/probe/process.go
@@ -1626,10 +1626,6 @@ func (p *Probe) addContainerCandidate(proc *procInternal, scanMode bool) (*procC
 		}
 		return nil, -1 // not ready
 	}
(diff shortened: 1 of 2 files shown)


jayhuang-suse (approved):
good

PR #1034, opened 2023-09-27
NVSHAS-8322 & NVSHAS-8323: fix unnecessary signature rescans

alopez-suse:
The controller correctly checks whether an unscanned image is signed, skipping signature scans for unsigned images. The skip logic has also been cleared up.

### DIFF
--- a/controller/scan/registry.go
+++ b/controller/scan/registry.go
@@ -1506,6 +1506,7 @@ func (rs *Registry) scheduleScanImages(
 				Vulnerability: false,
 				Signature:     false,
 			}
+			isSignedImage := info.SignatureDigest != ""
 			sum, ok := rs.summary[info.ID]
 			if ok {
 				smd.scanLog.WithFields(log.Fields{
@@ -1532,6 +1533,7 @@ func (rs *Registry) scheduleScanImages(
 					}
 				}
 
+				// determine if vuln scan is required
 				if !newImage {
 					scanTypesRequired.Vulnerability = false
 				} else if rs.bSkipVulnScanForImage(sum) {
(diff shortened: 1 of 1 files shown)


controller/scan/registry.go
```diff
@@ -1544,29 +1550,21 @@ func (rs *Registry) scheduleScanImages(
...
-				sigstoreConfigurationChanged := sigstoreTimestamp != sum.SigstoreTimestamp
-
-				if signatureInfoChanged {
+				// determine if signature scan is required
+				if !newImage {
+					scanTypesRequired.Signature = false
+				} else if signatureInfoChanged {
+					smd.scanLog.WithFields(log.Fields{"registry": rs.config.Name, "image": image, "status": sum.Status, "imageID": info.ID}).Debug("Signature info for image changed, signature scan required.")
+					scanTypesRequired.Signature = true
 					sum.SignatureDigest = info.SignatureDigest
```
williamlin-suse, line 1559: When newImage is true & imageChanged is true & signatureInfoChanged is true, it reaches here.
sum.SignatureDigest is set to the new signature image digest.
Then in line 1594 the summary is written to kv and scan task is scheduled. 
Does it mean if the scan fails later(because of network, registry, ....), sum.SignatureDigest is already set to the new digest? Is this correct?

controller/scan/registry.go
```diff
@@ -1544,29 +1550,21 @@ func (rs *Registry) scheduleScanImages(
...
 					imageChanged = true
+				} else if sigstoreConfigurationChanged && isSignedImage {
+					smd.scanLog.WithFields(log.Fields{"registry": rs.config.Name, "image": image, "status": sum.Status, "imageID": info.ID}).Debug("Sigstore config changed and image is signed, signature scan required.")
 					scanTypesRequired.Signature = true
-					smd.scanLog.WithFields(log.Fields{"registry": rs.config.Name, "image": image, "status": sum.Status, "imageID": info.ID}).Debug("Signature Info Changed")
-				}
-
-				if sigstoreConfigurationChanged {
 					sum.SigstoreTimestamp = sigstoreTimestamp
```
williamlin-suse, line 1564: When newImage is true & imageChanged is true & signatureInfoChanged is false and (sigstoreConfigurationChanged && isSignedImage) is true, it reaches here.
sum.SigstoreTimestamp is set to the latest sigstore timestamp.
Then in line 1594 the summary is written to kv and scan task is scheduled.
Does it mean if the scan fails later(because of network, registry, ....), sum.SigstoreTimestamp is already set to the latest sigstore timestamp? Is this correct?

controller/scan/registry.go
```diff
@@ -1544,29 +1550,21 @@ func (rs *Registry) scheduleScanImages(
...
-						// image is signed, rescan with new sigstore configuration
-						smd.scanLog.WithFields(log.Fields{"registry": rs.config.Name, "image": image, "status": sum.Status, "imageID": info.ID}).Debug("Sigstore Config Changed")
-						scanTypesRequired.Signature = true
-					}
-				}
-
-				if sum.SignatureStatus == api.ScanStatusFailed {
-					scanTypesRequired.Signature = true
-				} else if newImage && info.SignatureDigest != "" {
+				} else if sum.SignatureStatus == api.ScanStatusFailed {
```
williamlin-suse, line 1566: should it be
else if sum.SignatureStatus == api.ScanStatusFailed && isSignedImage {
?

controller/scan/registry.go
```diff
@@ -1544,29 +1550,21 @@ func (rs *Registry) scheduleScanImages(
 					scanTypesRequired.Vulnerability = true
 				}
 
-				signatureInfoChanged := info.SignatureDigest != sum.SignatureDigest
-				sigstoreConfigurationChanged := sigstoreTimestamp != sum.SigstoreTimestamp
-
-				if signatureInfoChanged {
+				// determine if signature scan is required
+				if !newImage {
```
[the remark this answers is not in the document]
alopez-suse: This specific `newImage` variable doesn't actually mean the image hasn't been scanned, but instead that the we have seen the ID before in this batch of potential images to scan.

controller/scan/registry.go
alopez-suse: If the scan fails, then it will set the `SignatureStatus` result in the scan field to `failed` which should then retry the signature scan regardless.

controller/scan/registry.go
alopez-suse: Similar situation to `signatureDigest`, see this comment:
https://github.com/neuvector/neuvector/pull/1034#discussion_r1339259333

controller/scan/registry.go
alopez-suse: With the logic now, `sum.SignatureStatus` should only ever show a fail for a signed image anyways, since we are only scanning signed images.

controller/scan/registry.go
```diff
@@ -1507,6 +1507,9 @@ func (rs *Registry) scheduleScanImages(
 				Signature:     false,
 			}
 			sum, ok := rs.summary[info.ID]
+
+			isSignedImage := info.SignatureDigest != "" || (ok && sum.SignatureDigest != "")
```
williamlin-suse, line 1511: does sum.SignatureDigest really matter for var isSignedImage ?

alopez-suse: After reviewing the logic, it was not necessary. I've removed it.

williamlin-suse (approved):

PR #1039, opened 2023-09-29
NVSHAS-8053 - 'ps' process incidents are raised by root process

jaimeyu-suse:
Added fixes to alleviate the issue.
* We get the process' parent path now. The issue before was that we were passing the path of the process to isParentProcessException(). This would not give a root process an exception because the function would not be able to get determine the parent's info. Now that I have the parent's path and passed it to the function, the process is given the exception.
* Username and euid is available in the event so its a bit clearer in the event.


jayhuang-suse (changes requested):
I agree with the solution, but the call to prober might violate the map's access violation. I prefer without the user information in the incident event.

username := fa.prober.getUserName(ppid, euid)

PR #1082, opened 2023-11-07
NVSHAS-5999 AWS bottlerocket benchmark execution error

pohanhuangtw:
### Summary
1. we check if the current OS is bottlerocket,

- we access the config with /proc/1/root/ and run it like a pure shell (without nstool)

2. **Export base image bin and LD bin as part of the env variable** to make sure we are able to run under lack of instruction circumstance
3. Remove **kubeCheckPrerequisite.**,  since we will make sure the env has the corresponding linux command.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -116,6 +116,7 @@ func getLocalInfo(selfID string, pid2ID map[int]string) error {
 	}
 
 	agentEnv.startsAt = time.Now().UTC()
+	Agent.Pid = os.Getpid()
 	if agentEnv.runInContainer {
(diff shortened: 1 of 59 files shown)


lsongsuse (approved):

jayhuang-suse (changes requested):
(1) newBench: no need to pass the agent pid. Use the "Agent.Pid" in the function.
(2) kubeCheckPrerequisites: please evaluate the negative results. Does it block its further actions?

jayhuang-suse (approved):
Agent codes are good.

PR #1120, opened 2023-12-04
NVSHAS-8334: Remote Repository Export

alopez-suse:
These changes implement NVSHAS-8334 so that a user may export their CRD files directly to a configured remote repository. Currently GitHub is the only supported provider.

The export implementation is found in `controller/remote_repository/`.

### DIFF
--- a/controller/access/access.go
+++ b/controller/access/access.go
@@ -772,6 +772,8 @@ func CompileUriPermitsMapping() {
 				"v1/system/license",
 				"v1/system/summary",
 				"v1/internal/system",
+				"v1/system/config/remote_repository",
+				"v1/system/config/remote_repository/*",
 			},
 			CONST_API_FED: []string{
 				"v1/fed/join_token",
@@ -869,6 +871,7 @@ func CompileUriPermitsMapping() {
 			CONST_API_SYSTEM_CONFIG: []string{
 				"v1/system/license/update",
 				"v1/system/config/webhook",
+				"v1/system/config/remote_repository",
 			},
 			CONST_API_IBMSA: []string{
 				"v1/partner/ibm_sa/*/setup/*",
@@ -958,6 +961,7 @@ func CompileUriPermitsMapping() {
 				"v1/system/config",
 				"v2/system/config",
 				"v1/system/config/webhook/*",
+				"v1/system/config/remote_repository/*",
 			},
 			CONST_API_FED: []string{
 				"v1/fed/cluster/*/**",
@@ -1020,6 +1024,7 @@ func CompileUriPermitsMapping() {
 			CONST_API_SYSTEM_CONFIG: []string{
 				"v1/system/license",
 				"v1/system/config/webhook/*",
+				"v1/system/config/remote_repository/*",
 			},
 			CONST_API_FED: []string{
 				"v1/fed/cluster/*",
--- a/controller/api/apis.go
+++ b/controller/api/apis.go
@@ -66,6 +66,7 @@ const RESTErrPasswordExpired int = 48
 const RESTErrPromoteFail int = 49
 const RESTErrPlatformAuthDisabled int = 50
 const RESTErrRancherUnauthorized int = 51
+const RESTErrRemoteExportFail int = 52
 
 const FilterPrefix string = "f_"
 const SortPrefix string = "s_"
@@ -519,27 +520,32 @@ type RESTListData struct {
 }
 
 type RESTGroupExport struct {
-	Groups     []string `json:"groups"`
-	PolicyMode string   `json:"policy_mode,omitempty"`
+	Groups              []string                  `json:"groups"`
+	PolicyMode          string                    `json:"policy_mode,omitempty"`
+	RemoteExportOptions *share.RemoteExportConfig `json:"remote_export_options,omitempty"`
 }
 
 type RESTAdmCtrlRulesExport struct {
-	ExportConfig bool     `json:"export_config"`
-	IDs          []uint32 `json:"ids"` // used when ExportRules is true
+	ExportConfig        bool                      `json:"export_config"`
+	IDs                 []uint32                  `json:"ids"` // used when ExportRules is true
+	RemoteExportOptions *share.RemoteExportConfig `json:"remote_export_options,omitempty"`
 }
 
 type RESTWafSensorExport struct {
-	Names []string `json:"names"`
+	Names               []string                  `json:"names"`
+	RemoteExportOptions *share.RemoteExportConfig `json:"remote_export_options,omitempty"`
 }
 
 // vlunerability profile export. only support "default" profile to export(5.3+)
 type RESTVulnProfilesExport struct {
-	Names []string `json:"names"`
+	Names               []string                  `json:"names"`
+	RemoteExportOptions *share.RemoteExportConfig `json:"remote_export_options,omitempty"`
 }
 
 // compliance profile export. only support "default" profile to export(5.3+)
 type RESTCompProfilesExport struct {
-	Names []string `json:"names"`
+	Names               []string                  `json:"names"`
+	RemoteExportOptions *share.RemoteExportConfig `json:"remote_export_options,omitempty"`
 }
 
 type RESTUser struct {
@@ -1876,6 +1882,7 @@ type RESTSystemConfig struct {
 	ScannerAutoscale          RESTSystemConfigAutoscale `json:"scanner_autoscale"`
 	NoTelemetryReport         bool                      `json:"no_telemetry_report"`
 	CspType                   string                    `json:"csp_type"`
+	RemoteRepositories        []share.RemoteRepository  `json:"remote_repositories"`
 }
 
 type RESTSystemConfigData struct {
@@ -2882,7 +2889,8 @@ type RESTCrdDlpGroupConfig struct {
 }
 
 type RESTDlpSensorExport struct {
-	Names []string `json:"names"`
+	Names               []string                  `json:"names"`
+	RemoteExportOptions *share.RemoteExportConfig `json:"remote_export_options,omitempty"`
 }
 
 type RESTDerivedWorkloadDlpRule struct {
--- a/controller/rest/dlp_rule.go
+++ b/controller/rest/dlp_rule.go
@@ -17,7 +17,6 @@ import (
 	"github.com/julienschmidt/httprouter"
 	cmetav1 "github.com/neuvector/k8s/apis/meta/v1"
 	log "github.com/sirupsen/logrus"
-	"sigs.k8s.io/yaml"
 
 	"github.com/neuvector/neuvector/controller/access"
 	"github.com/neuvector/neuvector/controller/api"
@@ -27,7 +26,6 @@ import (
 	"github.com/neuvector/neuvector/controller/rpc"
 	"github.com/neuvector/neuvector/share"
 	"github.com/neuvector/neuvector/share/cluster"
-	"github.com/neuvector/neuvector/share/utils"
 )
 
 func handlerDlpSensorList(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
@@ -292,7 +290,7 @@ func wildCardToRegexp(pattern string) string {
 			if match[1] == '*' {
 				return string(match[0]) + ".*"
 			} else if match[1] == '?' {
-				/* If the pattern starts with 2 continuous "?" characters, convert them to ".*". 
+				/* If the pattern starts with 2 continuous "?" characters, convert them to ".*".
 				   Additionally, single "?" character will be converted to ".". */
 				if match[0] == '?' {
 					return ".*"
@@ -304,7 +302,7 @@ func wildCardToRegexp(pattern string) string {
 			}
 		} else {
 			/* If the pattern starts with multiple continuous "?" characters, convert them to ".*".
- 			   Additionally, convert multiple continuous "?" characters to ".*". */
+			   Additionally, convert multiple continuous "?" characters to ".*". */
 			if match[1] == '?' {
 				if match[0] == '?' {
 					return ".*"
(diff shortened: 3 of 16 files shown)


controller/rest/rest.go
```diff
@@ -1759,6 +1759,12 @@ func StartRESTServer() {
 	r.DELETE("/v1/api_key/:name", handlerApikeyDelete)
 	r.GET("/v1/selfapikey", handlerSelfApikeyShow) // Skip API document
 
+	// remote export repository
+	r.POST("/v1/remote_export_repository", handlerRemoteExportRepositoryPost)
+	r.GET("/v1/remote_export_repository/:alias", handlerRemoteExportRepositoryGet)
```
williamlin-suse, line 1764: This will require 2 API requests for loading system configuration page.
Why not follow the similar path as webhook configurations?

share/clus_apis.go
```diff
@@ -669,6 +669,10 @@ func CLUSSigstoreTimestampKey() string {
 	return fmt.Sprintf("%s%s", CLUSConfigStore, "sigstore_timestamp")
 }
 
+func CLUSRemoteExportRepositoryKey(alias string) string {
+	return fmt.Sprintf("%s%s%s", CLUSConfigStore, "remote_export_repository/", alias)
```
williamlin-suse, line 673: Why not put the settings in object/config/system?
Separate it from object/config/system will require changes on configmap  & kv backup/restore about the new key object/config/remote_export_repository but I don't see it in this PR

controller/api/apis.go
```diff
@@ -3625,3 +3632,30 @@ type REST_SigstoreRootOfTrustCollection struct {
 type REST_SigstoreVerifierCollection struct {
 	Verifiers []REST_SigstoreVerifier `json:"verifiers"`
 }
+
```
williamlin-suse, line 3635: The following 3 types are not found in controller/api/apis.yaml

controller/api/apis.go
```diff
@@ -3625,3 +3632,30 @@ type REST_SigstoreRootOfTrustCollection struct {
...
+type REST_RemoteExportRepository_POST struct {
```
williamlin-suse, line 3636: 1. Which field is for "git service type field" ?
2. What's the difference between "Alias" & "RepositoryName" ? If user can specify "RepositoryName", what's the purpose of "Alias"? Which function reads/writes "Alias" field in this PR?

controller/api/apis.go
```diff
@@ -3625,3 +3632,30 @@ type REST_SigstoreRootOfTrustCollection struct {
...
+	Alias                            string `json:"alias"`
+	RepositoryOwnerUsername          string `json:"repository_owner_username"`
+	RepositoryName                   string `json:"repository_name"`
+	RepositoryBranchName             string `json:"repository_branch_name"`
+	PersonalAccessToken              string `json:"personal_access_token"`
+	PersonalAccessTokenCommitterName string `json:"personal_access_token_committer_name"`
+	PersonalAccessTokenEmail         string `json:"personal_access_token_email"`
+}
+
+type REST_RemoteExportRepository_GET struct {
```
williamlin-suse, line 3646: "PersonalAccessToken" field is not in this type. How does user know it's empty or configured?

controller/api/apis.go
```diff
@@ -2882,7 +2882,14 @@ type RESTCrdDlpGroupConfig struct {
 }
 
 type RESTDlpSensorExport struct {
-	Names []string `json:"names"`
+	Names         []string           `json:"names"`
+	ExportOptions *RESTExportOptions `json:"export_options"`
+}
+
+type RESTExportOptions struct {
```
williamlin-suse, line 2889: Missing in controller/api/apis.yaml

controller/api/apis.go
```diff
@@ -2882,7 +2882,14 @@ type RESTCrdDlpGroupConfig struct {
...
+	Type            string `json:"type"`
```
williamlin-suse, line 2890: What is "type" field for? Is it for "git service type"? If yes, why it's in RESTExportOptions but not in REST_RemoteExportRepository_POST?

controller/rest/dlp_rule.go
```diff
@@ -1675,16 +1675,30 @@ func handlerDlpExport(w http.ResponseWriter, r *http.Request, ps httprouter.Para
...
-	w.Write(data)
+
+	if isRemoteExport(rconf.ExportOptions) {
+		remoteExport := RemoteExport{
+			DefaultFilePath: filename,
+			Options:         rconf.ExportOptions,
+			Content:         data,
+		}
+		err = remoteExport.Do()
+		if err != nil {
```
williamlin-suse, line 1690: When err is nil, does it need to return 200?
When err is non-nil, what is returned to the REST API client?

controller/api/apis.go
williamlin-suse, line 2890: Please put a comment here & in apis.yaml if "type" field must be "remote" for now

controller/export/github.go
```diff
@@ -0,0 +1,169 @@
...
+type GitHubExport struct {
+	repo          githubRepo
+	committer     githubCommitter
+	commitMessage string
+	filePath      string
+	fileContents  []byte
+	url           *url.URL
+}
+
+func (exp GitHubExport) Do() error {
```
williamlin-suse, line 56: What about the proxy support for exporting to github?

controller/rest/remote_export.go
```diff
@@ -0,0 +1,249 @@
...
+	} else if !acc.Authorize(&share.CLUSRemoteExportRepository{}, nil) {
+		restRespAccessDenied(w, login)
+		return
+	}
+
+	alias := ps.ByName("alias")
+	clusRemoteExportRepository, _, err := clusHelper.GetRemoteExportRepository(alias)
+	if err != nil {
+		msg := fmt.Sprintf("Could not retrieve remote export repository from kv store: %s", err.Error())
+		restRespErrorMessage(w, http.StatusInternalServerError, api.RESTErrFailWriteCluster, msg)
```
williamlin-suse, line 83: In other API handlers, they simply call
restRespNotFoundLogAccessDenied(w, login, err)
when the method fails to get the object based on the name/alias.
I wonder why it uses http.StatusInternalServerError/api.RESTErrFailWriteCluster here?

controller/rest/remote_export.go
```diff
@@ -0,0 +1,249 @@
...
+	acc, login := getAccessControl(w, r, "")
+	if acc == nil {
+		return
+	} else if !acc.Authorize(&share.CLUSRemoteExportRepository{}, nil) {
+		restRespAccessDenied(w, login)
+		return
+	}
+
+	alias := ps.ByName("alias")
+	err := clusHelper.DeleteRemoteExportRepository(alias)
```
williamlin-suse, line 111: No lock protection for deletion? See handlerSystemWebhookDelete()

controller/export/github.go
becitsthere: Let's create a jira case. We can do this as bug fix.

controller/rest/remote_export.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	body, _ := ioutil.ReadAll(r.Body)
+	var remoteExportRepository api.REST_RemoteExportRepository_POST
+	err := json.Unmarshal(body, &remoteExportRepository)
+	if err != nil {
+		msg := fmt.Sprintf("Could not unmarshal request body: %s", err.Error())
+		restRespErrorMessage(w, http.StatusBadRequest, api.RESTErrInvalidRequest, msg)
+		return
+	} else if !isObjectNameValid(remoteExportRepository.Alias) {
```
williamlin-suse, line 41: In 5.3 we only allow one entry with alias as "default".
Please check whether the Alias field is "default". Return error if it's not.

controller/access/access.go
```diff
@@ -772,6 +772,8 @@ func CompileUriPermitsMapping() {
 				"v1/system/license",
 				"v1/system/summary",
 				"v1/internal/system",
+				"v1/remote_export_repository",
```
williamlin-suse, line 775: this remote repository could be used for import in the future as well.
can we use "remote_repository" instead of "remote_export_repository"?

controller/rest/rest.go
alopez-suse: I will implement a functionality to retrieve all configured remote repositories.

As for best practices for the `system/config` store, I'll message you.

share/clus_apis.go
alopez-suse: `As for best practices for the `system/config` store, I'll message you to discuss further.`

controller/api/apis.go
alopez-suse: I'll update the `apis.yaml` file when PR conversations have settled. The proxy support can be implemented.

controller/api/apis.go
alopez-suse: 1. I wonder if we should expose the personal access token after it has been initialized, perhaps another field is necessary. We can discuss this.
2. Good point, I have added a nickname/alias field to the remote repository objects.

controller/api/apis.go
```diff
@@ -2882,7 +2882,14 @@ type RESTCrdDlpGroupConfig struct {
 }
 
 type RESTDlpSensorExport struct {
```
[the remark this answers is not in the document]
alopez-suse: I will update the `apis.yaml` when all of the changes are settled again.

controller/api/apis.go
alopez-suse: I will update the `apis.yaml` when all of the changes are settled again.

controller/api/apis.go
alopez-suse: This type field no longer exists. Instead, if a remote export request has the `RemoteExportOptions` field configured, it will trigger a remote export.

controller/rest/remote_export.go
alopez-suse: The `http.StatusInternalServerError` & `api.RESTErrFailWriteCluster` is a mistake here, changed in a recent commit.

controller/kv/helper.go
```diff
@@ -3286,3 +3292,80 @@ func (m clusterHelper) GetSigstoreTimestamp() (string, *uint64, error) {
...
+
+	value, err := json.Marshal(remoteExportRepository)
+	if err != nil {
+		return err
+	}
+
+	if txn != nil {
+		txn.Put(remoteExportRepositoryKey, value)
+	} else {
+		cluster.Put(remoteExportRepositoryKey, value)
```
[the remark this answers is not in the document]
alopez-suse: Fixed in recent commit.

controller/rest/remote_export.go
alopez-suse: With the way the endpoints work, only 1 remote repository can be created via the http api with the nickname/alias "default".

However, I just realized that I'll have ensure the object and the path parameter dont have different names.

controller/access/access.go
alopez-suse: Good point. I have changed the verbiage everywhere to just `remote repository` instead of `remote export repository`.

controller/api/apis.go
alopez-suse: I have implemented 1 and 2 in recent commits.

Yes, the `alias` (now `nickname`) is the NV identifier, not the _actual_ repository name as it appears in github.

controller/rest/dlp_rule.go
alopez-suse: Added relevant responses in recent commit.

controller/rest/remote_export.go
alopez-suse: As discussed.

controller/rest/rest.go
```diff
@@ -1759,6 +1759,12 @@ func StartRESTServer() {
 	r.DELETE("/v1/api_key/:name", handlerApikeyDelete)
 	r.GET("/v1/selfapikey", handlerSelfApikeyShow) // Skip API document
 
+	// remote export repository
```
williamlin-suse, line 1762: How does UI list all the configured remote repository entries?
Why not follow the similar path as webhook APIs like
r.POST("/v1/system/config/webhook", handlerSystemWebhookCreate)
r.PATCH("/v1/system/config/webhook/:name", handlerSystemWebhookConfig)
r.DELETE("/v1/system/config/webhook/:name", handlerSystemWebhookDelete)
?

share/clus_apis.go
```diff
@@ -669,6 +669,10 @@ func CLUSSigstoreTimestampKey() string {
 	return fmt.Sprintf("%s%s", CLUSConfigStore, "sigstore_timestamp")
 }
 
+func CLUSRemoteRepositoryKey(nickname string) string {
```
williamlin-suse, line 672: Why not put the settings in object/config/system?
Separate it from object/config/system will require changes on configmap & kv backup/restore about the new key object/config/remote_export_repository but I don't see it in this PR

controller/rest/remote_repository.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	body, _ := ioutil.ReadAll(r.Body)
+	var remoteRepository api.RemoteRepository
+	err := json.Unmarshal(body, &remoteRepository)
+	if err != nil {
+		msg := fmt.Sprintf("Could not unmarshal request body: %s", err.Error())
+		restRespErrorMessage(w, http.StatusBadRequest, api.RESTErrInvalidRequest, msg)
+		return
+	} else if !isObjectNameValid(remoteRepository.Nickname) {
```
williamlin-suse, line 35: In 5.3 we only allow one entry with alias as "default".
Please check whether the Alias field is "default". Return error if it's not.

controller/kv/helper.go
```diff
@@ -3286,3 +3292,80 @@ func (m clusterHelper) GetSigstoreTimestamp() (string, *uint64, error) {
...
+
+	value, err := json.Marshal(remoteRepository)
+	if err != nil {
+		return err
+	}
+
+	if txn != nil {
+		txn.Put(remoteRepositoryKey, value)
+	} else {
+		cluster.Put(remoteRepositoryKey, value)
```
williamlin-suse, line 3307: should it be cluster.PutIfNotExist() called?

alopez-suse: Correct. Implemented.

controller/rest/remote_repository.go
alopez-suse: I don't think this should be enforced by the back-end. As it currently stands, the API supports multiple destinations, and it has already been implemented that way.

If the web UI front-end will only support a single destination for now, it does not have to make the `nickname` field editable by the user, and only send requests using the `default` alias behind the scenes.

controller/rest/rest.go
alopez-suse: The remote repositories are now part of the `system/config` object and can be retrieved as such.

share/clus_apis.go
alopez-suse: Implemented.

PR #1122, opened 2023-12-05
[NVSHAS-8474] Resolve the symbolic link files under proc's root path

pohanhuangtw:
### Root cause

- The symlink in different mnt namespace sometime would escape the proc root by accident.

### Summary
1. Resolve the symbolic link files under proc's root path by parsing the symlink, and make sure it doesn't escape the proc root by check the prefix of the resolved path

2. Add unit test to cover GetContainerRealFilePath function with following cases

- one layer of symlink
- abs path for symlink
- append another layer on type 1, make it a nest link
- circular link
- one node accidentally access to the circular link

### Flow chart
image

### DIFF
--- a/share/osutil/file_linux.go
+++ b/share/osutil/file_linux.go
@@ -45,40 +45,71 @@ type FileInfoExt struct {
 }
 
 
-func fileExists(path string) bool {
-    _, err := os.Lstat(path)
-    return !errors.Is(err, os.ErrNotExist)
-}
-
-
-func extractProcRootPath(input string) (string, error) {
-    // Regular expression to match the pattern /proc/[number]/root/
-    re := regexp.MustCompile(`.*/proc/\d+/root/`)
-    matches := re.FindStringSubmatch(input)
-
-    if len(matches) == 0 {
-        return "", fmt.Errorf("no match found")
+func fileExists(path string) (err error) {
+    _, err = os.Lstat(path)
+    if err == nil {
+        // No error, file exists and is accessible
+        return nil
     }
 
-    return matches[0], nil
+    if errors.Is(err, os.ErrNotExist) {
+        // File does not exist
+        return err
+    }
+
+    if os.IsPermission(err) {
+        return fmt.Errorf("Permission error accessing file: %s", err)
+    }
+
+    // Other types of errors
+    return err
 }
 
-//try to get sym link
-func GetContainerRealFilePath(pid int, symlinkPath string) (string, error) {
+func extractProcRootPath(pid int, input string, inTest bool) (string, error) {
+	if inTest {
+		// Since we use ioutil.TempDir("", "proc") to mock proc file system
+		// Regular expression to match the pattern /proc/[number]/root/
+		re := regexp.MustCompile(`.*/proc/\d+/root/`)
+		matches := re.FindStringSubmatch(input)
+
+		if len(matches) == 0 {
+			return "", fmt.Errorf("no match found")
+		}
+		return matches[0], nil
+
+	} else {
+		return fmt.Sprintf("/proc/%d/root", pid), nil
+	}
+
+}
+
+// GetContainerRealFilePath resolves the real file path of a container file from a given symlink. 
+// It handles nested symlinks and detects circular references to prevent infinite loops.
+// Input: pid (process id), symlinkPath (path of the symlink)
+// Output: real file path (string) or an error if the resolution fails.
+func GetContainerRealFilePath(pid int, symlinkPath string, inTest bool) (string, error) {
 	var symlink, procRoot, currentPath, resolvedPath string
 	var err error
+
+	// Flag to check if the current path is under the process root.
+	var underProcRoot bool
+
+	// Keeps track of already visited symlinks to detect loops.
 	visitedSymlink := make(map[string]struct{})
 	currentPath = symlinkPath
 
+	// Nest symlink would look for the the symlink for many times, we limit it to 5 layer of searching of its symlink.
 	layer := 0
 	for ; layer < linkMaxLayers; layer++ {
+		underProcRoot = false
+
 		if _, exists := visitedSymlink[currentPath]; exists {
             return "", fmt.Errorf("Error: Circular symlink detected. The symlink structure creates a loop and cannot be resolved.")
         }
 
-		if !fileExists(currentPath) {
+		if err = fileExists(currentPath); err != nil {
 			log.WithFields(log.Fields{"currentPath": currentPath}).Debug("File not exist")
-			break
+			return "", err
 		}
 
 		if symlink, err = os.Readlink(currentPath); err != nil {
@@ -86,7 +117,15 @@ func GetContainerRealFilePath(pid int, symlinkPath string) (string, error) {
 			return "", err
 		}
 	
-		if procRoot, _ = extractProcRootPath(currentPath); err != nil {
+		// Convert an absolute symlink into a relative path, ensuring that the symlink is processed correctly relative to the current working directory of the function.
+		if filepath.IsAbs(symlink) {
+			if symlink, err = filepath.Rel(filepath.Dir(currentPath), symlink); err != nil {
+				log.WithFields(log.Fields{"error": err}).Debug("Read file relative path")
+				return "", err
+			}
+		}
+
+		if procRoot, err = extractProcRootPath(pid, currentPath, inTest); err != nil {
 			log.WithFields(log.Fields{"error": err}).Debug("Get Proc Root Path fail")
 			return "", err
 		}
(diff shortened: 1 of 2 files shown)


share/osutil/file_linux.go
```diff
@@ -42,14 +43,72 @@ type FileInfoExt struct {
...
+		}
+	}
+
+
+    return "", errors.New("invalid file path")
+}
+
 //try to get sym link
 func GetContainerRealFilePath(pid int, path string) (string, error) {
+	log.WithFields(log.Fields{"pid": pid, "path": path}).Info("in GetContainerRealFilePath")
```
holyspectral, line 100: Better to be `Debug()`

share/osutil/file_linux.go
```diff
@@ -42,14 +43,72 @@ type FileInfoExt struct {
...
+	resultPath := pathParts[len(pathParts)-1]
+	relativePath, err := filepath.Rel(filepath.Dir(symlinkPath), resultPath)
+
+    if err != nil {
+        fmt.Println("Error:", err)
+        return ""
+    }
+
+	return relativePath
+}
```
holyspectral, line 72: It's not a good idea to depend on error message because it relies on c-runtime and locale.  There is no guarantee that c runtime will use exactly the same error message format, so you may get different message from time to time.

share/osutil/file_linux.go
```diff
@@ -42,14 +43,72 @@ type FileInfoExt struct {
...
 		linkPath, err := os.Readlink(path)
 		if err != nil {
-			log.WithFields(log.Fields{"error": err}).Debug("Read file link fail")
-			return "", err
+			if resolvedPath, resolveErr := resolvePath(symlinkPath, err.Error()); resolveErr != nil{
+				log.WithFields(log.Fields{"error": err}).Debug("Read file link fail")
+				return "", err
+			} else {
+				return resolvedPath, nil
+			}
```
holyspectral, line 111: You shouldn't only rely on whether the file exists to determine if it's a correct path. You need to examine its path to make sure it doesn't escape from `/proc/*/root`. For example, the file pointed by this symlink would exist, but it's not correct.  (It will find host's `/etc/passwd` instead of container's. )

```
ln -s ../../../../../etc/passwd /bin/rpm
```

share/osutil/file_linux_test.go
```diff
@@ -0,0 +1,116 @@
...
+	Pid 		  int
+	SymlinkFolder string
+	SymlinkFile   string
+	Symlink 	  string
+	Exist  		  bool
+	ResolvePath   string
+	ExpectResult  string
+}
+
+func PrepareMockFilesMetaData(root string) []MockFile {
```
holyspectral, line 22: Make sure you test circular symlink.

holyspectral (commented):
Overall it's on the right track, but it needs to be polished a little bit more.
1. Examine path instead of check if file exists or not.
2. Add more unit tests to cover circular symlink and `ln -s ../../../../../etc/passwd /bin/rpm`.

share/osutil/file_linux.go
```diff
@@ -42,27 +44,75 @@ type FileInfoExt struct {
...
+			if strings.HasPrefix(resolvedPath, procRoot) && fileExists(resolvedPath) {
+				break
+			}
 		}
-		finfo, err := os.Lstat(path)
-		if err == nil && (finfo.Mode()&os.ModeSymlink) == 0 {
-			return path, nil
+
+		// nest link
+		if strings.HasPrefix(resolvedPath, procRoot) && fileExists(resolvedPath) {
```
holyspectral, line 104: You can keep the last result using a boolean and re-examine the boolean value, so you don't have to check the path and file again,

share/osutil/file_linux.go
```diff
@@ -42,27 +44,75 @@ type FileInfoExt struct {
...
+			if symlink, err = os.Readlink(resolvedPath); err == nil {
+				visitedSymlink[currentPath] = struct{}{}
```
holyspectral, line 106: linkMaxLayers might be enough?

share/osutil/file_linux.go
```diff
@@ -42,27 +44,75 @@ type FileInfoExt struct {
...
+		if procRoot, _ = extractProcRootPath(currentPath); err != nil {
+			log.WithFields(log.Fields{"error": err}).Debug("Get Proc Root Path fail")
+			return "", err
+		}
+	
+		parts := strings.Split(symlink, "/")
+		for i := range parts {
+			partialSymlink := strings.Join(parts[i:], "/")
+			resolvedPath = filepath.Join(filepath.Dir(currentPath), partialSymlink)
+			if strings.HasPrefix(resolvedPath, procRoot) && fileExists(resolvedPath) {
```
holyspectral, line 98: You may want to move the `fileExists()` check to elsewhere.  It's good to check if the file pointed by symlink exists or not, but having it here will affect symlink resolution, given that the file can just be absent.

Say you have a symlink `/bin/rpm -> a/b/../../bin/rpm`, the correct destination is `/bin/bin/rpm` in this case. When the file doesn't exist, your logic may skip that correct path and try `b/../../bin/rpm` and find `/bin/rpm`.  The file could be there, but it's not the file you're supposed to find.

holyspectral (commented):
Overall I think it's great.  Some logics can be simpler and two potential bugs are found, but I'm satisfied with its completeness at this stage.

1. Make sure you add the `filepath.IsAbs(linkPath)` check back and add more absolute path test cases.
2. Make sure the `fileExists()` check does what you expect.

share/osutil/file_linux.go
```diff
@@ -45,48 +45,87 @@ type FileInfoExt struct {
...
+        // File does not exist
+        return err
+    }
+
+    if os.IsPermission(err) {
+        return fmt.Errorf("Permission error accessing file: %s", err)
+    }
+
+    // Other types of errors
+    return err
```
holyspectral, line 65: Sorry if I didn't explain clearly in my last comment.  You can use one of below patterns to implement a function that would return error occasionally:
1. Return `(bool, error)`.  It becomes caller's responsibility to handle the error.
2. Still return `bool` but prints log when an error happens.

share/osutil/file_linux.go
```diff
@@ -42,27 +44,119 @@ type FileInfoExt struct {
...
+		// Convert an absolute symlink into a relative path, ensuring that the symlink is processed correctly relative to the current working directory of the function.
+		if filepath.IsAbs(symlink) {
+			if symlink, err = filepath.Rel(filepath.Dir(currentPath), symlink); err != nil {
+				log.WithFields(log.Fields{"error": err}).Debug("Read file relative path")
+				return "", err
+			}
+		}
+
+		if procRoot, err = extractProcRootPath(pid, currentPath, inTest); err != nil {
+			log.WithFields(log.Fields{"error": err}).Debug("Get Proc Root Path fail")
```
holyspectral, line 129: minor: You can use `WithError(err)`

share/osutil/file_linux.go
```diff
@@ -95,21 +134,26 @@ func GetContainerRealFilePath(pid int, symlinkPath string) (string, error) {
...
+			// nest link
 			if symlink, err = os.Readlink(resolvedPath); err == nil {
 				visitedSymlink[currentPath] = struct{}{}
 				currentPath = resolvedPath;
 			} else {
 				return resolvedPath, nil 
 			}
-		} else {
-			break
 		}
```
holyspectral, line 157: `fileExists()` and `os.Readlink()` are called twice in one single round of this loop.  Any chance to not call them twice?

Note: caller already knows symlinkPath is a symlink and it exists.

share/osutil/file_linux.go
```diff
@@ -42,27 +44,119 @@ type FileInfoExt struct {
...
+			// Assume the first resolved path under proc root is the correct path
+			if strings.HasPrefix(resolvedPath, procRoot) {
+				underProcRoot = true
+				break
+			}
+		}
+
+		if underProcRoot {
+			if err = fileExists(resolvedPath); err != nil {
+				log.WithFields(log.Fields{"resolvedPath": resolvedPath}).Debug("File not exist")
```
holyspectral, line 146: minor: You can use `WithError(err)`

share/osutil/file_linux.go
```diff
@@ -45,48 +45,87 @@ type FileInfoExt struct {
...
 		}
 	
-		if procRoot, _ = extractProcRootPath(currentPath); err != nil {
+		// Convert an absolute symlink into a relative path, ensuring that the symlink is processed correctly relative to the current working directory of the function.
+		if filepath.IsAbs(symlink) {
+			if symlink, err = filepath.Rel(filepath.Dir(currentPath), symlink); err != nil {
+				log.WithFields(log.Fields{"error": err}).Debug("Read file relative path")
+				return "", err
+			}
+		}
```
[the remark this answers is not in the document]
pohanhuangtw: Let follow your idea, separate the abs and symbolic link

share/osutil/file_linux.go
```diff
@@ -42,27 +44,123 @@ type FileInfoExt struct {
...
+					return "", err
+				} else if os.IsPermission(err) {
+					log.WithError(err).Error("Permission error accessing file")
+					return "", err
+				} else {
+					// Others we assume such file exist
+					return resolvedPath, nil 
+				}
+			}
 		}
```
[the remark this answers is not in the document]
pohanhuangtw: Thanks for pointing out the potential issue.

holyspectral:
Generally LGTM.  I can approve it after you fix the indent. (Didn't you have gofmt integrated with your IDE?)

holyspectral (approved):

PR #1133, opened 2023-12-18
NVSHAS-8460  Support newer version of kubernetes cis benchmark

pohanhuangtw:
### Summary

- Revamp the CIS benchmark workflow from pure shell to read the yaml and run with a runner.
- Add unit-test for GetComplianceMeta, let GetComplianceMeta Dynamically read the kubernete / open-shift yaml 
- Support new version cis-1.6.0 / cis-1.23 / cis-1.24 / cis-1.8.0 / openshift-1.4
- Add manual tag, so that we let user know, some of the test has to done by user.
CIS-manual

### How it work
**we do this is to improve the readability of the cis benchmark and the mantainablity**
1. when the enforcer start, if the version is new enough it will trigger a runner shell script
2. runner shell script will handle both worker and master
3. then execute `sh master/worker.sh folder` => the folder represent the cis version to the env
```shell
run_check() {
  local RUN_FOLDER=$1

  find $RUN_FOLDER -type f -name "*.yaml" | sort | while read -r YAML_FILE; do
    # Get the number of groups
    NUM_GROUPS=$(yq e '.groups | length' "$YAML_FILE")

    while [ $group_index -lt $NUM_GROUPS ]; do
        while [ $check_index -lt $NUM_CHECKS ] ; do
            # Read info
            id=$(yq e ".groups[$group_index].checks[$check_index].id" "$YAML_FILE")
            description=$(yq e ".groups[$group_index].checks[$check_index].description" "$YAML_FILE")
            audit=$(yq e ".groups[$group_index].checks[$check_index].audit" "$YAML_FILE")
            profile=$(yq e ".groups[$group_index].checks[$check_index].profile" "$YAML_FILE")
            scored=$(yq e ".groups[$group_index].checks[$check_index].scored" "$YAML_FILE"

            # run the audit here
            eval "$audit"
            check_index=$((check_index + 1))
        done
        group_index=$((group_index + 1))
      done
  done
}

# Input folder
folder=$1

run_check $folder
```
4. the yaml record the audit function and the corresponding remediation.
```yaml
      - id: 1.1.2
        description: Ensure that the API server pod specification file ownership is set to root:root (Automated)
        type: master
        category: kubernetes
        scored: true
        profile: Level 1
        automated: false
        tags:
        - HIPAA
        - NIST
        - PCI
        - GDPR
        audit: |
              check="$id  - $description"
              if [ -f "$file" ]; then
                if [ "$(stat -c %u%g $file)" -eq 00 ]; then
                  pass "$check"
                else
                  warn "$check"
                  warn "      * Wrong ownership for $file"
                fi
              else
                info "$check"
              fi
        remediation: Run the below command (based on the file location on your system) on the master node. For example, chown root:root /etc/kubernetes/manifests/kube-apiserver.yaml
```
5. the result should be the same as usual.


agent/bench.go
jayhuang-suse, line 1: "loadRemediation()"; what if the "filepath.Walk()" fails?

pohanhuangtw: currently not load the yaml, but the yaml is actually get in GetComplianceMeta, based on my understanding.
what do you think?

PR #1159, opened 2024-01-10
NVSHAS-8591 fix watchPlans race condition

holyspectral:
## Why we need this change

A global object, watchPlans, which is accessed by multiple reader and writer go routines, is not protected by a mutex or lock.  This causes random crash and potential unexpected behavior when watching consul changes.

When it crashes, the call stack would be like this:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x1a6bbbf]
```

`add=0x50` shows that one of element retrieved from the map becomes null, despite all items inserted are non-null.

Two POCs are implemented to show that being without protection is the root cause. 
- Reader + Writer: https://go.dev/play/p/K4dq64QWTIy
  - Crash in 30% chance.
- Two writers: https://go.dev/play/p/p-PbvhERzwu
  - Crash in after about 30k attempts. 

Both POC crashed with a similar error:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x493ffa]
```

In this commit, a sync.RWLock is added to protect watchPlans.  Also a staticcheck finding is addressed.

## Test performed
- Scanning nodes and containers. 
- Learned/added new process rules and examined if they have effects.

### DIFF
--- a/share/cluster/consul.go
+++ b/share/cluster/consul.go
@@ -804,8 +804,12 @@ func (m *consulMethod) Transact(entries []transactEntry) (bool, error) {
 
 // Watch related
 var watchPlans []*watch.WatchPlan = make([]*watch.WatchPlan, 0)
+var watchPlansLock sync.RWMutex
 
 func (m *consulMethod) StopAllWatchers() {
+	watchPlansLock.Lock()
+	defer watchPlansLock.Unlock()
+
 	for _, wp := range watchPlans {
 		wp.Stop()
 	}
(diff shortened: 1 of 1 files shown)


share/cluster/consul.go
```diff
@@ -1349,6 +1372,9 @@ func (m *consulMethod) RegisterExistingWatchers() {
 }
 
 func (m *consulMethod) RegisterWatcherMonitor(failFunc func() bool, recoverFunc func()) {
+	watchPlansLock.RLock()
+	defer watchPlansLock.RUnlock()
+
```
jayhuang-suse, line 1377: Move the lock after the debug log. Thus, we can learn what is the last function holds the lock.

PR #1189, opened 2024-01-25
NVSHAS-8532 - Failed to find container's symbolic link at the same folder

pohanhuangtw:
### Summary
1. Fix the symlink issues in a folder, since the folder itself can be another symlink
2. Add unit-test cases on both relative symlink and absolute symlink
3. retryResolveSymlink is a function to iterative resolve the symlink by checking if the prefix of the not exist files can be potential symlink folder, then resolve it (**only call at fail of os.Lstat**)

### How to verify
pohanhuangtw/allinone:8532

### DIFF
--- a/share/osutil/file_linux.go
+++ b/share/osutil/file_linux.go
@@ -81,6 +81,74 @@ func extractProcRootPath(pid int, input string, inTest bool) (string, error) {
 	}
 }
 
+// The retryResolveSymlink splits the path into parts and checks each segment for symlinks, updating the path as it resolves them.
+func retryResolveSymlink(procRoot, prefix, remaining string) (bool, string) {
+    visited := make(map[string]bool)
+    parts := strings.Split(remaining, "/")
+
+    for i, part := range parts {
+        if part == "" {
+            continue
+        }
+
+        // Construct the current path
+        currentPath := filepath.Join(prefix, part)
+		suffix := strings.Join(parts[i+1:], "/")
+
+		var resolvedPath string
+
+		// It could be nest symlink in one folder
+		for {
+			// Check if the current path segment is a symlink
+			fileInfo, err := os.Lstat(currentPath)
+			if err != nil {
+				if os.IsNotExist(err) {
+					// If the path doesn't exist, continue to the next part
+					prefix = currentPath
+					break
+				}
+				// Handle other errors
+				return false, err.Error()
+			}
+
+			if fileInfo.Mode()&os.ModeSymlink != 0 {
+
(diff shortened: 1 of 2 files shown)


share/osutil/file_linux.go
```diff
@@ -139,6 +207,17 @@ func GetContainerRealFilePath(pid int, symlinkPath string, inTest bool) (string,
 
 		visitedSymlink[currentPath] = struct{}{}
 		if underProcRoot {
+			if _, err := os.Lstat(resolvedPath); err != nil {
```
holyspectral, line 210: Check the error first with https://pkg.go.dev/os#IsNotExist ?

share/osutil/file_linux.go
```diff
@@ -81,6 +81,74 @@ func extractProcRootPath(pid int, input string, inTest bool) (string, error) {
 	}
 }
 
+// The retryResolveSymlink splits the path into parts and checks each segment for symlinks, updating the path as it resolves them.
+func retryResolveSymlink(procRoot, prefix, remaining string) (bool, string) {
```
holyspectral, line 85: Can this logic support this case?
- Resolving `/a/b/c` in a container.
- `/a` is a symlink pointing to `/d/e/f`
- `/d/e` is another symlink pointing to `/g/h/i`

The expected result should be `/g/h/i/f/b/c`.  The flow is:

1. `/a/b/c` => `/d/e/f` + `/b/c`
2. `/d/e/f/b/c` => `/g/h/i` + `/f/b/c` => `/g/h/i/f/b/c`

But from your code, it might become `/d/e/f/b/c`?

1. `/a` => `/d/e/f`
2. `/d/e/f` + `/b` => `/d/e/f/b`
3. `/d/e/f/b` + `/c` => `/d/e/f/b/c`

```
# ls -al a 
lrwxrwxrwx 1 root root 6 Jan 25 16:22 a -> /d/e/f
# ls -al /d/e
lrwxrwxrwx 1 root root 6 Jan 25 16:22 /d/e -> /g/h/i
# stat /a/b/c
Device: 20002eh/2097198d	Inode: 58720946    Links: 1
# stat /g/h/i/f/b/c
Device: 20002eh/2097198d	Inode: 58720946    Links: 1
```

pohanhuangtw: Hi @holyspectral ,
Thanks for the review!
You're correct, the correct way is to recursive back to re-run it from beginning.
In new commit, I add a traverse count to limit the number of traverse, and guarantee retry from the beginning.

pohanhuangtw:
Closing this PR due to a decision not to support it.

PR #1211, opened 2024-02-12
NVSHAS-7145: implement no proxy hosts for registry scans

alopez-suse:
Adds a new `NoProxyHosts` field to the both HTTP and HTTPS proxy settings in the system config object. Any host in these lists will NOT use the configured proxy during registry scans.


williamlin-suse (commented):
Shouldn't type RESTRegistry in apis.go contain the new field IgnoreProxy as well?

williamlin-suse:
Is controller/api/apis.yaml updated correctly for this PR?

williamlin-suse:
No, it's not

williamlin-suse (approved):

PR #1228, opened 2024-02-20
NVSHAS-8711 Risk page - only return the filtered assets while advance…

jeffhuang4704:
…d filter was applied


williamlin-suse:
it looks good

PR #1234, opened 2024-02-27
NVSHAS-8286: Suggest the unhandled NV CR objects should be handled ag…

williamlin-suse:
…ain after rolling update thru some mechanism


jeffhuang4704:
this change looks good !

PR #1307, opened 2024-04-18
fix: do not call cancel in all path in newClient()

holyspectral:
In the previous version, `cancel()` is called when leaving `newClient()`.  However, it should be called in `s.cancel()` instead in most of cases.
This PR fixes the issue.


holyspectral:
This issue doesn't exist in main.

PR #1313, opened 2024-04-24
attempt to bump deps to go1.22

Dentrax:
Update outdated deps to latest and use go1.22.


becitsthere:
Thanks for the PR. Upgrade to golang 1.22 is done in dev-5.4 branch. It will be in 5.4 release.

PR #1338, opened 2024-05-14
NVSHAS-7822: federation automation without having to script API calls

williamlin-suse:


idogancay:
Hi @williamlin-suse,

I have a question about federation automation without having to script API calls. 

We aim to automate the installation and configuration of NeuVector as much as possible. Currently, we have automated federation management using API calls, but I read that this is no longer necessary. What replaces this—perhaps a CRD? I can’t seem to find documentation about this, or maybe I’m overlooking it. For example, in the following link, I see how to automate federation using the API. Shouldn’t this be updated?
https://open-docs.neuvector.com/automation/automation#manage-federation-for-master-and-remote-worker-clusters

williamlin-suse:
A documentation case has been submitted to the technical writer.

PR #1358, opened 2024-05-28
[NVSHAS-9009] Move the format prefix logic to scanner.

pohanhuangtw:
### Summary 
Move the process prefix logic to scanner repo.


pohanhuangtw:
The original addition was too complex, so we used the same method to supplement the scanner. Thus close this PR.

PR #1372, opened 2024-05-30
NVSHAS-9024 AdmissionControl Risky Role Perf

jeffhuang4704:


jeffhuang4704:
close for adjust new k8s definitions

PR #1461, opened 2024-07-11
[NVSHAS-9189] Scan will stuck in scheduling after controller is shutdown and restarted

pohanhuangtw:
### Summary

1. Add HealthCheck grpc call to check controller health periodically, if the controller is down, we restart the re-register loop


controller/grpc.go
```diff
@@ -150,6 +150,26 @@ func (ss *ScanService) ScannerRegister(ctx context.Context, data *share.ScannerR
 	}
 }
 
+// HealthCheck checks if the scanner is in the list of controller Consul key-value pairs and if the controller is alive
+// This function follows the logic from how scannerRegister function puts the scanner into the KV store to retrieve it.
+func (ss *ScanService) HealthCheck(ctx context.Context, data *share.ScannerRegisterData) (*share.ScannerAvailable, error) {
+	visible := false
```
holyspectral, line 156: optional: `found` would be more appropriate word choice than `visible` as you need a UI to really see it.

holyspectral (approved):

controller/grpc.go
```diff
@@ -150,6 +150,26 @@ func (ss *ScanService) ScannerRegister(ctx context.Context, data *share.ScannerR
...
+	scannerID := data.ID
+	if data.RPCServer == "127.0.0.1" {
```
williamlin-suse, line 158: it seems like build-in scanner is no longer supported

williamlin-suse (approved):

PR #1478, opened 2024-07-25
NVSHAS-8925: Log/Alert when Federated cluster disconnects

williamlin-suse:
When a user log in NV console, an alert message is generated 
1. if a managed cluster is disconnected for > 1 hour (when user is on primary cluster UI)
2. if the primary cluster is disconnected for > 1 hour (when user is on managed cluster UI)
There is another UI jira case so that user can accept this alert so that this alert won't show up in 1 hour.
After 1 hour, if the disconnection still persist, the alert is re-generated in user's next login.

The 1 hour value can be overridden by env var specified in controller deployment yaml:
```
            - name: TELEMETRY_FREQ
              value: "3"   // minutes

```

### DIFF
--- a/controller/rest/system.go
+++ b/controller/rest/system.go
@@ -2017,6 +2017,37 @@ func handlerSystemGetRBAC(w http.ResponseWriter, r *http.Request, ps httprouter.
 		NvCrdSchemaErrors:        acceptable[4],
 	}
 
+	fedRole := cacher.GetFedMembershipRoleNoAuth()
+	if (fedRole == api.FedRoleMaster && (acc.IsFedReader() || acc.IsFedAdmin() || acc.HasPermFed())) ||
+		(fedRole == api.FedRoleJoint && acc.HasGlobalPermissions(share.PERMS_CLUSTER_READ, 0)) {
+		otherAlerts := map[string]string{}
+		// _fedClusterLeft(206), _fedClusterDisconnected(204)
+		//disconnectedStates := utils.NewSet(_fedClusterLeft, _fedClusterDisconnected)
+		var ids map[string]bool
+		if fedRole == api.FedRoleMaster {
+			ids = cacher.GetFedJoinedClusterIdMap(acc)
+		} else {
+			if m := cacher.GetFedMasterCluster(acc); m.ID != "" {
+				ids = map[string]bool{
+					m.ID: true,
+				}
+			}
+		}
+		if len(ids) > 0 {
+			for id, _ := range ids {
+				s := cacher.GetFedJoinedClusterStatus(id, acc)
+				if elapsed := time.Since(s.LastConnectedTime); s.LastConnectedTime.IsZero() || elapsed > (time.Duration(_teleFreq)*time.Minute) {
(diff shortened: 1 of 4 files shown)


jayhuang-suse (commented):
Will the "SwitchToUnreachable" return "0" again?

williamlin-suse:
> Will the "SwitchToUnreachable" return "0" again?

Yes, as long as the peer side is connected in any occasion/request, SwitchToUnreachable is set to 0 & LastConnectedTime is set to now.
It means the disconnection needs to continuously last for > 1 hour so that the alert is generated when the user login in UI

jeffhuang4704:
It looks like we will return the message directly to the user. Since NeuVector has a localized version, should we consider this in our case?

jeffhuang4704 (commented):
It looks like we will return the message directly to the user. Since NeuVector has a localized version, should we consider this in our case?

williamlin-suse:
Controller doesn't support L10N/I18N yet.
NV's localized(cn) version is only for manager image

controller/rest/system.go
```diff
@@ -2017,6 +2017,45 @@ func handlerSystemGetRBAC(w http.ResponseWriter, r *http.Request, ps httprouter.
...
+			}
+		}
+		if len(ids) > 0 {
+			for id, _ := range ids {
+				s := cacher.GetFedJoinedClusterStatus(id, acc)
+				if elapsed := time.Since(s.LastConnectedTime); s.LastConnectedTime.IsZero() || elapsed > (time.Duration(_teleFreq)*time.Minute) {
+					var alert string
+					if fedRole == api.FedRoleMaster {
+						clusterInfo := cacher.GetFedJoinedCluster(id, acc)
+						alert = fmt.Sprintf("Managed cluster %s is disconnected from primary cluster", clusterInfo.Name)
```
holyspectral, line 2043: minor: The same string shows up in both rest/system.go and rest/auth.go.  Given that these string have to be the same for the acceptedAlerts to work, maybe make the string as a constant?

controller/rest/system.go
```diff
@@ -2017,6 +2017,45 @@ func handlerSystemGetRBAC(w http.ResponseWriter, r *http.Request, ps httprouter.
...
+					} else {
+						alert = "This cluster is disconnected from primary cluster"
+					}
+					b := md5.Sum([]byte(alert))
+					key := hex.EncodeToString(b[:])
```
holyspectral, line 2048: Consider: md5 in this case is acceptable, but md5 can be flagged by a few code scan tools, which is probably not what we want.  Maybe use sha1 or sha256 next time?

holyspectral (approved):
LGTM.  I will be adding similar alerts soon so this PR is very helpful.

Just two comments that we can revisit later.

controller/rest/system.go
holyspectral: Sounds good.  I will look into it.  Feel free to resolve this comment if you want.  :-)

controller/rest/system.go
williamlin-suse: use getFedDisconnectAlert() to generate the alert/key

PR #1484, opened 2024-07-31
NVSHAS-9225: Do not allow UI on 5.4 master cluster to switch to pre-5…

williamlin-suse:
….4 managed clusters because of REST API changes


jeffhuang4704 (approved):
it looks good to me.

PR #1486, opened 2024-07-31
NVSHAS-8539 proxy password lost

holyspectral:
### Problem
Controller doesn't check if password is null or not before overwriting the existing value.

### Solution

This commit provides an API update for UI to
resolve this issue.

### Tests performed

Slack integration with tinyproxy.

### DIFF
--- a/controller/rest/system.go
+++ b/controller/rest/system.go
@@ -330,6 +330,16 @@ func handlerSystemGetConfigBase(apiVer string, w http.ResponseWriter, r *http.Re
 						RegistryHttpsProxyEnable: rconf.RegistryHttpsProxyEnable,
 						RegistryHttpProxy:        rconf.RegistryHttpProxy,
 						RegistryHttpsProxy:       rconf.RegistryHttpsProxy,
+						RegistryHttpProxyCfg: api.RESTProxyConfig{
+							URL:      &rconf.RegistryHttpProxy.URL,
+							Username: &rconf.RegistryHttpProxy.Username,
+							Password: &rconf.RegistryHttpProxy.Password,
+						},
+						RegistryHttpsProxyCfg: api.RESTProxyConfig{
+							URL:      &rconf.RegistryHttpsProxy.URL,
+							Username: &rconf.RegistryHttpsProxy.Username,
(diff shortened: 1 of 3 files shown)


jeffhuang4704:
looks good to me.

holyspectral:
@williamlin-suse @jeffhuang4704 The implementation has been updated per our discussion.  Please help to re-review.  Thanks!

controller/rest/system.go
```diff
@@ -330,6 +330,16 @@ func handlerSystemGetConfigBase(apiVer string, w http.ResponseWriter, r *http.Re
 						RegistryHttpsProxyEnable: rconf.RegistryHttpsProxyEnable,
 						RegistryHttpProxy:        rconf.RegistryHttpProxy,
 						RegistryHttpsProxy:       rconf.RegistryHttpsProxy,
+						RegistryHttpProxyCfg: api.RESTProxyConfig{
```
williamlin-suse, line 333: handlerSystemGetConfigBase() is for GET("/v1/system/config") & GET("/v2/system/config").
Why add fields RegistryHttpProxyCfg & RegistryHttpsProxyCfg (both look like for PATCH purpose) in RESTSystemConfigProxyV2 which is for GET?

holyspectral: yeah that's correct.  I added it to GET because it would be more consistent with REST concept, where every data is a resource.  Any concern about it?  The data in consul remains unchanged.

holyspectral: For example, when I’m calling PATCH /v2/system/config, I would expect that the field I’m patching also exists in its GET response.  Hope this clarifies things a bit.

williamlin-suse: I think this is the only type among all response types for GET requests that has this kind of fields.
So RegistryHttpProxyCfg/RegistryHttpsProxyCfg will always have the same value as RegistryHttpProxy/RegistryHttpsProxy in the response json string, correct?

holyspectral: Yes, that’s correct. It’s purely for REST API.

jeffhuang4704 (approved):
looks good to me.

williamlin-suse (approved):

PR #1487, opened 2024-08-02
[NVSHAS-8799] Fix the compliance profile filter when filter V1 result.

pohanhuangtw:
### Problem
When we set up the compliance profile, the result is not aligned in node > compliance.

### Test perform
Set up the compliance profile, the result is aligned in node > compliance.

### DIFF
--- a/controller/rest/bench.go
+++ b/controller/rest/bench.go
@@ -34,9 +34,9 @@ func bench2REST(bench share.BenchType, item *share.CLUSBenchItem, cpf *complianc
 		// update the Tags with compliance profile
 		// if tagVersion == V2, return with TagV2: map[string]share.TagDetails{}, otherwise Tag: []string
 		if tagVersion == scanUtils.V2 {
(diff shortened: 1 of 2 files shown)


controller/rest/bench.go
```diff
@@ -49,7 +49,7 @@ func bench2REST(bench share.BenchType, item *share.CLUSBenchItem, cpf *complianc
 			}
 		} else {
 			if _, ok := cpf.filter[r.TestNum]; ok {
-				r.Tags = metaMap[r.TestNum].Tags
+				r.Tags = cpf.filter[r.TestNum]
```
williamlin-suse, line 52: why not just
```
                        if tags, ok := cpf.filter[r.TestNum]; ok {
				r.Tags = tags
			}
```
? Same for lines 37/39 & 90/91 (like in lines 83/85)

pohanhuangtw: you have a point

williamlin-suse (approved):

PR #1488, opened 2024-08-03
NVSHAS-7853 TLS handshake EOF

holyspectral:
When controller's service is configured as LoadBalancer, LoadBalancer can perform health check on the endpoint and generate a lot of error messages in controller:

```
http: TLS handshake error from xxxxxx: EOF
```
This commit skips this kind of error by providing a different logger for controller's REST server.

### DIFF
--- a/controller/rest/rest.go
+++ b/controller/rest/rest.go
@@ -1937,6 +1937,7 @@ func StartRESTServer(isNewCluster bool, isLead bool) {
 		// ReadTimeout:  time.Duration(5) * time.Second,
 		// WriteTimeout: time.Duration(35) * time.Second,
 		TLSNextProto: make(map[string]func(*http.Server, *tls.Conn, http.Handler), 0), // disable http/2
+		ErrorLog:     newHttpServerErrorWriter(),
 	}
 	for {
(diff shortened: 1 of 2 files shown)


holyspectral:
Verified that errors other than EOF will still be printed.
```
2024/08/03 02:10:48 http: TLS handshake error from 192.168.58.101:42412: local error: tls: bad record MAC
2024/08/03 02:10:52 http: TLS handshake error from 192.168.58.101:42420: local error: tls: bad record MAC
```

controller/rest/rest.go
```diff
@@ -1937,6 +1937,7 @@ func StartRESTServer(isNewCluster bool, isLead bool) {
 		// ReadTimeout:  time.Duration(5) * time.Second,
 		// WriteTimeout: time.Duration(35) * time.Second,
 		TLSNextProto: make(map[string]func(*http.Server, *tls.Conn, http.Handler), 0), // disable http/2
+		ErrorLog:     newHttpServerErrorWriter(),
```
williamlin-suse, line 1940: Should the server := &http.Server{} object in startFedRestServer() have the same fix?

williamlin-suse (approved):

PR #1491, opened 2024-08-07
feat: NVSHAS-9287 enable internal cert rotation

holyspectral:
## NVSHAS-9287 enable internal cert rotation

1. Refactored post-sync and added more unit-tests. 
2. Provide a new option enable-rotation, which allows rotation from old deploymemt.  The option is off by default.  For fresh install, new certificate will always be created.
3. Update informer, so in upgrade case, old instances can continue using built-in certs.

### New behavior

Now internal.autoGenerateCert is enabled by default and a new option internal.autoRotateCert is added, which is off by default.

- When internal.autoGenerateCert is enabled, init container and neuvector-cert-upgrader job will run and generate certificate based on settings. Informers in each component will load these certificates.
- When internal.autoRotateCert is enabled, neuvector-cert-upgrader job will attempt to rotate certificate if it's already created previously.

In a nutshell:

- internal.autoGenerateCert + internal.autoRotateCert => The current behavior when internal.autoGenerateCert is enabled.
- internal.autoGenerateCert only => certificate will only be generated in fresh install.
- internal.autoRotateCert only or none is enabled => pre-5.4 behavior. No certificate is generated and built-in cert will be used.


jeffhuang4704 (approved):
looks good to me.

PR #1493, opened 2024-08-08
NVSHAS-8562: Add internal cert expiration notification

kyledong-suse:
1. Add a new rest API "/v1/system/alerts"
2. This API includes existing RBAC alerts and new internal cert expiration alerts


controller/rest/rest.go
```diff
@@ -2246,3 +2247,33 @@ func getRequestApiVersion(r *http.Request) ApiVersion {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to parse certificate: %v", err)
+	}
+
+	// Check the expiration date
+	if time.Now().After(cert.NotAfter.Add(-expireThreshold)) {
+		log.WithFields(log.Fields{
+			"expiry":    cert.NotAfter,
+			"threshold": expireThreshold,
+		}).Info("Nearly expired certificate is detected")
```
holyspectral, line 2274: minor: unlike the scenario in cert-upgrader, the function here would be called from time to time.  Maybe consider changing it to Debug level?

controller/rest/system.go
```diff
@@ -2116,6 +2117,223 @@ func handlerSystemGetRBAC(w http.ResponseWriter, r *http.Request, ps httprouter.
...
 }
 
+func getAlertGroup(alerts []string, alertType api.AlertType, acceptedAlerts utils.Set) (*api.RESTNvAlertGroup, error) {
+	alertGroup := &api.RESTNvAlertGroup{
+		Type: alertType,
+	}
+
+	if len(alerts) > 0 {
+		for _, alert := range alerts {
+			b := md5.Sum([]byte(alert))
```
holyspectral, line 2127: Can you change it to sha256?

controller/rest/system.go
```diff
@@ -2116,6 +2117,223 @@ func handlerSystemGetRBAC(w http.ResponseWriter, r *http.Request, ps httprouter.
...
+				})
+			}
+		}
+	}
+
+	if len(alertGroup.Data) > 0 {
+		return alertGroup, nil
+	}
+
+	return alertGroup, errors.New("no alert is retrieved")
```
holyspectral, line 2142: Do we consider the case with no alerts as an error?  How about this?
```suggestion
	return nil, nil
```

controller/rest/system.go
```diff
@@ -2116,6 +2117,223 @@ func handlerSystemGetRBAC(w http.ResponseWriter, r *http.Request, ps httprouter.
...
+}
+
+func getInternalCertExpireAlert(certFilePath string) (string, error) {
+	expireThresholds := []time.Duration{30 * 24 * time.Hour, 90 * 24 * time.Hour, 180 * 24 * time.Hour}
```
holyspectral, line 2146: minor: How about using #day instead of time.Duration, so we can format the alert directly in the switch case below?  For example:

```
expireThreholdDays := []int{30, 90, 180}
...
for _, expireThresholdDay := range expireThreholdDays {
    certExpiredMsg = fmt.Sprintf("Internal certificate will be expired in %d day(s)", expireThresholdDay)
}
```

controller/rest/system.go
```diff
@@ -2116,6 +2117,223 @@ func handlerSystemGetRBAC(w http.ResponseWriter, r *http.Request, ps httprouter.
...
+			if strings.Contains(certFilePath, "ca.cert") {
+				certExpiredMsg = "Internal CA certificate will be expired in 180 days."
+			} else {
+				certExpiredMsg = "Internal certificate will be expired in 180 days."
+			}
+		}
+		return certExpiredMsg, nil
+	}
+
+	return "", errors.New("no nearly expired internal certificate is detected")
```
holyspectral, line 2181: Maybe it's better to consider the case where the code runs successfully but no alert is found as a success?

controller/rest/system.go
```diff
@@ -2116,6 +2117,223 @@ func handlerSystemGetRBAC(w http.ResponseWriter, r *http.Request, ps httprouter.
...
+}
+
+func handlerSystemGetAlerts(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
holyspectral, line 2184: How about refactoring `handlerSystemGetRBAC()` a little bit, so `handlerSystemGetAlerts()` can go through the same path as `handlerSystemGetRBAC()` to get the existing alerts?

holyspectral (commented):
Thanks for working on this!  Please see my comments inline.

controller/rest/rest.go
kyledong-suse: fixed

controller/rest/system.go
kyledong-suse: changed

controller/rest/system.go
kyledong-suse: I was thinking to use error to filter the case to pass the alertGroup with data. But it makes more sense to me to return nil, and in this case, we don't even need error :P

controller/rest/system.go
kyledong-suse: changed as above

controller/rest/system.go
kyledong-suse: changed

controller/rest/system.go
kyledong-suse: Add couple of helper functions. handlerSystemGetRBAC() and handlerSystemGetAlerts() can use those helpers to keep them retrieve existing alerts in the same way.

controller/rest/rest.go
```diff
@@ -1604,6 +1604,7 @@ func StartRESTServer(isNewCluster bool, isLead bool) {
 	r.GET("/v1/system/config", handlerSystemGetConfig)   // supported 'scope' query parameter values: ""(all, default)/"fed"/"local". no payload
 	r.GET("/v2/system/config", handlerSystemGetConfigV2) // supported 'scope' query parameter values: ""(all, default)/"fed"/"local". no payload. starting from 5.0, rest client should call this api.
 	r.GET("/v1/system/rbac", handlerSystemGetRBAC)       // skip API document
+	r.GET("/v1/system/alerts", handlerSystemGetAlerts)   // skip API document
```
holyspectral, line 1607: Since this is going to be a public API, you probably want to include API docs in this case.  
```suggestion
	r.GET("/v1/system/alerts", handlerSystemGetAlerts)
```

controller/api/internal_apis.go
```diff
@@ -224,6 +224,39 @@ type RESTK8sNvAcceptableAlerts struct {
...
+	CertificateAlerts        *RESTNvAlertGroup `json:"certificate_alerts,omitempty"`
+	OtherAlerts              *RESTNvAlertGroup `json:"other_alerts,omitempty"`
+}
+
+type AlertType string
+
+const (
+	RBAC           AlertType = "RBAC"
+	TlsCertificate AlertType = "TLS_CERTIFICATE"
+)
```
holyspectral, line 248: Optional: Add a prefix or suffix so these constants can be distinguished from other types?

controller/rest/system.go
```diff
@@ -2100,22 +2067,218 @@ func handlerSystemGetRBAC(w http.ResponseWriter, r *http.Request, ps httprouter.
...
+	}
+
+	// populate acceptable_alerts
+	clusterRoleAlerts, clusterRoleBindingAlerts, roleAlerts, roleBindingAlerts, nvCrdSchemaAlerts, otherAlerts, acceptedAlerts := getAcceptableAlerts(acc, login)
+	resp.AcceptableAlerts = &api.RESTNvAcceptableAlerts{}
+	if clusterRoleAlertGroup := getAlertGroup(clusterRoleAlerts, api.RBAC, acceptedAlerts); clusterRoleAlertGroup != nil {
+		resp.AcceptableAlerts.ClusterRoleAlerts = clusterRoleAlertGroup
+	}
+	if clusterRoleBindingAlertGroup := getAlertGroup(clusterRoleBindingAlerts, api.RBAC, acceptedAlerts); clusterRoleBindingAlertGroup != nil {
+		resp.AcceptableAlerts.ClusterRoleAlerts = clusterRoleBindingAlertGroup
```
holyspectral, line 2220:
```suggestion
		resp.AcceptableAlerts.ClusterRoleBindingAlerts = clusterRoleBindingAlertGroup
```

holyspectral (commented):
Thanks,  It generally looks good to me depsite a c&p issue spotted.  
Please also add API docs (controller/api/apis.yaml) and consider adding some unit tests.

controller/rest/system.go
kyledong-suse: Thanks for catching it! Fixed

controller/api/internal_apis.go
kyledong-suse: Changed it to "RESTNvAlertType" to make it consistent.

controller/rest/rest.go
kyledong-suse: change it to public interface and document it

holyspectral (approved):
LGTM

PR #1495, opened 2024-08-12
NVSHAS-9298: Add remote_global_permissions info in GET(v1/selfuser) r…

williamlin-suse:
…esponse payload


jeffhuang4704 (approved):
it looks good to me.

PR #1496, opened 2024-08-13
[NVSHAS-9270] Support k3s for cis benchmark pipeline

pohanhuangtw:
### Problem
CIS benchmark is not support k3s before.

### Solution
- support CIS benchmark k3s
- Read the journalctl for the command processing. (use nstools)
- add `isKubeProcess` to detect the executable of k3s not proc name, since the proc name could change over time, and I keep the original logic for backward compatible.

### Test perform
Set up the k3s env, and make sure the other env works as well.

### DIFF
--- a/agent/bench.go
+++ b/agent/bench.go
@@ -84,8 +84,6 @@ const (
 	containerTimerStart = time.Second * 10
 	kubeTimerStart      = time.Second * 10
 	scriptTimerStart    = time.Second * 1
-	cmdK3sServer        = "k3s-server"
-	cmdK3sAgent         = "k3s-agent"
 	cmdKubeApiServer    = "kube-apiserver"
 	cmdKubeManager      = "kube-controller-manager"
 	cmdKubeScheduler    = "kube-scheduler"
@@ -562,30 +560,31 @@ func (b *Bench) RerunKube(cmd, cmdRemap string, forced bool) {
 	}
 
 	// Check if the node is a Kubernetes master
-	_, b.isKubeMaster = b.kubeCisCmds[cmdKubeApiServer]
+	_, isKubeMaster := b.kubeCisCmds[cmdKubeApiServer]
 
 	// Check if the node is a Kubernetes worker
-	_, b.isKubeWorker = b.kubeCisCmds[cmdKubelet]
+	_, isKubeWorker := b.kubeCisCmds[cmdKubelet]
 
 	k8sVer, _ := global.ORCH.GetVersion(false, false)
 
-	// Only fall back to K3s detection if the node is neither a master nor a worker
-	if !b.isKubeMaster && !b.isKubeWorker && strings.Contains(k8sVer, "k3s") {
+	// Only fall back to K3s detection if the node is neither a master nor a worker.
+	// if k3s is both server and agent, NeuVector treat it as server
+	if !isKubeMaster && !isKubeWorker && strings.Contains(k8sVer, "k3s") {
 		// use dir := "/proc/1/root/var/lib/rancher/k3s/server" to check if in k3s master
+		// On K3s by default, all servers are also agents.
 		dir := "/proc/1/root/var/lib/rancher/k3s/server"
 		if _, err := os.Stat(dir); os.IsNotExist(err) {
-			b.isKubeWorker = true
-			b.kubeCisCmds[cmdKubelet] = cmdKubelet
+			isKubeWorker = true
 			b.isK3s = true
 		} else if err == nil {
-			b.isKubeMaster = true
-			b.kubeCisCmds[cmdKubeApiServer] = cmdKubeApiServer
+			isKubeMaster = true
 			b.isK3s = true
 		} else {
 			log.WithFields(log.Fields{"error": err, "directory": dir}).Error("Error checking directory")
 		}
 	}
-
+	b.isKubeMaster = isKubeMaster
+	b.isKubeWorker = isKubeWorker
 	var sched bool
 
 	if b.isKubeMaster {
@@ -1343,7 +1342,10 @@ func (b *Bench) runKubeBench(bench share.BenchType, script, remediationFolder st
 	var config string
 
 	if b.cisYAMLMode {
-		if bench == share.BenchKubeMaster {
+		// On K3s by default, all servers are also agents, thus we let it reads all yamls
+		if bench == share.BenchKubeMaster && b.isK3s {
+			config = remediationFolder
+		} else if bench == share.BenchKubeMaster {
 			config = remediationFolder + "master"
 		} else if bench == share.BenchKubeWorker {
 			config = remediationFolder + "worker"
--- a/agent/bench.go
+++ b/agent/bench.go
@@ -571,12 +571,18 @@ func (b *Bench) RerunKube(cmd, cmdRemap string, forced bool) {
 
 	// Only fall back to K3s detection if the node is neither a master nor a worker
 	if !b.isKubeMaster && !b.isKubeWorker && strings.Contains(k8sVer, "k3s") {
-		if _, b.isKubeMaster = b.kubeCisCmds[cmdK3sServer]; b.isKubeMaster {
-			b.kubeCisCmds[cmdKubeApiServer] = cmdKubeApiServer
-			b.isK3s = true
-		} else if _, b.isKubeWorker = b.kubeCisCmds[cmdK3sAgent]; b.isKubeWorker {
+		// use dir := "/proc/1/root/var/lib/rancher/k3s/server" to check if in k3s master
+		dir := "/proc/1/root/var/lib/rancher/k3s/server"
+		if _, err := os.Stat(dir); os.IsNotExist(err) {
+			b.isKubeWorker = true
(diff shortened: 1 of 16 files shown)


agent/bench.go
```diff
@@ -549,7 +562,20 @@ func (b *Bench) RerunKube(cmd, cmdRemap string, forced bool) {
 	}
 
 	_, b.isKubeMaster = b.kubeCisCmds[cmdKubeApiServer]
+	if !b.isKubeMaster {
+		if _, b.isKubeMaster = b.kubeCisCmds[cmdK3sServer]; b.isKubeMaster {
```
holyspectral, line 566: What is the command in your k3s environment?  I installed my instance following [the quick start guide](https://docs.k3s.io/quick-start), and my command line looks like this:

```
root        2061       1 10 09:22 ?        00:29:13 /usr/local/bin/k3s agent
```

```
$ /usr/local/bin/k3s -v
k3s version v1.29.6+k3s2 (b4b156d9)
go version go1.21.11
```

Not sure if it would be supported in your case.

agent/bench.go
```diff
@@ -1271,8 +1339,8 @@ func (b *Bench) runKubeBench(bench share.BenchType, script, remediationFolder st
 			config = remediationFolder + "worker"
 		}
 	}
-
-	cmd = exec.Command("sh", script, config)
+	// Read the journal string for K3s to check if some of the command arguments are set correctly.
+	cmd = exec.Command("sh", script, config, journalStr)
```
holyspectral, line 1343: Question:  maybe it's a little like a corner case, but if the host has both k3s and k8s running, what should the result be?

holyspectral (commented):
Thanks for working on this!  I have a few questions.

agent/bench.go
pohanhuangtw: the journalStr will only be used in k3s cases, the k8s can ignore such input.

pohanhuangtw: or you think it will be better if we separate like the followiing?
```go
	var journalStr string
	var journalErr error
	if b.isK3s {
		journalStr, journalErr = b.readJournal()
		if journalErr != nil {
			log.WithFields(log.Fields{"error": journalErr}).Error("Error while read the journal")
		}
		cmd = exec.Command("sh", script, config, journalStr)
	} else {
	        cmd = exec.Command("sh", script, config)
	}
```

holyspectral: It's more like a general question.  I should have put the comment elsewhere.  Sorry for the confusion.

My understanding is that when enforcer detects a new program running and the program meets a certain criteria, it will trigger the kubebench scan.  

So my question is, when both `kube-apiserver` and `k3s-agent` run at the same node, I think k3s will take the precedence.  Is that right?

pohanhuangtw: yes, if both kube-apiserver and k3s-agent run at the same node, it will run wih normal k8s not k3s cis benchmark

agent/bench.go
pohanhuangtw: Hi @holyspectral ,
I add `isKubeProcess` to detect the executable of k3s not proc name, since the proc name could change over time, and I keep the original logic for backward compatible.

agent/bench.go
```diff
@@ -548,9 +561,31 @@ func (b *Bench) RerunKube(cmd, cmdRemap string, forced bool) {
...
+			b.kubeCisCmds[cmdKubelet] = cmdKubelet
+			b.isK3s = true
+		} else if err == nil {
+			b.isKubeMaster = true
+			b.kubeCisCmds[cmdKubeApiServer] = cmdKubeApiServer
+			b.isK3s = true
+		} else {
+			log.WithFields(log.Fields{"error": err, "directory": dir}).Error("Error checking directory")
+		}
+	}
```
holyspectral, line 587: I like the way you get k8s version to make sure it's a k3s.
1. With false and false in GetVersion's argument, do you get a valid `k8sVer`?
2. minor: As a better practice, you probably should leave `b.isKubeWorker` and `b.isKubeMaster` intact (line 564-568) until you make the final decision about its platform?
3. Can k3s have both server and worker on the same node?

pohanhuangtw: Hi @holyspectral ,
Thanks for review.

1. With false and false in GetVersion's argument, do you get a valid k8sVer? yes
2. minor: As a better practice about this, I think line 573-586 shows it's k3s, thus this should be fine?
3. if this is a single node cluster, then it's both server and worker node.

pohanhuangtw: about 3, if this, I think we should have another case for this?

holyspectral: 2 is more about coding style.  When you update `b.isKubeMaster` and `b.isKubeWorker` before the logic completes, it's prone to leave the two fields in an unexpected state if the error handling is not implemented correctly.  One example is when you have to exit the function with an unrecoverable error. 

A better way is like this:
```
	// Keep the variables in local variables for now,
	_, isKubeMaster := b.kubeCisCmds[cmdKubeApiServer]
	_, isKubeWorker := b.kubeCisCmds[cmdKubelet]
	k8sVer, _ := global.ORCH.GetVersion(false, false)

	// Only fall back to K3s detection if the node is neither a master nor a worker
	if !isKubeMaster && !isKubeWorker && strings.Contains(k8sVer, "k3s") {
            // determine the value for k3s.
	}
	
        // assign the value
	b.isKubeMaster = isKubeMaster
	b.isKubeWorker = isKubeWorker
	b.isK3s = isK3s
```

So even if the function has to exit early, the state in this struct won't be changed.  

For 3, agree.  We can get this merged first.

pohanhuangtw: Hi @holyspectral ,
Just found the github question about if the k3s node can be server and agent at the same time. https://github.com/k3s-io/k3s/issues/1279
I think we can resolve it in this PR as well

What do you think?

holyspectral:
Yeah it sounds good to me. On Aug 15, 2024, at 9:10 PM, pohanhuangtw ***@***.***> wrote:﻿
@pohanhuangtw commented on this pull request.



In agent/bench.go:
> +	// Check if the node is a Kubernetes master
 	_, b.isKubeMaster = b.kubeCisCmds[cmdKubeApiServer]
+
+	// Check if the node is a Kubernetes worker
 	_, b.isKubeWorker = b.kubeCisCmds[cmdKubelet]
 
+	k8sVer, _ := global.ORCH.GetVersion(false, false)
+
+	// Only fall back to K3s detection if the node is neither a master nor a worker
+	if !b.isKubeMaster && !b.isKubeWorker && strings.Contains(k8sVer, "k3s") {
+		// use dir := "/proc/1/root/var/lib/rancher/k3s/server" to check if in k3s master
+		dir := "/proc/1/root/var/lib/rancher/k3s/server"
+		if _, err := os.Stat(dir); os.IsNotExist(err) {
+			b.isKubeWorker = true
+			b.kubeCisCmds[cmdKubelet] = cmdKubelet
+			b.isK3s = true
+		} else if err == nil {
+			b.isKubeMaster = true
+			b.kubeCisCmds[cmdKubeApiServer] = cmdKubeApiServer
+			b.isK3s = true
+		} else {
+			log.WithFields(log.Fields{"error": err, "directory": dir}).Error("Error checking directory")
+		}
+	}

Hi @holyspectral ,
Just found the github question about if the k3s node can be server and agent at the same time. k3s-io/k3s#1279
I think we can resolve it in this PR as well
What do you think?

agent/bench.go
holyspectral: Yeah it sounds good to me.

pohanhuangtw: please take a look, thanks

holyspectral (approved):
It looks good to me.  I didn't review those yaml files thoroughly, so please make sure at least someone has gone through it.  

I'm not 100% sure if k3s can run with worker only mode as they said `by default`.  Please confirm about it.

> On K3s by default, all servers are also agents.
https://github.com/k3s-io/k3s/issues/1279

Good job!

pohanhuangtw:
> It looks good to me. I didn't review those yaml files thoroughly, so please make sure at least someone has gone through it.
> 
> I'm not 100% sure if k3s can run with worker only mode as they said `by default`. Please confirm about it.
> 
> > On K3s by default, all servers are also agents.
> > [k3s-io/k3s#1279](https://github.com/k3s-io/k3s/issues/1279)
> 
> Good job!

Hi @holyspectral ,
Thanks for review, the goal for checking the node is worker node or not is mainly for running the kublet check.
In F&Q in [k3s doc](https://docs.k3s.io/architecture) show, Both servers and agents run the kubelet, container runtime, and CNI. See the [Advanced Options](https://docs.k3s.io/advanced#running-agentless-servers-experimental) documentation for more information on running agentless servers.
Thus, I think it matches the current behavior.

For the scripts, @songlongcn would you mind take a look at this, I follow the [k3s security ](https://docs.k3s.io/security/self-assessment-1.8)to implement it.

songlongcn (approved):
I have reviewed the k3s audit YAML files. The content aligns well with the guidelines provided in [K3s Self-Assessment](https://docs.k3s.io/security/self-assessment-1.8), and the YAML files are well formatted.

PR #1498, opened 2024-08-13
NVSHAS-9301: NV deployed on Rancher Prime cannot tell it's Rancher flavor

williamlin-suse:


jeffhuang4704 (approved):
it looks good to me.

PR #1499, opened 2024-08-13
fix: remove AWS keys in API docs

holyspectral:
I'm getting a lot of this warning when pushing into my fork:

```
remote: - GITHUB PUSH PROTECTION
remote:   —————————————————————————————————————————
remote:     Resolve the following violations before pushing again
remote: 
remote:     - Push cannot contain secrets
```

This commit removes the invalid AWS keys which trigger the alerts.


jeffhuang4704 (approved):
it looks good to me.

PR #1502, opened 2024-08-14
fix: NVSHAS-9289 allow upgrade when RBAC missed

holyspectral:
This PR contains two parts of changes:

### Allow NV to start when RBAC settings are out-of date

It's possible that users have out-of-date RBAC settings in their environment, for example, these settings are maintained by other team.  Without proper RBAC settings, NV components can fail to start.

This commit fixes the issue by checking required RBAC permission and falling back if the check fails.

### Generate alerts when required RBAC settings not not present

When the resources are missing, controller will remind users by generating alerts.

### DIFF
--- a/controller/resource/kubernetes_resource.go
+++ b/controller/resource/kubernetes_resource.go
@@ -51,6 +51,7 @@ const (
 	K8sApiVersionV1Beta1          = "v1beta1"
 	K8sApiVersionV1Beta2          = "v1beta2"
 	K8sResCronjobs                = "cronjobs"
+	K8sResCronjobsFinalizer       = "cronjobs/finalizers"
 	K8sResDaemonsets              = "daemonsets"
 	K8sResDeployments             = "deployments"
 	K8sResDeploymentConfigs       = "deploymentconfigs"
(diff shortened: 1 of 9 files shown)


holyspectral:
image

controller/resource/kubernetes_resource.go
```diff
@@ -81,34 +82,38 @@ const (
...
+	nvCrdCompProfileRole         = "neuvector-binding-nvcomplianceprofiles"
+	nvCrdCompProfileRoleBinding  = nvCrdCompProfileRole
+	NvScannerRole                = "neuvector-binding-scanner"
+	NvScannerRoleBinding         = NvScannerRole
+	NvSecretRole                 = "neuvector-binding-secret"
+	nvSecretRoleBinding          = NvSecretRole
+	NvAdminRoleBinding           = "neuvector-admin"
+	nvViewRoleBinding            = "neuvector-binding-view"
+	NvJobCreationRole            = "neuvector-binding-job-creation"
+	NvJobCreationRoleRoleBinding = NvJobCreationRole
```
williamlin-suse, line 114: Is it a typo? NvJobCreationRoleRoleBinding or NvJobCreationRoleBinding ?

williamlin-suse (approved):

PR #1508, opened 2024-08-16
NVSHAS-9292: Fix Ingress Egress esposure shows 0 Vulnerabilities

kyledong-suse:
1. Fix Ingress Egress esposure shows 0 Vulnerabilities
2. Rename AGE_SYSLOG_LEVEL to ENF_SYSLOG_LEVEL to make naming consistent

### DIFF
--- a/controller/cache/compliance.go
+++ b/controller/cache/compliance.go
@@ -199,6 +199,12 @@ type wlMini struct {
 	mode string
 }
 
+type podVuls struct {
+	CriticalVuls int
+	HighVuls int
+	MedVuls int
+}
+
 func (m CacheMethod) GetRiskScoreMetrics(acc, accCaller *access.AccessControl) *api.RESTInternalSystemData {
 	var s api.RESTRiskScoreMetrics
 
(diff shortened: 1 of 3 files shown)


controller/cache/compliance.go
```diff
@@ -275,6 +310,11 @@ func (m CacheMethod) GetRiskScoreMetrics(acc, accCaller *access.AccessControl) *
 
 		if wl.ShareNetNS == "" {
 			epMap[cache.workload.ID] = &wlMini{mode: mode}
+			if _, ok := podVulsMap[cache.podName]; ok {
```
williamlin-suse, line 313: why not use 
`if podVuls, ok := podVulsMap[cache.podName]; ok`
so that podVuls is referenced in the following 3 lines?

controller/cache/compliance.go
```diff
@@ -246,6 +252,35 @@ func (m CacheMethod) GetRiskScoreMetrics(acc, accCaller *access.AccessControl) *
...
+		if cache.platformRole != "" && disableSystem {
+			// skip system containers
+			continue
+		}
+		if !wl.Running || cache.scanBrief == nil {
+			continue
+		}
+
+		// do not re-sum existing pod vuls
+		if wl.ShareNetNS != "" {
```
williamlin-suse, line 274: here it's
```
if wl.ShareNetNS != "" {
   // put podVulsMap
}
```
but in line 311 it's 
```
if wl.ShareNetNS == "" {
   // get podVulsMap
}
```
it looks like the calculated podVulsMap will never be used, or it will never be calculated, for a workload?

kyledong-suse: After scan is done, the critical, high, med Vuls of each containers will be added into workload cache scan brief. However, this GetRiskScoreMetrics API will collect pod Vuls instead of each containers.
For the calculation part, 
`if wl.ShareNetNS != "" {
   // put podVulsMap
}`
I'll sum all containers Vuls into the map but not for the pod(avoid double calculate because it will be saved into cache later).

controller/cache/compliance.go
kyledong-suse: fixed

williamlin-suse (approved):

PR #1509, opened 2024-08-16
NVSHAS-9243 support deep UI integration, fetch workload scan result b…

jeffhuang4704:
…y container id list

### DIFF
--- a/controller/rest/workload.go
+++ b/controller/rest/workload.go
@@ -262,7 +262,21 @@ func handlerWorkloadListBase(apiVer string, w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	cached := cacher.GetAllWorkloads(view, acc)
+	// get POST body
+	idlist := utils.NewSet()
+	if r.Method == http.MethodPost {
+		var idListData api.RESTAssetIDList
+		body, _ := io.ReadAll(r.Body)
+		err := json.Unmarshal(body, &idListData)
+		if err != nil {
+			// TODO: error handling
+			log.WithFields(log.Fields{"error": err}).Error("Request error")
+			restRespError(w, http.StatusBadRequest, api.RESTErrInvalidRequest)
+		}
+		idlist = utils.NewSetFromStringSlice(idListData.IDs)
+	}
+
+	cached := cacher.GetAllWorkloads(view, acc, idlist)
 
 	// Sort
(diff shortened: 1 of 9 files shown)


controller/rest/workload.go
```diff
@@ -262,7 +262,21 @@ func handlerWorkloadListBase(apiVer string, w http.ResponseWriter, r *http.Reque
...
+	// get POST body
+	idlist := utils.NewSet()
+	if r.Method == http.MethodPost {
+		var idListData api.RESTAssetIDList
+		body, _ := io.ReadAll(r.Body)
+		err := json.Unmarshal(body, &idListData)
+		if err != nil {
+			// TODO: error handling
+			log.WithFields(log.Fields{"error": err}).Error("Request error")
+			restRespError(w, http.StatusBadRequest, api.RESTErrInvalidRequest)
```
williamlin-suse, line 274: should it return after restRespError(w, http.StatusBadRequest, api.RESTErrInvalidRequest) is called?

controller/rest/workload.go
```diff
@@ -262,7 +262,21 @@ func handlerWorkloadListBase(apiVer string, w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	cached := cacher.GetAllWorkloads(view, acc)
+	// get POST body
+	idlist := utils.NewSet()
+	if r.Method == http.MethodPost {
```
williamlin-suse, line 267: Could query.brief be true when POST("/v2/workload") is requested?
Then the payload is ignored & returns in line 232?

controller/rest/workload.go
williamlin-suse, line 1: POST("/v2/workload") is actually for user with read permission.
Should line 238 in be handlerWorkloadListBase()
acc, login := getAccessControl(w, r, access.AccessOPRead) 
?

controller/rest/workload.go
jeffhuang4704: yes, you're right.
fixed.

controller/rest/workload.go
jeffhuang4704: fixed.

controller/rest/workload.go
jeffhuang4704: discussed with the use case with UI team. The query.brief will not be used in this case.

williamlin-suse (approved):

PR #1511, opened 2024-08-16
NVSHAS-9320/NVSHAS-9321: NV SSO user with fed(r/w)+regscan(r/w) permi…

williamlin-suse:
…ssions cannot create registry & see managed clusters on master cluster UI


jeffhuang4704 (approved):
looks good to me.

PR #1519, opened 2024-08-20
[NVSHAS-8461] Improve SetGlobalObjects to detect the clolud platform correctly

pohanhuangtw:
### Problem
In rancher eks, some node is unable to detect the eks.
### Solution
Extend the existing logic to get the cloud platform when parsing the platform out.
### Test performed
Test on the eks cluster, and make sure the detection correctly.

### DIFF
--- a/share/global/global.go
+++ b/share/global/global.go
@@ -80,7 +80,7 @@ func SetGlobalObjects(rtSocket string, regResource RegisterDriverFunc) (string, 
 // Get the container platform and the cloud provider in getContainerPlatform
 func getContainerPlatform(c *container.ContainerMeta) (string, string) {
 	var platform, cloudPlatform string
-
+	platform = share.PlatformDocker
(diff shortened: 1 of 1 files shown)


share/global/global.go
```diff
@@ -95,31 +77,33 @@ func SetGlobalObjects(rtSocket string, regResource RegisterDriverFunc) (string,
 	return platform, flavor, cloudPlatform, network, containers, nil
 }
 
-func getContainerPlatform(c *container.ContainerMeta) string {
+// Get the container platform and the cloud provider in getContainerPlatform
+func getContainerPlatform(c *container.ContainerMeta) (string, string) {
+	var platform, cloudPlatform string
```
williamlin-suse, line 82: platform is not default to share.PlatformDocker anymore ?

pohanhuangtw: I will fix this, thanks for the noticed

PR #1520, opened 2024-08-20
fix: NVSHAS-9336 secret controller not work

holyspectral:
In the previous version, annotation changes get ignored, which lead to these components wait indefinitely if annotation is the only part changed.

This commit fixes the issue by not ignoring the annotation change.


jeffhuang4704 (approved):
it looks good !

PR #1522, opened 2024-08-21
NVSHAS-9338:Alert 'Managed cluster [id] is disconnected from primary …

williamlin-suse:
…cluster' is incorrectly displayed even when the managed cluster is connected


jeffhuang4704 (approved):
it looks good.

PR #1524, opened 2024-08-21
Rancher SSO: User with namespace compliance read custom role is unable to login and system config too

williamlin-suse:


jeffhuang4704 (approved):
it looks good to me.

PR #1526, opened 2024-08-22
improve scan logs & stop recursive signature fetching

alopez-suse:
This makes the scan logs more understandable, by reducing unnecessary paths and adding verbosity to certain logical forks.

This also fixes the signature image info retrieval from spamming both the registry test and the scan logs, previously when finding all the images to scan for a registry/repo, the signature images would be considered regular images, and thus the controller would attempt to fetch the "signature's signature", resulting in a flood of "not found" errors.

### DIFF
--- a/share/scan/registry.go
+++ b/share/scan/registry.go
@@ -227,7 +227,7 @@ func (rc *RegClient) GetImageInfo(ctx context.Context, name, tag string, manifes
 	}
 
 	if v2SchemaError != nil {
-		log.WithFields(log.Fields{"name": name, "tag": tag, "error": err.Error()}).Error("could not build v2 image info, trying v1")
+		log.WithFields(log.Fields{"name": name, "tag": tag, "error": v2SchemaError.Error()}).Error("could not build v2 image info, trying v1")
 		// get schema v1
 		manV1, err := rc.Manifest(ctx, name, tag)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


share/scan/registry.go
```diff
@@ -219,58 +226,61 @@ func (rc *RegClient) GetImageInfo(ctx context.Context, name, tag string, manifes
...
-
-		// Even we send request with accept v1 manifest, we still get v2 format back
-		if manV1.SignedManifest.SchemaVersion <= 1 {
-			if len(manV1.SignedManifest.FSLayers) > 0 {
-				imageInfo.Layers = make([]string, len(manV1.SignedManifest.FSLayers))
-				for i, des := range manV1.SignedManifest.FSLayers {
-					imageInfo.Layers[i] = string(des.BlobSum)
-					// log.WithFields(log.Fields{"i": i, "layer": string(des.BlobSum)}).Debug("v1 manifest")
+	if v2SchemaError != nil {
+		log.WithFields(log.Fields{"name": name, "tag": tag, "error": err.Error()}).Error("could not build v2 image info, trying v1")
```
holyspectral, line 230: Given that v2SchemaError already has part of the err.Error(), maybe print `v2SchemaError` instead?
```suggestion
		log.WithFields(log.Fields{"name": name, "tag": tag, "error": v2SchemaError.Error()}).Error("could not build v2 image info, trying v1")
```

holyspectral (commented):
Generally looks good to me.

holyspectral (approved):

PR #1528, opened 2024-08-23
NVSHAS-9351: Do not generate User.Resource.Access.Denied event for Ra…

williamlin-suse:
…ncher SSO user who has global federation permission but not fedAdmin/fedReader role


jeffhuang4704 (approved):
it looks good to me.

PR #1531, opened 2024-08-26
NVSHAS-9361: Add nil pointer check in GetRiskScoreMetrics

kyledong-suse:


controller/cache/compliance.go
```diff
@@ -314,10 +314,12 @@ func (m CacheMethod) GetRiskScoreMetrics(acc, accCaller *access.AccessControl) *
 		// Assign containers' sum of vulnerabilities to pod cache
 		if wl.ShareNetNS == "" {
 			epMap[cache.workload.ID] = &wlMini{mode: mode}
-			if podVuls, ok := podVulsMap[cache.podName]; ok {
-				cache.scanBrief.CriticalVuls = podVuls.CriticalVuls
-				cache.scanBrief.HighVuls = podVuls.HighVuls
-				cache.scanBrief.MedVuls = podVuls.MedVuls
+			if cache.scanBrief != nil {
+				if podVuls, ok := podVulsMap[cache.podName]; ok {
+					cache.scanBrief.CriticalVuls = podVuls.CriticalVuls
```
jeffhuang4704, line 319: I'm just curious—since we update the wlCacheMap (global data), if scanBrief is nil, why not create it instead? Is this behavior intentional?

kyledong-suse: I think this scanBrief is assigned during scanning. if it doesn't have the right ScanObjectType, it won't be created. In our existing code, we're using this cache.scanBrief != nil for the condition in many cases. So it's better to just keep it nil for now. Maybe we can think of better options after 5.4 is released.

williamlin-suse: scanBrief is assigned during workload/host scanning.
for pod entries in wlCacheMap, when will their scanBrief be non-nil if it's not allocated here?

kyledong-suse: > scanBrief is assigned during workload/host scanning. for pod entries in wlCacheMap, when will their scanBrief be non-nil if it's not allocated here?

That's a good point! in this specific scenario, I should initialize the cache.scanBrief if the pod is in podVulsMap. @williamlin-suse please review again. Thanks.

williamlin-suse (approved):

PR #1533, opened 2024-08-27
NVSHAS-9384: Do not generate Access.Denied events in Rancher SSO sess…

williamlin-suse:
…ions (lots of such events because of custom permissions mapping)


jeffhuang4704 (approved):
it looks good.

PR #1535, opened 2024-08-28
[NVSHAS-8799] Remove prime config after update the controller, and add ReadPimeConfig to help support debug

pohanhuangtw:
### Summary

- Remove prime config after update the controller
- Add ReadPimeConfig to help support debug, for support bundle

### Test perform

- Deploy on k8s
- Simulate node die, and make sure the support match the result

### DIFF
--- a/share/scan/compliance.go
+++ b/share/scan/compliance.go
@@ -25,7 +25,7 @@ const (
 
 var (
 	dstPrefix         = "/usr/local/bin/scripts/cis_yamls/"
-	primeConfigFolder = "/tmp/compliance/"
+	primeConfigFolder = "/etc/neuvector/prime/compliance/"
 	kube160           = "cis-1.6.0"
 	kube123           = "cis-1.23"
 	kube124           = "cis-1.24"
--- a/share/scan/compliance.go
+++ b/share/scan/compliance.go
@@ -3140,19 +3140,21 @@ func PrepareBenchMeta(items map[string]api.RESTBenchCheck, metaMap map[string]ap
 	}
 }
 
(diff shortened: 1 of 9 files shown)


share/scan/compliance.go
```diff
@@ -3207,89 +3222,43 @@ func LoadConfig(primeConfig string, params *UpdateConfigParams, updateCompliance
...
-	<-done
 }
 
 func UpdateComplianceConfigs() {
-	// update prime cis configs
-	primeCISConfig := fmt.Sprintf("%s%s.yaml", primeConfigPrefix, cisVersion)
-
-	// update prime cis docker configs
-	primeDockerConfig := fmt.Sprintf("%s%s.yaml", primeConfigPrefix, "cis-docker")
+	ReadPrimeConfig = false
```
holyspectral, line 3229: `UpdateComplianceConfigs()` is called in a [go routine](https://github.com/neuvector/neuvector/blob/main/controller/controller.go#L745)

And `updateComplianceWithPrimeConfig()` and `updateImageBenchWithPrimeConfig()` are called in different go routines, e.g., [this](https://github.com/neuvector/neuvector/blob/main/share/scan/compliance.go#L3233) and [this](https://github.com/neuvector/neuvector/blob/main/share/scan/compliance.go#L3291)

You probably have to protect the flag using sync.Mutex.

holyspectral (approved):
LGTM.  `LoadConfig()` seems to be called multiple times.  Is it expected to treat the first success as the criteria to set ReadPrimeConfig=true?

holyspectral (approved):

PR #1536, opened 2024-08-28
fix: NVSHAS-9387 get namespace from k8s token

holyspectral:
In one of the previous changes, POD_NAMESPACE environment is removed from cert-upgrader.  This leads to not being able to get the correct namespace when running in a non-default namespace, e.g., `cattle-system`.

This PR fixes the issue by getting namespace from service account folder.


jeffhuang4704 (approved):
it looks good to me.

PR #1539, opened 2024-08-29
NVSHAS-9194: add alerting msg in both /v1/system/rbac and /v1/eula if…

williamlin-suse:
… the neuvector-binding-secret role(binding) is incorrect


holyspectral (approved):
LGTM.  Just about to look into this.  Thanks!

jeffhuang4704 (approved):
it looks good to me.

PR #1540, opened 2024-08-29
NVSHAS-9392: Rancher SSO user what assigned Project Role with registryscan(r/w) pe…

williamlin-suse:
…rmissions on namespace nvlab cannot Test Scan quay.io/nvlab/iperf


jeffhuang4704 (approved):
it looks good to me.

PR #1543, opened 2024-08-30
NVSHAS-7687: Fix setting protect mode and KvCongestCtl

kyledong-suse:


jayhuang-suse (approved):
Need to simulate more on the inputs. **Debug:[policy session log ctrl parser cluster ddos cpath timer error conn monitor].** on what if they are ignored.

PR #1544, opened 2024-08-31
NVSHAS-7601: Improve restore from PV config backup during scenario wh…

williamlin-suse:
…ere replacement controllers failed to sync


jeffhuang4704 (approved):
it looks good to me.

PR #1559, opened 2024-09-12
Use base image tag: pre54

jayhuang-suse:
Since 5.4, our latest base image has been upgraded to BCI. The change enable the backport with alpine containers.


jayhuang-suse:
Changing the tag ....

PR #1561, opened 2024-09-13
fix: NVSHAS-9430 redundant type from array warnings

holyspectral:
Fix redundant type from array, slice, or map composite literal pointer warnings from gopls.


jeffhuang4704 (approved):
looks good to me.

jayhuang-suse (changes requested):
No files under the agent now.

jayhuang-suse (approved):

PR #1571, opened 2024-09-20
feat: golangci-lint as git hook

holyspectral:
Adding golangci-lint for developers to scan local changes before committing.

To use this feature, you will need to install [pre-commit](https://pre-commit.com/). 
After that, use `pre-commit install` to install pre-commit hooks and use `pre-commit uninstall` to remove the hooks.

### DIFF
--- a/build/pre-commit/golangci-lint-staged
+++ b/build/pre-commit/golangci-lint-staged
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+GOLANGCI_LINT=`which golangci-lint`
+
+# Check for golangci-lint
+if [[ ! -x "$GOLANGCI_LINT" ]]; then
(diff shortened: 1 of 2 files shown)


build/pre-commit/golangci-lint-staged
pohanhuangtw, line 1: Hi @holyspectral  
Overall the code change looks good to me, however I am not sure how you add the pre commit and golangci-lint in the env

holyspectral: > To use this feature, install [pre-commit](https://pre-commit.com/).
> After that, use pre-commit install to install pre-commit hooks and use pre-commit uninstall to remove the hooks.

See my comments above.  You would also need to install `golangci-lint` in your local to utilize this pre-commit hook.

PR #1584, opened 2024-09-27
CVE-2024-41110: Upgrade docker package

jayhuang-suse:
Upgrade docker's vendor package to v26.1.5


vendor/github.com/neuvector/k8s/watch.go
williamlin-suse, line 1: This file should not be changed in this PR

vendor/github.com/neuvector/k8s/client.go
williamlin-suse, line 1: This file should not be changed in this PR

PR #1599, opened 2024-10-11
NVSHAS-8824: User fails to delete own groups, cannot create namespace…

williamlin-suse:
…-scoped groups


jeffhuang4704 (approved):
looks good to me.

PR #1614, opened 2024-10-16
NVSHAS-9468: Fix CVE for using jwt:v5

kyledong-suse:


controller/rest/auth.go
```diff
@@ -662,7 +666,7 @@ func restReq2User(r *http.Request) (*loginSession, int, string) {
 
 	// Check token end-of-life
 	now := time.Now()
-	if now.Unix() >= claims.ExpiresAt {
+	if claims.ExpiresAt != nil && now.After(claims.ExpiresAt.Time) {
```
holyspectral, line 669: We should reject the claims without `ExpiresAt`, otherwise this JWT token will never expire.

controller/rest/auth.go
```diff
@@ -2094,7 +2100,7 @@ func fedMasterTokenAuth(userName, masterToken, secret string) (*share.CLUSUser,
 		return nil, nil, err
 	}
 
-	if time.Now().Unix() >= claims.ExpiresAt {
+	if claims.ExpiresAt != nil && time.Now().After(claims.ExpiresAt.Time) {
```
holyspectral, line 2103: We should reject the claims without ExpiresAt, otherwise the JWT token would never expire.

holyspectral (changes requested):
Generally it looks okay to me, but please fix the expiry date check to make sure that claims without expiry date is rejected correctly.

controller/rest/auth.go
kyledong-suse: fixed

yasker (changes requested):
1. Please squash the commits into two or more. One for functional change, another one for vendor directory change, to make it easy to understand the change. 
2. In the future commits, make sure there is no 'merge' commit. Rebase is preferred. I will talk more about it next week.

kyledong-suse:
> 1. Please squash the commits into two or more. One for functional change, another one for vendor directory change, to make it easy to understand the change.
> 2. In the future commits, make sure there is no 'merge' commit. Rebase is preferred. I will talk more about it next week.

Hi @yasker, totally agree with these 2 comments. I can squash those merges into

kyledong-suse:
> 1. Please squash the commits into two or more. One for functional change, another one for vendor directory change, to make it easy to understand the change.
> 2. In the future commits, make sure there is no 'merge' commit. Rebase is preferred. I will talk more about it next week.

Hi @yasker , since the original commit(includes functional change and vendor directory) is in one commit already. It's not easy to be separated. So I'm going to create a new PR which includes separate commits to include different components.

PR #1616, opened 2024-10-17

williamlin-suse:
neuvector/share - part 1


holyspectral (commented):
Thanks for doing this!  It generally looks good to me and it will help a lot for us to troubleshoot those issues previously hidden.

A few comments:
1. Are all unhandled errors in `share/cluster` are already handled in consul level correctly?  If so, we may consider making them NOT return the error.
2. Errors from syscalls should be handled, otherwise it would be hard to troubleshoot those environment issues. 
3. Some codes can be slightly simplified.

williamlin-suse:
Will address the concerns in the following PRs. Thanks

holyspectral (approved):

PR #1618, opened 2024-10-18
fix: NVSHAS-9525 lint issues

holyspectral:
This PR fixes a few lint issues.

### DIFF
--- a/controller/scan/scan.go
+++ b/controller/scan/scan.go
@@ -235,14 +235,18 @@ func AddScanner(id string) {
 		regScher.Init()
 	}
 
-	regScher.AddProcessor(id)
+	if err := regScher.AddProcessor(id); err != nil {
+		log.WithError(err).Error("failed to add processor")
+	}
 }
 
 func RemoveScanner(id string) {
 	log.WithFields(log.Fields{"id": id}).Info()
 
 	if regScher != nil {
-		regScher.DelProcessor(id)
+		if _, err := regScher.DelProcessor(id); err != nil {
+			log.WithError(err).Error("failed to delete processor")
+		}
 	}
 }
 
--- a/controller/scan/scan.go
+++ b/controller/scan/scan.go
@@ -1,6 +1,7 @@
 package scan
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"regexp"
@@ -227,7 +228,7 @@ func FedRoleChangeNotify(fedRole string) {
 	smd.fedRole = fedRole
 }
 
-func AddScanner(id string) {
+func AddScanner(id string) error {
 	log.WithFields(log.Fields{"id": id}).Info()
 
 	if regScher == nil {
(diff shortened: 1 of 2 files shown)


controller/scan/scan.go
```diff
@@ -235,14 +235,18 @@ func AddScanner(id string) {
 		regScher.Init()
 	}
 
-	regScher.AddProcessor(id)
+	if err := regScher.AddProcessor(id); err != nil {
```
yasker, line 238: What would be the side effects if those actions failed? Should we populate the error to the caller?

controller/scan/scan.go
```diff
@@ -235,14 +235,18 @@ func AddScanner(id string) {
...
+		log.WithError(err).Error("failed to add processor")
+	}
 }
 
 func RemoveScanner(id string) {
 	log.WithFields(log.Fields{"id": id}).Info()
 
 	if regScher != nil {
-		regScher.DelProcessor(id)
+		if _, err := regScher.DelProcessor(id); err != nil {
```
yasker, line 247: Same here.

controller/scan/scan.go
holyspectral: Hi @yasker, good question! When this happens, it means someone tries to add a processor when it's already added. It's a programming error and should be fixed by the developer. The best action here is to log the error so the developer knows.

holyspectral: Similar to EINVAL errno.

yasker: If it's a duplicate provider, it should be fine. However, there is no error type check here. It's not immediately obvious it will only due to a duplicate provider by reading the code. The normal process is to populate the error up in the chain to the caller. Should we populate the error here as well?

jeffhuang4704 (approved):
looks good to me.

controller/scan/scan.go
holyspectral: yes I think that would be perfect if we can have all the code follow this convention, which is golang's standard.  I can work on it.  
Currently we don't have that kind of coding convention, which makes error handling and logging very complicated.  I think we can discuss about this in the meeting next week too.

yasker: Let's raise it up next week. We need to consider the error path to improve the robustness of the code.

yasker (changes requested):
Need to consider the error path and response the right error to the caller.

holyspectral:
@yasker I've reworked the error handling to the highest caller that we can handle, i.e., consul event handler.  Please take a look.

yasker (approved):

yasker:
LGTM now. I have a bigger problem with [the main function](https://github.com/holyspectral/neuvector/blob/476c77f70757dbf7d083bec21d34ad1ca130559e/controller/controller.go#L212) 's error handling but that's the topic for another day.

holyspectral:
yes...more than 800 lines in main().  We should definitely work on it.

PR #1619, opened 2024-10-18
NVSHAS-9468: Fix CVE-2020-26160 to replace jwt-go with jwt:v5

kyledong-suse:
1. Fix CVE-2020-26160 in controller to replace deprecated "github.com/dgrijalva/jwt-go" with "github.com/golang-jwt/jwt/v5"
2. Fix CVE-2020-26160 in vendor to replace deprecated "github.com/dgrijalva/jwt-go" with "github.com/golang-jwt/jwt/v5"


holyspectral (approved):
LGTM

yasker (changes requested):
Check with distro team. It's better if we don't call out CVE fix in the title. It might got some unwanted attention. 

Just say replacing A with B is fine. 

Otherwise LGTM.

yasker (commented):
Well, getting different opinions on this. @flavio any suggestions?

kyledong-suse:
> Well, getting different opinions on this. @flavio any suggestions?

Apologies, I just saw this comment without refreshing the page. I'll stay tuned until all parties reach an agreement.

flavio:
We have to be careful about embargoed vulnerabilities. This is a CVE that comes from one of our dependencies, it was also publicly advertised by GitHub under the security tab of the project (which is visible to everybody).

Because of that, it's fine to mention inside of the commit messages AND the PR that we're fixing a CVE. I would also recommend to add the CVE number to make easier to understand what was fixed and why we did these changes

kyledong-suse:
According to the comment from @flavio , I modified the title to mention that we're fixing CVE-2020-26160. @yasker , please let me know if it works for you. Thanks.

yasker:
@kyledong-suse Thanks. I agree with Flavio here after more discussion. Can we put the CVE number in the commit title or message? Since the GitHub PR name won't make it into the Git history.

Otherwise LGTM.

kyledong-suse:
@yasker done with the fix. Thanks for reviewing it.

yasker (approved):

PR #1620, opened 2024-10-19
NVSHAS-9542: remove ringbuffer package from enforcer

jayhuang-suse:
Remove the confusing package.


agent/probe/process.go
```diff
@@ -2670,37 +2647,31 @@ func (p *Probe) addProcHistory(id string, proc *procInternal, bFromMonitor bool)
...
-	// if histProc.Length()() == 400 {
-	//		var list []*procInternal
-	//		elements := histProc.DumpExt()
-	//		for i := 0; i < len(elements); i++ {
-	//			list = append(list, elements[i].(*procInternal))
-	//		}
-	//		printLastProcElements(list, 10)
-	//	}
+	// cap the history length to 400 entries
+	if len(histProc) < 400 {
```
holyspectral, line 2656: Better to make the `400` as a const.

agent/probe/process.go
```diff
@@ -2670,37 +2647,31 @@ func (p *Probe) addProcHistory(id string, proc *procInternal, bFromMonitor bool)
 		defer p.unlockProcMux()
 	}
 
-	if histProc, ok = p.procHistoryMap[id]; !ok {
-		histProc = ringbuffer.New(400) // TODO: more ?
-		p.procHistoryMap[id] = histProc
+	histProc, ok := p.procHistoryMap[id]
+	if !ok {
+		histProc = make([]*procInternal, 0)
```
holyspectral, line 2652: We should give `histProc` a default capacity.

agent/probe/process.go
```diff
@@ -2670,37 +2647,31 @@ func (p *Probe) addProcHistory(id string, proc *procInternal, bFromMonitor bool)
...
-	//			list = append(list, elements[i].(*procInternal))
-	//		}
-	//		printLastProcElements(list, 10)
-	//	}
+	// cap the history length to 400 entries
+	if len(histProc) < 400 {
+		histProc = append(histProc, proc)
+	} else {
+		histProc = append(histProc[1:], proc)
+	}
```
holyspectral, line 2660: Is this function performance critical?  In the original implementation, all write operation is done via assignment, which is pretty quick.  With `append()`, when it reaches the capacity of the slice, it could be very slow. In my experiment, it would be 2~3 times slower on my laptop.

yasker:
LGTM from a coding style perspective.

agent/probe/process.go
jayhuang-suse: If you understand how the struct works, you should know why the assignment is not a proper fit here.

agent/probe/process.go
jayhuang-suse: It is optional.

yasker: Agree with @holyspectral here. This is a magic number and should be a constant at least.

agent/probe/process.go
holyspectral: If I’m not mistaken, the original implementation used a ring buffer which manipulates slice operation using head/tail indices.  From I can see, this will not make slice resize, therefore it will have better performance - exactly the purpose of a ring buffer. 

Now, can you elaborate more about why you think it’s not a proper fit?

jayhuang-suse: No proper open source project is suitable, it needs encoding or allocate extra locks. For performance-wise, all of them are not a good choice. I have a previous implementation which meets all performance requirement for the special 
case but not everyone is happy with that. This PR is just only an acceptable solution.

agent/probe/process.go
jayhuang-suse: Done

agent/probe/process.go
holyspectral: ok it sounds fair.  Let's merge it and keep an eye on the performance. If the slice gets filled up quickly, it might help to set a default capacity too.  In the previous implementation the default capacity was 400. 
https://github.com/neuvector/neuvector/pull/1620#discussion_r1807156051

holyspectral (approved):
Generally looks good to me.  We should keep an eye on its performance impact.

PR #1621, opened 2024-10-19
ReadCmdLine failure: Fix golint PR error.

jayhuang-suse:
The ReadCmdLine() failed because it did not fetch the line.

### DIFF
--- a/share/system/system_linux.go
+++ b/share/system/system_linux.go
@@ -327,14 +327,13 @@ func (s *SystemTools) ReadCmdLine(pid int) ([]string, error) {
 	defer file.Close()
 
 	scanner := bufio.NewScanner(file)
-	//	for scanner.Scan() {
-	// first line only
-	cmds = strings.Split(string(scanner.Text()), "\x00")
-	for i, t := range cmds { // it guarantees the tokens aren't pinning memory.
-		cmds[i] = string([]byte(t))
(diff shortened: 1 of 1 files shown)


share/system/system_linux.go
```diff
@@ -327,14 +327,13 @@ func (s *SystemTools) ReadCmdLine(pid int) ([]string, error) {
...
 	scanner := bufio.NewScanner(file)
-	//	for scanner.Scan() {
-	// first line only
-	cmds = strings.Split(string(scanner.Text()), "\x00")
-	for i, t := range cmds { // it guarantees the tokens aren't pinning memory.
-		cmds[i] = string([]byte(t))
-	}
-	//		break
-	//	}
+	if scanner.Scan() {
```
holyspectral, line 330: If `scanner.Scan()` returns false, should we treat this as an error for its caller to handle? 

According to its man page, the content is possible to be empty or ill-formed, as it’s controlled by the process. 

https://www.man7.org/linux/man-pages/man5/proc_pid_cmdline.5.html

holyspectral (approved):
Given the original issue is breaking automation, approving it right now.

share/system/system_linux.go
jayhuang-suse: It is normal to read nothing from the command line.

PR #1626, opened 2024-10-22

williamlin-suse:
neuvector/share - part 2

### DIFF
--- a/share/cluster/grpc.go
+++ b/share/cluster/grpc.go
@@ -94,6 +94,8 @@ func NewGRPCServerTCP(endpoint string) (*GRPCServer, error) {
 		return nil, fmt.Errorf("failed to create new server credentials: %w", err)
 	}
 
+	//TODO: addressing go-linter "SA1019: grpc.NewGZIPDecompressor is deprecated: use package encoding/gzip. (staticcheck)"
+	//nolint:staticcheck
 	opts := []grpc.ServerOption{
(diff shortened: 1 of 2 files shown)


share/cluster/grpc.go
```diff
@@ -94,6 +94,8 @@ func NewGRPCServerTCP(endpoint string) (*GRPCServer, error) {
 		return nil, fmt.Errorf("failed to create new server credentials: %w", err)
 	}
 
+	//to avoid go-linter "SA1019: grpc.NewGZIPDecompressor is deprecated: use package encoding/gzip. (staticcheck)"
```
yasker, line 97: Can we add `TODO: addressing go-linter....` here? Same below, so we can address it later and not forget about it.

yasker (changes requested):

share/cluster/grpc.go
yasker: I meant...
`TODO: addressing go-linter "SA1019: grpc.NewGZIPDecompressor is deprecated: use package encoding/gzip. (staticcheck)"`

yasker (changes requested):
Can you also use `git rebase` instead of merge for the new commits? There shouldn't be any `merge` commits in the PR.

jeffhuang4704 (approved):
looks good to me.

yasker (approved):

PR #1629, opened 2024-10-23
NVSHAS-9532: The image scan is completed but deployment is still not allowed

williamlin-suse:
This PR will allow configuring non-https URL for default registry used in admission control rules evaluation.
When the container image specified in a pod spec doesn't contain the registry, for example:
```
spec:
      containers:
        - name: iperfserver
          image: iperf
```
NV controller uses the following 4 registries to find the scan result of the image:
```
https://docker.io/
https://index.docker.io/
https://registry.hub.docker.com/
https://registry-1.docker.io/
```
If user wants to use different registries as the default registries for admission control rule evaluation, they can specify an env var in NV controller deployment like:
```
env:
  - name: CTRL_SEARCH_REGISTRIES
    value: "http://myregistry1.net,https://myregistry2.net"
```
However, the logic in NV 5.4 doesn't support http scheme in the value.
For above example, NV controller uses https://myregistry1.net & https://myregistry2.net, not http://myregistry1.net & https://myregistry2.net, as the default registries for admission control rule evaluation.
In NVSHAS-9532, we see the user's registry server is http://harbor.freeit-gyh.com/ (http scheme) whose image scan result can not be referenced for admission control rule evaluation.

This PR is to support http scheme specified in the CTRL_SEARCH_REGISTRIES env var of NV controller pod

### DIFF
--- a/controller/rest/rest.go
+++ b/controller/rest/rest.go
@@ -1493,6 +1493,36 @@ func PreInitContext(ctx *Context) {
 	cfgHelper = kv.GetConfigHelper()
 }
 
+func initSearchRegistries(ctx *Context) {
+	searchRegistries = utils.NewSet()
+
+	for _, reg := range strings.Split(ctx.SearchRegistries, ",") {
+		reg = strings.Trim(reg, " ")
+		if len(reg) > 0 {
+			parsedReg, err := url.Parse(reg)
+			if err != nil {
+				log.WithError(err).WithFields(log.Fields{"registry": reg}).Warn("unable to parse registry")
+				continue
+			}
+			var regURL string
+			if parsedReg.Scheme == "" {
+				// when scheme is not specified in reg, host is in parsedReg.Path
+				path := parsedReg.Path
+				if i := strings.Index(parsedReg.Path, "/"); i > 0 {
+					path = parsedReg.Path[:i]
+				}
+				regURL = fmt.Sprintf("https://%s/", path)
+			} else {
+				// when scheme is specified in reg, host is in parsedReg.Host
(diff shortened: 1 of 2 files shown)


yasker:
Hi William, can you put the explanation into the commits in addition to the PR description? So the reader of `git log` will have context on what is being changed and why.

williamlin-suse:
> Hi William, can you put the explanation into the commits in addition to the PR description? So the reader of `git log` will have context on what is being changed and why.

Do you mean to put the whole description(above) in 
`git commit -m "<very long description>.."`
?

yasker:
> > Hi William, can you put the explanation into the commits in addition to the PR description? So the reader of `git log` will have context on what is being changed and why.
> 
> 
> 
> Do you mean to put the whole description(above) in 
> 
> `git commit -m "<very long description>.."`
> 
> ?

Normally I will use `got commit -a --amend` to add the comments in the editor. Make sure the comments are wrapped at 80-ish characters. Of course you can choose to use `-a` or not.

williamlin-suse:
> > > Hi William, can you put the explanation into the commits in addition to the PR description? So the reader of `git log` will have context on what is being changed and why.
> > 
> > 
> > Do you mean to put the whole description(above) in
> > `git commit -m "<very long description>.."`
> > ?
> 
> Normally I will use `got commit -a --amend` to add the comments in the editor. Make sure the comments are wrapped at 80-ish characters. Of course you can choose to use `-a` or not.

done

jeffhuang4704 (approved):
looks good to me.

yasker:
@williamlin-suse Sorry, but I don't see the new commit message.

PR #1630, opened 2024-10-23
NVSHAS-9525 resolve golinter issue

jeffhuang4704:


yasker (changes requested):
Let's discuss this tomorrow. Error handling is more than logging the errors. The unhappy path needs to be thought through.

jeffhuang4704:
Based on our discussion, we need to address the error with consideration for context. I am closing this PR as it may have significant differences.

PR #1637, opened 2024-10-25
NVSHAS-9425: create nfq when container has vxlan 

gfsuse:
Container has vxlan setup, in protect mode if we split veth interface into veth bridge the vxlan setup will be affected, so we use nfq(iptable rules) instead in this scenario.

### DIFF
--- a/agent/pipe/port.go
+++ b/agent/pipe/port.go
@@ -62,6 +62,7 @@ type InterceptPair struct {
 	BCMAC  net.HardwareAddr
 	UCMAC  net.HardwareAddr
 	Addrs  []share.CLUSIPAddr
+	Vxlan  bool
 }
 
 func (m *InterceptPair) InPort() string {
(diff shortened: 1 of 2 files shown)


agent/pipe/port.go
```diff
@@ -935,6 +937,10 @@ func readAllContainerPorts(pid int, existPairs map[string]*InterceptPair) ([]*In
 			}
 		}
 
+		if link.Type() == "vxlan" {
+			hasVxlan = true
+			log.Debug("vxlan")
```
yasker, line 942: Need more clear debug message than just `vxlan`. Maybe `VXLAN detected in container` or something more verbose.

yasker (changes requested):
1. Use `:` instead of `,` after the ticket ID in the title.
2. There is a typo in the commit title.
3. Better to keep the title under 50 characters. 
4. Better to describe why it matters in the commit message.

yasker (approved):

kyledong-suse (approved):
LGTM

jayhuang-suse (approved):
Look good for me.

PR #1638, opened 2024-10-25
NVSHAS-9571 fix image asset advanced filter issue

jeffhuang4704:
The full name displayed in the UI is made up of two separate fields (specifically, name and tag). 
The backend needs to take this into account when performing quick filter search.


yasker (changes requested):
Please add the detail explanation to the git commit message as well (since `git log` won't show the PR). Otherwise LGTM coding style-wise.

yasker (approved):

PR #1639, opened 2024-10-26
NVSHAS-9589: Managed clusters disconnected - Version mismatch with primary cluster

williamlin-suse:
Reason: managed clusters' REST version is not stored in primary cluster's cache


jeffhuang4704 (approved):
looks good to me

yasker (approved):

PR #1640, opened 2024-10-29
NVSHAS-8824: User fails to delete own groups, cannot create namespace-scoped groups

williamlin-suse:
Handle access denied error

### DIFF
--- a/controller/rest/group.go
+++ b/controller/rest/group.go
@@ -615,10 +615,17 @@ func handlerGroupCreate(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
 	}
 
 	// Do not lock, reply on cluster.PutIfNotExist() for consistency
-	if exist, _ := cacher.DoesGroupExist(rg.Name, acc); exist {
+	if exist, err := cacher.DoesGroupExist(rg.Name, acc); exist {
 		e := "Group already exists"
 		log.WithFields(log.Fields{"name": rg.Name}).Error(e)
 		restRespErrorMessage(w, http.StatusBadRequest, api.RESTErrDuplicateName, e)
+		return
+	} else if err != common.ErrObjectNotFound {
+		if err == common.ErrObjectAccessDenied {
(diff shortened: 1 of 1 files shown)


controller/rest/group.go
```diff
@@ -615,11 +615,14 @@ func handlerGroupCreate(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
...
-	if exist, _ := cacher.DoesGroupExist(rg.Name, acc); exist {
+	if exist, err := cacher.DoesGroupExist(rg.Name, acc); exist {
 		e := "Group already exists"
 		log.WithFields(log.Fields{"name": rg.Name}).Error(e)
 		restRespErrorMessage(w, http.StatusBadRequest, api.RESTErrDuplicateName, e)
 		return
+	} else if err == common.ErrObjectAccessDenied {
+		restRespAccessDenied(w, login)
+		return
 	}
```
yasker, line 626: What about other errors?

yasker (changes requested):

controller/rest/group.go
williamlin-suse: There are 3 cases:
1. The to-be-created group exists & the caller has permission for that group -> handled by: if { ... }
2. The to-be-created group exists but the caller doesn't have permission for that group -> handled by: else if { ... }
3. The to-be-created does not exist -> the execution keeps going for creating the group

yasker: From reading the code, it is possible that `!exists && err != NotFound` exists. It is better not to rely on the assumption that the function won't return other errors. If it's unexpected, we need to at least log the error and return it.

williamlin-suse: code change as advised

jeffhuang4704 (approved):
looks good to me.

yasker (approved):

yasker (changes requested):
LGTM. Please squash the commits locally.

yasker (approved):

PR #1643, opened 2024-10-29
NVSHAS-9605: Export group with invalid policy mode & process profile mode values is mistakenly allowed

williamlin-suse:
Export group with invalid policy/profile mode value is wrongly allowed
Use a shared function to validate the policy/profile mode values
Also remove license check for configuring Protect mode

### DIFF
--- a/controller/rest/crdsecurityrule.go
+++ b/controller/rest/crdsecurityrule.go
@@ -2222,9 +2222,7 @@ targetpass:
 	var defPolicyMode, defProfileMode string
 	if utils.DoesGroupHavePolicyMode(crdCfgRet.TargetName) {
 		if policyModeCfg != nil {
-			if *policyModeCfg != share.PolicyModeLearn &&
-				*policyModeCfg != share.PolicyModeEvaluate &&
-				*policyModeCfg != share.PolicyModeEnforce {
+			if !share.IsValidPolicyMode(*policyModeCfg) {
 				errMsg = fmt.Sprintf("%s Rule format error:   Target group %s invalide policy mode %s",
 					reviewTypeDisplay, crdCfgRet.TargetName, *policyModeCfg)
 				errCount++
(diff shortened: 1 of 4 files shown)


controller/rest/crdsecurityrule.go
```diff
@@ -3358,6 +3358,15 @@ func handlerGroupCfgExport(w http.ResponseWriter, r *http.Request, ps httprouter
 
 		// export process and file profiles
 		exportProcessRule(gname, &rconf, &(resptmp.Spec), acc)
+		if utils.DoesGroupHavePolicyMode(gname) {
+			allowed := utils.NewSetFromStringSlice([]string{share.PolicyModeLearn, share.PolicyModeEvaluate, share.PolicyModeEnforce})
```
yasker, line 3362: It's better to define valid policies as a constant. It doesn't look like we will change it from time to time.

yasker (changes requested):
Also better to keep the title under 50 chars. We can put more context in the commit message.

controller/rest/crdsecurityrule.go
williamlin-suse: share.PolicyModeLearn, share.PolicyModeEvaluate, share.PolicyModeEnforce are string constant

controller/rest/crdsecurityrule.go
```diff
@@ -3358,6 +3358,15 @@ func handlerGroupCfgExport(w http.ResponseWriter, r *http.Request, ps httprouter
...
+			for _, mode := range []*string{resptmp.Spec.Target.PolicyMode, resptmp.Spec.ProcessProfile.Mode} {
+				if mode != nil && !allowed.Contains(*mode) {
+					restRespError(w, http.StatusBadRequest, api.RESTErrInvalidRequest)
```
holyspectral, line 3365: If I understand the logic correctly, when hitting the logic here, the caller is asking an export data for an group that contains invalid mode, which can't be exported, is that correct?  If so, I would suggest adding some error message in the payload at least, for example:
```suggestion
					restRespErrorMessage(w, http.StatusBadRequest, api.RESTErrInvalidRequest, "the group XXX can't be exported because xxx")
```

holyspectral (approved):
LGTM.  I specially like those duplicate codes are simplified with a utility function. :+1:

yasker (approved):

PR #1644, opened 2024-10-30
NVSHAS-9608:grpc client failed to rx > 4MB pkt

jayhuang-suse:
The default "rx" grpc packet size of a client at controller is 4MB but the remote scanner grpc server TX is not set and its default 4GB. When the server send response over 4MB (6MB in this case), the client failed it. 

Current solution is to set the packet sizes of both directions as 20MB. If we hit the limit again, it needs to evaluate the performance on the bigger packet sizes or make chunks between two endpoints.


kyledong-suse (approved):
LGTM

yasker (changes requested):
That's a good explanation of the problem in the PR's comment. Can we add it to the commit's content too?

Also better to keep the title less than 50 chars.

jayhuang-suse:
Updated.

yasker (approved):

PR #1646, opened 2024-10-31
NVSHAS-9609:Fix clang-tidy(lint) warning on c code in repository.

gfsuse:
to run clang-tidy check on c code, user need have llvm, clang and clang-tidy installed, and download 'run-clang-tidy.py' at [here](https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clang-tidy/tool/run-clang-tidy.py)
on any c code directory please run 'run-clang-tidy.py $PWD'
script will search compile_commands.json in current directory and .clang-tidy file in source code's parent directory


dp/dpi/dpi_log.c
jayhuang-suse, line 1: Would you double check the data pointer of the "cache"?

jayhuang-suse, line 1: Would you double check the data pointer of the "cache"?

monitor/monitor.c
```diff
@@ -908,7 +907,9 @@ int main (int argc, char **argv)
         break;
     case MODE_AGENT:
         ret = system(SCRIPT_SYSCTL);
-
+        if (ret != 0) {
+            debug("Sysctl -p failed rc=%d.\n", ret);
+        }
```
yasker, line 912: Can we continue with the syscall failure? Especially the variables assignment below.

yasker (commented):
A side comment: we need to enable C lint for the repo as well. It should be a part of local build process too. cc @BinX-Suse @holyspectral

dp/dpi/dpi_log.c
holyspectral: Agreed.  It probably needs `rcu_map_add()` or `rcu_map_add_replace()`.

holyspectral (commented):
Overall it looks good to me.

Maybe it's a little out of scope, but I think we should also consider normalizing out of memory error handling.  For example, `-ENOMEM` is a good return value when it fails to allocate memory block.

dp/dpi/dpi_log.c
gfsuse: updated with cache NULL check

jayhuang-suse (approved):

monitor/monitor.c
gfsuse: I think 'sysctl -p' is to Load  in  sysctl  settings  from  the  file specified or /etc/sysctl.conf if none given, sometime if customer does not give enough permission it can fail, but it is not a fatal error to stop whole agent(enforcer) process from being run.

holyspectral (approved):
There is still things to figure out about clang-tidy version and llvm/clang version in order to setup a GHA, but I'm ok with merging this change first.

kyledong-suse (approved):
From code perspective, it LGTM.

monitor/monitor.c
yasker: Talked with George. This is a non-fatal error but will have some implications on performance or capability. We will add more information here (and the same with the change below).

yasker:
> @yasker yes we have been discussing this. Thanks for the reminder!
> 
> In brief, there are two options right now depending on how many findings are there and whether we want to include them in Makefile.
> 
> 1. Include linter steps in dp's Makefile => it will make sure no linter errors, but we need to update builder image and SLSA build.
> 2. Create a separate GHA for this => less changes in builder/SLSA build, but developers have to make sure they have linter installed at their local.

I am leaning towards option 1 since adding steps in Makefile can make sure it's consistent across local and PR builds. But if it's too much effort, we can go by option 2 and implement option 1 in the future.

monitor/monitor.c
gfsuse: I have added more debug message.

yasker: Maybe something along the line of
```
WARNING: sysctl failed to load /etc/sysctl.conf (including net.core.somaxconn and net.unix.max_dgram_qlen).
It might have a performance implication on the system.
rc = %d.\n
```

Same below.

yasker (changes requested):

monitor/monitor.c
gfsuse: I have updated debug messages in both places, thanks

gfsuse:
I have updated with latest warning messages.

yasker (approved):

PR #1652, opened 2024-11-08
fix: NVSHAS-9624 rewrite swagger validation

holyspectral:
1. Use redocly/cli to replace swagger-validator-action.
2. Upgrade related actions and golang version.
3. Fixed a few issues in api docs.
4. Validate internal_apis.yaml

### DIFF
--- a/.github/workflows/unitest.yaml
+++ b/.github/workflows/unitest.yaml
@@ -8,13 +8,17 @@ jobs:
   unitest:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-go@v2
+    - uses: actions/checkout@v4
+    - uses: actions/setup-go@v5
       with:
-        go-version: '1.22.1'
+        go-version: '1.22.9'
     - run: |
         ./unitest.sh
-    - name: openapi-lint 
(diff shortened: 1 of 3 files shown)


.github/workflows/unitest.yaml
```diff
@@ -8,13 +8,17 @@ jobs:
   unitest:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-go@v2
+    - uses: actions/checkout@v4
+    - uses: actions/setup-go@v5
       with:
-        go-version: '1.22.1'
+        go-version: '1.22.9'
```
williamlin-suse, line 14: isn't neuvector repo using 1.22.7 ?

holyspectral: Yes.  I think we can go faster on unit test given that it’s usually compatible and it can help to discover issues earlier if it’s not the case when we upgrade to 1.22.9 later.

williamlin-suse:
LGTM

pohanhuangtw (approved):
LGTM

holyspectral:
@williamlin-suse if you're okay with it, could you help to approve this PR?  Thanks!

williamlin-suse (approved):

PR #1659, opened 2024-11-13
[NVSHAS-9131] Implement Workqueue Solution for Efficient Massive Image Scan Requests

pohanhuangtw:
### Summary

- Introduced a workqueue solution to handle large-scale image scan requests efficiently, with a configurable default capacity of 2048.
- Implemented a semaphore within the ScannerManager to limit scanner usage.
- Added tests to validate functionality under high-capacity workloads.
- Added a shutDown function to clean up goroutines created in the manager, preventing goroutine leaks when the controller shuts down.
- Add scanner load balancer to balance the scanner usage.

### Test 

- Run NV jenkins pipeline with controller mode to scan more than 300/600/1000 images scan at once

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -22,6 +22,7 @@ import (
 	"github.com/neuvector/neuvector/controller/opa"
 	"github.com/neuvector/neuvector/controller/resource"
 	"github.com/neuvector/neuvector/controller/rest"
+	"github.com/neuvector/neuvector/controller/rpc"
 	"github.com/neuvector/neuvector/controller/ruleid"
 	"github.com/neuvector/neuvector/controller/scan"
 	"github.com/neuvector/neuvector/db"
@@ -272,6 +273,12 @@ func main() {
 	en_icmp_pol := flag.Bool("en_icmp_policy", false, "Enable icmp policy learning")
 	autoProfile := flag.Int("apc", 1, "Enable auto profile collection")
 	custom_check_control := flag.String("cbench", share.CustomCheckControl_Disable, "Custom check control")
+	maxScannerTasks := flag.Int("max_scanner_tasks", 2, "Maximum number of concurrent tasks a scanner can handle")
+	maxConcurrentRepoScanWorkers := flag.Int("max_concurrent_repo_scan_workers", 16, "Maximum number of concurrent repository scan workers")
+	scanJobQueueCapacity := flag.Int("scan_job_queue_capacity", 2048, "Capacity of the scan job queue for pending tasks")
+	scanJobFailRetryMax := flag.Int("scan_job_fail_retry_max", 5, "Maximum retry attempts for failed scan jobs")
+	staleScanJobCleanupIntervalHour := flag.Int("stale_scan_job_cleanup_interval_hour", 1, "Interval (in hours) for cleaning up stale scan jobs")
+
 	flag.Parse()
 
 	// default log_level is LogLevel_Info
@@ -970,8 +977,20 @@ func main() {
 	// To prevent crd webhookvalidating timeout need queue the crd and process later.
 	rest.CrdValidateReqManager()
 
+	// Each scanner can handle multiple requests concurrently, set 2 to avoid OOM.
+	rpc.ScannerMgr = rpc.NewScannerManager(*maxScannerTasks)
+
+	// These tasks are processed concurrently by up to 16 worker goroutines, ensuring efficient task handling.
+	rest.SetMaxConcurrentRepoScanTasks(*maxConcurrentRepoScanWorkers)
+
+	// Defines the buffer size for the work queue, allowing up to 2048 tasks to be queued simultaneously.
+	rest.SetScanJobQueueCapacity(*scanJobQueueCapacity)
+	rest.SetScanJobFailRetryMax(*scanJobFailRetryMax)
+	rest.SetStaleScanJobCleanupIntervalHour(*staleScanJobCleanupIntervalHour)
+
 	// start rest server
 	go rest.StartRESTServer(isNewCluster, Ctrler.Leader)
+	defer rest.RepoScanMgr.Shutdown() // RepoScanMgr is initialized in StartRESTServer; ensure it is closed to prevent goroutine leaks.
 
 	// go rest.StartLocalDevHttpServer() // for local dev only
 
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -279,6 +279,7 @@ func main() {
 	scanJobFailRetryMax := flag.Int("scan_job_fail_retry_max", 5, "Maximum retry attempts for failed scan jobs")
 	staleScanJobCleanupIntervalHour := flag.Int("stale_scan_job_cleanup_interval_hour", 1, "Interval (in hours) for cleaning up stale scan jobs")
 	repoScanLongPollTimeout := flag.Int("repo_scan_long_poll_timeout", 30, "Timeout for long polling repository scan jobs")
+	maxScannerLimitPerNode := flag.Int("max_scanner_limit_per_node", 128, "Maximum number of scanner tasks per node")
 	flag.Parse()
 
 	// default log_level is LogLevel_Info
@@ -978,7 +979,7 @@ func main() {
 	rest.CrdValidateReqManager()
 
 	// Each scanner can handle multiple requests concurrently, set 2 to avoid OOM.
-	rpc.ScanCreditMgr = rpc.NewScanCreditManager(*maxScannerTasks)
+	rpc.ScanCreditMgr = rpc.NewScanCreditManager(*maxScannerTasks, *maxScannerLimitPerNode)
 
 	// start rest server
 	go rest.StartRESTServer(
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -274,7 +274,7 @@ func main() {
 	autoProfile := flag.Int("apc", 1, "Enable auto profile collection")
 	custom_check_control := flag.String("cbench", share.CustomCheckControl_Disable, "Custom check control")
 	maxScannerTasks := flag.Int("max_scanner_tasks", 2, "Maximum number of concurrent tasks a scanner can handle")
-	maxConcurrentRepoScanTasks := flag.Int("max_concurrent_repo_scan_tasks", 16, "Maximum number of concurrent repository scan tasks")
+	maxConcurrentRepoScanTasks := flag.Int("max_concurrent_repo_scan_tasks", 32, "Maximum number of concurrent repository scan tasks")
 	scanJobQueueCapacity := flag.Int("scan_job_queue_capacity", 2048, "Capacity of the scan job queue for pending tasks")
 	scanJobFailRetryMax := flag.Int("scan_job_fail_retry_max", 5, "Maximum retry attempts for failed scan jobs")
 	staleScanJobCleanupIntervalHour := flag.Int("stale_scan_job_cleanup_interval_hour", 1, "Interval (in hours) for cleaning up stale scan jobs")
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -279,7 +279,7 @@ func main() {
 	scanJobFailRetryMax := flag.Int("scan_job_fail_retry_max", 5, "Maximum retry attempts for failed scan jobs")
 	staleScanJobCleanupIntervalHour := flag.Int("stale_scan_job_cleanup_interval_hour", 1, "Interval (in hours) for cleaning up stale scan jobs")
 	repoScanLongPollTimeout := flag.Int("repo_scan_long_poll_timeout", 30, "Timeout for long polling repository scan jobs")
-	maxScannerLimitPerNode := flag.Int("max_scanner_limit_per_node", 128, "Maximum number of scanner tasks per node")
+	scannerLBMax := flag.Int("scanner_lb_max", 128, "Maximum number of scanner per controller")
 	flag.Parse()
 
 	// default log_level is LogLevel_Info
@@ -979,7 +979,7 @@ func main() {
 	rest.CrdValidateReqManager()
 
 	// Each scanner can handle multiple requests concurrently, set 2 to avoid OOM.
-	rpc.ScanCreditMgr = rpc.NewScanCreditManager(*maxScannerTasks, *maxScannerLimitPerNode)
+	rpc.ScanCreditMgr = rpc.NewScanCreditManager(*maxScannerTasks, *scannerLBMax)
 
 	// start rest server
 	go rest.StartRESTServer(
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -274,11 +274,11 @@ func main() {
 	autoProfile := flag.Int("apc", 1, "Enable auto profile collection")
 	custom_check_control := flag.String("cbench", share.CustomCheckControl_Disable, "Custom check control")
 	maxScannerTasks := flag.Int("max_scanner_tasks", 2, "Maximum number of concurrent tasks a scanner can handle")
-	maxConcurrentRepoScanWorkers := flag.Int("max_concurrent_repo_scan_workers", 16, "Maximum number of concurrent repository scan workers")
+	maxConcurrentRepoScanTasks := flag.Int("max_concurrent_repo_scan_tasks", 16, "Maximum number of concurrent repository scan tasks")
 	scanJobQueueCapacity := flag.Int("scan_job_queue_capacity", 2048, "Capacity of the scan job queue for pending tasks")
 	scanJobFailRetryMax := flag.Int("scan_job_fail_retry_max", 5, "Maximum retry attempts for failed scan jobs")
 	staleScanJobCleanupIntervalHour := flag.Int("stale_scan_job_cleanup_interval_hour", 1, "Interval (in hours) for cleaning up stale scan jobs")
-
+	repoScanLongPollTimeout := flag.Int("repo_scan_long_poll_timeout", 30, "Timeout for long polling repository scan jobs")
 	flag.Parse()
 
 	// default log_level is LogLevel_Info
@@ -978,19 +978,18 @@ func main() {
 	rest.CrdValidateReqManager()
 
 	// Each scanner can handle multiple requests concurrently, set 2 to avoid OOM.
-	rpc.ScannerMgr = rpc.NewScannerManager(*maxScannerTasks)
-
-	// These tasks are processed concurrently by up to 16 worker goroutines, ensuring efficient task handling.
-	rest.SetMaxConcurrentRepoScanTasks(*maxConcurrentRepoScanWorkers)
-
-	// Defines the buffer size for the work queue, allowing up to 2048 tasks to be queued simultaneously.
-	rest.SetScanJobQueueCapacity(*scanJobQueueCapacity)
-	rest.SetScanJobFailRetryMax(*scanJobFailRetryMax)
-	rest.SetStaleScanJobCleanupIntervalHour(*staleScanJobCleanupIntervalHour)
+	rpc.ScanCreditMgr = rpc.NewScanCreditManager(*maxScannerTasks)
 
 	// start rest server
-	go rest.StartRESTServer(isNewCluster, Ctrler.Leader)
-	defer rest.RepoScanMgr.Shutdown() // RepoScanMgr is initialized in StartRESTServer; ensure it is closed to prevent goroutine leaks.
+	go rest.StartRESTServer(
+		isNewCluster,
+		Ctrler.Leader,
+		*maxConcurrentRepoScanTasks, // Max concurrent repository scan
+		*scanJobQueueCapacity,       // Capacity of scan job queue
+		*scanJobFailRetryMax,        // Max retries for failed scan jobs
+		time.Duration(*repoScanLongPollTimeout)*time.Second,       // Long poll timeout duration
+		time.Duration(*staleScanJobCleanupIntervalHour)*time.Hour, // Cleanup interval for stale jobs
+	)
 
 	// go rest.StartLocalDevHttpServer() // for local dev only
 
--- a/controller/rest/longpoll.go
+++ b/controller/rest/longpoll.go
@@ -373,9 +373,9 @@ type longpollManyMgr struct {
 	jobs                       map[interface{}]*longpollManyJob
 }
 
-func NewLongPollManyMgr(timeout, linger time.Duration, max, maxConcurrentRepoScanTasks int) *longpollManyMgr {
+func NewLongPollManyMgr(timeout, linger time.Duration, maxConcurrentRepoScanTasks int) *longpollManyMgr {
 	return &longpollManyMgr{
-		max:                        max,
+		max:                        maxConcurrentRepoScanTasks,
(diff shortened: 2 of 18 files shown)


controller/rest/repository.go
```diff
@@ -22,7 +22,7 @@ import (
 
 const (
 	repositoryDefaultTag      = "latest"
-	maxRepoScanTasks          = 32
+	maxRepoScanTasks          = 4
```
pohanhuangtw, line 25: maxRepoScanTasks determines the maximum number of scan tasks that can run concurrently.
We've set it to 4 after observing that each scan task consumes approximately 500 MB of memory.
Running 32 concurrent scans could consume up to 16 GB, leading to Out-Of-Memory (OOM) issues.
Therefore, limiting it to 4 is a safer choice to prevent excessive memory usage.

holyspectral:
Is there an issue analysis that I can refer to?

controller/rest/repository.go
```diff
@@ -21,11 +21,13 @@ import (
...
-	repoScanTimeout           = time.Minute * 20
-	repoScanLingeringDuration = time.Second * 30
-	repoScanLongPollTimeout   = time.Second * 30
+	repositoryDefaultTag = "latest"
+	// maxConcurrentRepoScanTasks determines the maximum number of scan tasks that can run concurrently.
+	// Limiting it to 4 is a safer choice to prevent excessive memory usage.
+	maxConcurrentRepoScanTasks = 4
+	repoScanTimeout            = time.Minute * 20
+	repoScanLingeringDuration  = time.Second * 30
+	repoScanLongPollTimeout    = time.Second * 30
```
holyspectral, line 30: Make these configurable?

controller/rest/longpoll.go
```diff
@@ -4,11 +4,17 @@ import (
...
+
+	log "github.com/sirupsen/logrus"
 )
 
 var errTooManyJobs = errors.New("Too many concurrent jobs")
 var errDuplicateJob = errors.New("Duplicate job")
 
+// Defines the buffer size for the work queue, allowing up to 2048 tasks to be queued simultaneously.
+// These tasks are processed concurrently by up to 4 worker goroutines, ensuring efficient task handling.
+var workQueueBufferSize = 2048
```
holyspectral, line 16: Make this configurable?

controller/rest/longpoll.go
```diff
@@ -52,54 +58,77 @@ func (j *longpollOnceJob) Poll() (interface{}, time.Duration) {
...
+			case <-time.After(m.linger):
+			case <-job.polled:
+			}
+			m.mux.Lock()
+			delete(m.jobs, workItem.key)
+			m.mux.Unlock()
+		} else {
+			log.WithFields(log.Fields{
+				"key": workItem.key,
+			}).Error("Job not found in manager")
```
holyspectral, line 95: Is this case supposed to happen?  There might be others, but from what I can see, in this path, `workItem.task.Run()` will not run, so `longpollOnceJob` would timeout.

controller/rest/longpoll.go
```diff
@@ -52,54 +58,77 @@ func (j *longpollOnceJob) Poll() (interface{}, time.Duration) {
...
-	max     int
-	timeout time.Duration
-	linger  time.Duration
-	jobs    map[interface{}]*longpollOnceJob
+	mux       sync.RWMutex
+	poolSize  int
+	timeout   time.Duration
+	linger    time.Duration
+	jobs      map[interface{}]*longpollOnceJob // use as a queue
+	workQueue chan PollWorkItem
```
holyspectral, line 73: `jobs` and `workQueue` seem to serve similar purposes.  Have you considered replacing `jobs` with `workQueue`?

holyspectral (commented):
Some comments/questions in high level.

controller/rest/longpoll.go
```diff
@@ -52,54 +58,77 @@ func (j *longpollOnceJob) Poll() (interface{}, time.Duration) {
...
-			case <-time.After(m.linger):
-			case <-job.polled:
-			}
-
-			m.mux.Lock()
-			delete(m.jobs, key)
-			m.mux.Unlock()
-		}()
-
+		m.workQueue <- PollWorkItem{key: key, args: args, task: task}
```
holyspectral, line 131: Sorry if the indent is not correct:
```suggestion
                    select {
                        case m.workQueue <- PollWorkItem{key: key, args: args, task: task}
                        default:
                            log.Error("workQueue is full")
                    }
```

controller/rest/longpoll.go
pohanhuangtw: Hi @holyspectral ,
I considered this approach before; however, there's a tricky case where the user continuously sends the same event. In this scenario, the controller needs a way to recognize that it's a duplicate job. This is why we require a map to keep track of which jobs have already been processing.

holyspectral: Is it possible for users to NOT send duplicate scan requests?

controller/rest/longpoll.go
pohanhuangtw: You are correct, it will just timeout.
Just want to log it for reference.

controller/rest/repository.go
pohanhuangtw: Do you mean config in the controller entry point?

controller/rest/longpoll.go
holyspectral: Overall I think the interaction between `jobs` and `workQueue` has to be thought through carefully.  
For example, when call#1 and call#2 calls the same API twice with the same payload, `m.jobs[key]` would be overwritten by the job that comes later.

However, when `pollWorker()` really starts the job, there is no guarantee that which job will be used, so this will make one of the request time out.  

If you want to keep both, you will have to be very careful to deal with this kind of issues.

controller/rest/longpoll.go
```diff
@@ -52,54 +58,77 @@ func (j *longpollOnceJob) Poll() (interface{}, time.Duration) {
...
+		job, exists := m.jobs[workItem.key]
+		m.mux.RUnlock()
+
+		if exists {
+			job.dataCh <- workItem.task.Run(workItem.args)
+			// In case polling returned, wait a while before discarding the record
+			select {
+			case <-time.After(m.linger):
+			case <-job.polled:
+			}
```
holyspectral, line 88: Why is there a need to wait before removing a completed job?

holyspectral (commented):
Thanks.  Now I have the general idea about how this works.
I'd recommend making sure that `errTooManyJobs` error is still generated (not just log the error) and making those variables configurable via command line, environment variable or other methods.  They would be handy when you want to troubleshoot.  

Also, [assert](https://pkg.go.dev/github.com/stretchr/testify/assert) could make your unit test much simpler.  
Before
```
	if errDuplicateJob.Error() != "Duplicate job" {
		t.Errorf("Expected errDuplicateJob to be 'Duplicate job', got '%s'", errDuplicateJob.Error())
	}
```
With assert
```
	assert.Equal(t, "Duplicate job", errDuplicateJob.Error())
```

controller/rest/longpoll.go
pohanhuangtw: I think you are right, the existing design to make sure the job delete after some second => which should wait for the job to be done.

we should remove this.

holyspectral (commented):
Looks like you're trying to implement a semaphore?  I'd need some time to look into the detail, but here is what I have right now.  (Nothing major except the input parameter)

Also, have you tried `go test -race` in your test?  It will run the test using golang built-in race detector and it's generally very useful.

controller/rpc/scanner.go
```diff
@@ -15,90 +15,142 @@ import (
...
+		maxConns:       maxConns,
+		availableCh:    make(chan struct{}, defaultScannerCount*maxConns),
+	}
+}
 
-func AddScanner(scanner *share.CLUSScanner) {
-	scanMutex.Lock()
-	defer scanMutex.Unlock()
-	scanners[scanner.ID] = &scannerAct{scanner: scanner}
+func (mgr *ScannerManager) signalAvailability() {
```
holyspectral, line 43: Can you elaborate more about the purpose of `availableCh`?  I'm not sure if I fully understand, but given that `signalAvailability()` would block when `defaultScannerCount*maxConns` is reached, I'm sure I definitely missed something.

controller/controller.go
```diff
@@ -272,6 +273,18 @@ func main() {
...
+	// Each scanner can handle multiple requests concurrently, set 2 to avoid OOM.
+	rpc.NewScannerManager(*flag.Int("max_conns", 2, "Maximum number of scanner tasks that can run concurrently."))
+
+	// These tasks are processed concurrently by up to 16 worker goroutines, ensuring efficient task handling.
+	rest.SetMaxConcurrentRepoScanTasks(*flag.Int("max_concurrent_repo_scan_tasks", 16, "Maximum number of repository scan workers that can run concurrently."))
+
+	// Defines the buffer size for the work queue, allowing up to 2048 tasks to be queued simultaneously.
+	rest.SetJobQueueCapacity(*flag.Int("job_queue_capacity", 2048, "Maximum capacity of the job queue."))
+	rest.SetJobFailRetryMax(*flag.Int("job_fail_retry_max", 5, "Maximum number of retries allowed for a failed job."))
+	rest.SetStaleJobCleanupInterval(*flag.Int("stale_job_cleanup_interval", 1, "Interval (in hours) for cleaning up stale jobs."))
```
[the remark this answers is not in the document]
pohanhuangtw: I forget to add it in the monitor.c, thanks

pohanhuangtw:
Hi @holyspectral,

> Can you elaborate more about the purpose of availableCh?

Thanks for the brief review! Let me elaborate on the design of availableCh. We operate under the assumption that users will configure their cluster with no more than `scannerChannelCapacity` scanners.

The current design functions like a semaphore, where availableCh acts as a signaling mechanism to manage scanner availability. Generally, when a failure occurs in getClient or DoneScanning, the decScanningCount function is invoked, which subsequently triggers signalAvailability.

However, there are edge cases to consider:

If there are no scanners available at all, getAvailableScanner will block indefinitely while waiting on availableCh. Is this behavior acceptable in such scenarios?
To improve this, I could enhance the semaphore-like mechanism by adding a timeout, preventing the system from waiting too long in these edge cases. What are your thoughts on this approach?

pohanhuangtw:
Hi @holyspectral 

- I use os.Getenv the value to read the data
- I add the timeout in getavailScanner to avoid blocking

controller/rpc/scanner.go
```diff
@@ -15,90 +15,147 @@ import (
...
 
 	// endpoint is also the key
-	endpoint := getScannerEndpoint(s.scanner)
+	endpoint := mgr.getScannerEndpoint(s.scanner)
 	if cluster.GetGRPCClientEndpoint(endpoint) == "" {
-		if err := cluster.CreateGRPCClient(endpoint, endpoint, true, createScannerServiceWrapper); err != nil {
+		if err := cluster.CreateGRPCClient(endpoint, endpoint, true, mgr.createScannerServiceWrapper); err != nil {
 			log.WithFields(log.Fields{"error": err}).Error("CreateGRPCClient")
+			s.activeScannerTasks--
+			mgr.signalAvailability()
```
holyspectral, line 136: Can you elaborate a little more about why we should signal availability here?

controller/rpc/scanner.go
```diff
@@ -15,90 +15,147 @@ import (
...
-	scanMutex.Unlock()
+	mgr.mutex.Unlock()
 
 	// endpoint is also the key
 	if s != nil {
-		endpoint := getScannerEndpoint(s.scanner)
+		endpoint := mgr.getScannerEndpoint(s.scanner)
 		cluster.DeleteGRPCClient(endpoint)
 	}
 }
```
holyspectral, line 72: I can see the availability will increase when we have a new scanner, but do we need to decrease it when a scanner is removed?

controller/rpc/scanner.go
```diff
@@ -186,17 +221,20 @@ func ScanRunning(scanner string, agentID, id string, objType share.ScanObjectTyp
 		return nil, fmt.Errorf("Cannot find enforcer endpoint")
 	}
 
-	client, err := getScannerServiceClient(scanner, true)
+	client, scanner, err := ScannerMgr.getScannerServiceClient(scanner, true)
 	if err != nil {
 		return nil, err
 	}
-	defer decScanningCount(scanner)
+	defer ScannerMgr.decScanningCount(scanner)
```
holyspectral, line 228: Is availableCh count balanced in this function?

controller/rpc/scanner.go
```diff
@@ -208,11 +246,11 @@ func ScanRunning(scanner string, agentID, id string, objType share.ScanObjectTyp
 }
 
 func ScanPlatform(scanner string, k8sVersion, ocVersion string, timeout time.Duration) (*share.ScanResult, error) {
-	client, err := getScannerServiceClient(scanner, true)
+	client, scanner, err := ScannerMgr.getScannerServiceClient(scanner, true)
 	if err != nil {
 		return nil, err
 	}
-	defer decScanningCount(scanner)
+	defer ScannerMgr.decScanningCount(scanner)
```
holyspectral, line 253: Is availabeCh balanced in this function?

controller/rpc/scanner.go
```diff
@@ -361,23 +398,25 @@ func ScanAwsLambdaFunc(ctx context.Context, funcInput *share.CLUSAwsFuncScanInpu
...
 func ScanCacheGetStat(scanner string) (*share.ScanCacheStatRes, error) {
 	log.WithFields(log.Fields{"scanner": scanner}).Debug()
-	client, err := getScannerServiceClient(scanner, true)
+	client, scanner, err := ScannerMgr.getScannerServiceClient(scanner, true)
 	if err != nil {
 		return nil, err
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
+	defer ScannerMgr.decScanningCount(scanner)
```
holyspectral, line 407: Is availabeCh balanced in this function?

controller/rpc/scanner.go
```diff
@@ -361,23 +398,25 @@ func ScanAwsLambdaFunc(ctx context.Context, funcInput *share.CLUSAwsFuncScanInpu
...
 
 func ScanCacheGetData(scanner string) (*share.ScanCacheDataRes, error) {
-	client, err := getScannerServiceClient(scanner, true)
+	client, scanner, err := ScannerMgr.getScannerServiceClient(scanner, true)
 	if err != nil {
 		return nil, err
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
+	defer ScannerMgr.decScanningCount(scanner)
```
holyspectral, line 419: Is availabeCh balanced in this function?

holyspectral (commented):
Thanks for the explanation, @pohanhuangtw !
I have a feeling that the channel `availableCh` is not balanced.  Apparently we call `signalAvailability()` more than `getAvailableScanner()`.  Can you double check?

If this happened, because you use `select` in `signalAvailability()`, there would probably be no significant adverse effect, but `getAvailableScanner()` would always exit immediately.  Not sure if this is what you want.

holyspectral:
Also it would be a good idea to name `signalAvailability()` and `getAvailableScanner()` similarly, e.g., `releaseScannerAvailability()` and `acquireScannerAvailability()`.  This way it will be easier to make sure that we call them in pair.

pohanhuangtw:
Hi @holyspectral,
Thank you for highlighting the critical issue regarding the imbalance in signaling. To address this, I have adjusted the Scanner Manager's signaling mechanism. Now, only functions directly related to getAvailableScanner can signal or cancel signals, ensuring better control and balance

pohanhuangtw:
Hi @holyspectral,
Could you please review the releaseScannerAvailability() and acquireScannerAvailability() functions? These methods handle the releasing and acquiring of scanners and ensure that availableCh remains balanced.

I have also written corresponding tests for these functions in scanner_test.go, which include:
1. [TestRemoveScanner](https://github.com/neuvector/neuvector/blob/dcf344cbcc84695f72e1b0eb4acc775b557ae876/controller/rpc/scanner_test.go#L78)
2. [TestAcquireScannerAvailability](https://github.com/neuvector/neuvector/blob/dcf344cbcc84695f72e1b0eb4acc775b557ae876/controller/rpc/scanner_test.go#L139C6-L139C36)
3. [TestReleaseScannerAvailability](https://github.com/neuvector/neuvector/blob/dcf344cbcc84695f72e1b0eb4acc775b557ae876/controller/rpc/scanner_test.go#L172C6-L172C36)

Your feedback would be greatly appreciated!

controller/controller.go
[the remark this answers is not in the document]
jayhuang-suse: ok

controller/rest/repository.go
```diff
@@ -194,43 +253,51 @@ func handlerScanRepositoryReq(w http.ResponseWriter, r *http.Request, ps httprou
...
-		if ret.errCode == api.RESTErrClusterRPCError {
-			restRespError(w, http.StatusInternalServerError, ret.errCode)
-		} else if ret.errCode != 0 {
-			restRespErrorMessage(w, http.StatusInternalServerError, ret.errCode, ret.errMsg)
-		} else {
-			resp := &api.RESTScanRepoReportData{Report: ret.report}
-			restRespSuccess(w, r, resp, acc, login, &data, "Request repository scan")
-		}
+		resp := &api.RESTScanRepoReportData{Report: ret.report}
+		restRespSuccess(w, r, resp, acc, login, &data, "Repository scan request completed successfully")
```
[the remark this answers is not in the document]
jayhuang-suse: ok

controller/rest/longpoll_test.go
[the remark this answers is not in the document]
jayhuang-suse: ok

jayhuang-suse (commented):
Please confirm the pretest results of QA scan suite.

controller/rpc/scanner.go
```diff
@@ -289,18 +361,18 @@ func ScanImage(scanner string, ctx context.Context, req *share.ScanImageRequest)
...
 		return nil, err
 	}
 
-	client, err := getScannerServiceClient(scanner, true)
+	client, scanner, err := ScannerMgr.getScannerServiceClient(scanner, true)
 	if err != nil {
 		return nil, err
 	}
-	defer decScanningCount(scanner)
+	defer ScannerMgr.releaseScannerAvailability(scanner)
```
holyspectral, line 375: This line should be moved to after `scanner := ScannerMgr.acquireScannerAvailability()`, otherwise the scanner availability would not be increased again if it returns between two calls.

controller/rpc/scanner.go
```diff
@@ -332,13 +404,13 @@ func ScanAwsLambdaFunc(ctx context.Context, funcInput *share.CLUSAwsFuncScanInpu
...
 
-	client, err := getScannerServiceClient(scanner, true)
+	client, scanner, err := ScannerMgr.getScannerServiceClient(scanner, true)
 	if err != nil {
 		err := fmt.Errorf("No scan client available")
 		log.WithFields(log.Fields{"error": err}).Error()
 		return nil, err
 	}
-	defer decScanningCount(scanner)
+	defer ScannerMgr.releaseScannerAvailability(scanner)
```
holyspectral, line 413: This line should be moved to after `scanner := ScannerMgr.acquireScannerAvailability()`, otherwise the scanner availability would not be increased again if it returns between two calls.

controller/rpc/scanner.go
```diff
@@ -256,21 +317,32 @@ func ScanPlatform(scanner string, k8sVersion, ocVersion string, timeout time.Dur
...
 	if err != nil {
 		return nil, err
 	}
-	defer decScanningCount(scanner)
+
+	// Ensure that resources are properly released or task counts are decremented when the function exits
+	defer func() {
+		if hasAcquiredScanner {
+			// Release the scanner slot back to the availability channel
+			ScannerMgr.releaseScannerAvailability(scanner)
```
holyspectral, line 340: You should call  `defer` as soon as you have a resource to release later, otherwise the scanner availability would not be increased again if it returns between two calls.

controller/rpc/scanner.go
```diff
@@ -15,90 +15,170 @@ import (
...
+	activeScanners map[string]*scannerAct
+	maxConns       int
+	mutex          sync.RWMutex
+	availableCh    chan struct{}
+}
+
+var ScannerMgr *ScannerManager
+
+// scannerChannelCapacity ensures that the availableCh channel has sufficient capacity to handle task signals without blocking.
+var scannerChannelCapacity = 15
```
holyspectral, line 32: Can we double confirm if `15` is an appropriate maximum number of scanners? Because if users have more than 15 scanners, the `AddScanner()` would cause imbalance to happen.  Say you have 3 scanners, then 100, then 3.  In this case, it would make #item in availableCh become 3*2, 3*15, then 0, if I was not mistaken.

controller/rpc/scanner.go
```diff
@@ -15,90 +15,170 @@ import (
...
+	activeScannerTasks int // Indicate how much scanner task is currently run
 }
 
-var scanners map[string]*scannerAct = make(map[string]*scannerAct)
-var scanMutex sync.RWMutex
+type ScannerManager struct {
+	activeScanners map[string]*scannerAct
+	maxConns       int
+	mutex          sync.RWMutex
+	availableCh    chan struct{}
```
holyspectral, line 26: minor: I think your design is basically a credit system.  When a new scanner is added, it gives you credits to run scan.  When a scanner is removed, those credits are removed again.  Each scan job requires 1 credit in order to run.  Not sure if it's a good idea to give it a better naming.

holyspectral (changes requested):
There are still some potential issues spotted.  Please take a look and let me know if they're not valid.

controller/rpc/scanner.go
pohanhuangtw: Hi @jayhuang-suse ,
About this, can you help confirm the maximum number of scanners? do we have a limit on this?

pohanhuangtw:
Hi @holyspectral @jayhuang-suse ,
The PR has updated with

1. add scanner load balancer to balance scanner usage, cover with unit test, and scanner manager concurrent add/remove/release test.
2. configure scanner max limit per node.
3. add new package for scanner lb.

Please help review, thanks.

pohanhuangtw:
Hi @holyspectral @jayhuang-suse ,
Based on the previous discuss, the only thing we needs to check now is controller/scannerlb/scanner_load_balancer.go.

it's a btree(for finding min fast)+map(for efficiently add/delete node), to make sure the scanner can evenly distributed. 

thanks for the help.

hhcs9527:
Hi @holyspectral , @jayhuang-suse .
The Overall changes has made, the remaining issue is the load balancer suggested by @jayhuang-suse 
Please take a look, thanks

jayhuang-suse (approved):
LGTM

pohanhuangtw:
Hi @holyspectral ,
The code has been approved by @jayhuang-suse , the main logic has no change, please take a look, thanks.

holyspectral (approved):
There might still be things to improve, but we probably don’t want to spend more time on this.  

I’m approving it now, so it won’t get blocked here indefinitely.

PR #1660, opened 2024-11-13
NVSHAS-9645: pre-existing CRD processing fails

williamlin-suse:
It is caused by incorrect type conversion introduced in 5.4.0 only


jeffhuang4704 (approved):
Looks good to me.

PR #1662, opened 2024-11-14
NVSHAS-9651: Quay.io needs user name and password for scanning

williamlin-suse:
It is because Get(https://quay.io/v2/) always return 401 that we should not always treat it as failure


jayhuang-suse (approved):
LGTM

PR #1664, opened 2024-11-15
NVSHAS-9592: requeue in-progress workloads on db update

alopez-suse:
In progress workloads would cause the DB updates to be "missed".

### DIFF
--- a/controller/cache/scan.go
+++ b/controller/cache/scan.go
@@ -619,7 +619,13 @@ func scannerDBChange(newVer string) {
 	go func() {
 		scanMutexLock()
 		for id, info := range scanMap {
-			if info.status == statusScanNone && info.version != newVer {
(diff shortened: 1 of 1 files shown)


controller/cache/scan.go
```diff
@@ -619,7 +619,13 @@ func scannerDBChange(newVer string) {
 	go func() {
 		scanMutexLock()
 		for id, info := range scanMap {
-			if info.status == statusScanNone && info.version != newVer {
+			if info.version == newVer { // no need for new scan
+				continue
+			}
+			if info.status == statusScanning || info.status == statusScanNone {
```
williamlin-suse, line 625: This fix means the reported issue only occurres to workloads that are being scanned when DB version change.
Is this what customer encountered?

williamlin-suse (approved):

PR #1668, opened 2024-11-18
[NVSHAS-9507] Fix JFrog fetch tag issue.

pohanhuangtw:
### Summary

- The old API for fetching JFrog resources encounters issues with the newer jfrog version.
- Created a backup function specifically for handling JFrog tag retrieval.
- Ensures compatibility with JFrog's updated API while preserving existing registry functionality.

### Test perform

- repository scan with jfrog
- reposiorty scan with docker

### DIFF
--- a/controller/scan/jfrog.go
+++ b/controller/scan/jfrog.go
@@ -216,7 +216,7 @@ func getSubdomainFromRepo(repo string) (string, string) {
 	return "", ""
 }
 
-// Only be called by in jfrog, since the api is not working for it, create it as a callback function.
+// GetArtifactoryTags is designed to work with a new API endpoint provided by JFrog.
 func (r *jfrog) GetArtifactoryTags(repositoryStr string, rc *scanUtils.RegClient) ([]string, error) {
 	tags := make([]string, 0)
 	key, repository, found := strings.Cut(repositoryStr, "/")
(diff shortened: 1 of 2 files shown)


controller/scan/jfrog.go
```diff
@@ -232,7 +249,12 @@ func (r *jfrog) GetTagList(domain, repo, tag string) ([]string, error) {
 		}
 		repo = subRepo
 	}
-	return rc.Tags(repo)
+
+	if tags, err := rc.Tags(repo); len(tags) > 0 {
+		return tags, err
+	}
```
holyspectral, line 255: Shall we check the error as well?

controller/scan/jfrog.go
```diff
@@ -216,6 +216,23 @@ func getSubdomainFromRepo(repo string) (string, string) {
...
+	tags := make([]string, 0)
+	var err error
+	parts := strings.Split(repositoryStr, "/")
+	if len(parts) != 2 {
+		log.WithFields(log.Fields{"repository": repositoryStr}).Error("Invalid input format")
+		return tags, err
+	}
+
+	key := parts[0]
+	repository := parts[1]
```
holyspectral, line 230:
```suggestion
	key, repository := strings.Cut(repositoryStr, "/")
```

pohanhuangtw: Thanks!

controller/scan/jfrog.go
pohanhuangtw: I don't think so. We can log this as a debug message; it typically just return the error and let the outer function handle it.

@holyspectral what do you think?

holyspectral: My point is that when `len(tags) == 0`, the `err` will be ignored.  Errors should be handled all the time because it can be something unexpected, where fallback might not be appropriate.

pohanhuangtw: I see, just add new error check

controller/scan/jfrog.go
```diff
@@ -250,7 +245,7 @@ func (r *jfrog) GetTagList(domain, repo, tag string) ([]string, error) {
 		repo = subRepo
 	}
 
-	if tags, err := rc.Tags(repo); len(tags) > 0 {
+	if tags, err := rc.Tags(repo); len(tags) > 0 || err != nil {
```
holyspectral, line 248: The only case where you need to fallback is the `len(tags) == 0 && err == nil`?  If so, I recommend leaving a comment.

pohanhuangtw: Thanks, I just left comment and interchange the call order, thanks

holyspectral (approved):
Looks good to me.  Maybe include the change in release notes?

jeffhuang4704 (commented):
looks good to me.

PR #1670, opened 2024-11-19
NVSHAS-9568:replace built-in fanotify with upstream library

jayhuang-suse:
Update fanotify package with MIT-licensed upstream repo


agent/probe/faccess_linux.go
```diff
@@ -818,25 +822,34 @@ func (fa *FileAccessCtrl) processEvent(ev *fsmon.EventMetadata) bool {
...
-	if events, err := fa.fanfd.GetEvents(); err == nil {
-		for _, ev := range events {
-			if dbgError := fa.fanfd.Response(ev, fa.processEvent(ev)); dbgError != nil {
-				log.WithFields(log.Fields{"dbgError": dbgError}).Debug()
-			}
-			ev.File.Close()
+	for {
+		ev, err := fa.fanfd.GetEvent(os.Getpid())
+		if err != nil || ev == nil {
+			return
```
yasker, line 832: Error handling?

share/fsmon/fanotify_linux.go
```diff
@@ -560,53 +563,62 @@ func (fn *FaNotify) MonitorFileEvents() {
...
+		} else { // monitor or protect mode
+			allowRead := resp && !change
+			bReporting = !allowRead // allowed app by block_access
+		}
+
+		if bReporting || change { // report changed file
+			ifile.cb(ifile.path, mask, ifile.params, pInfo)
 		}
 	}
-	return nil
```
yasker, line 609: Hmm, why are we removing this line?

yasker (changes requested):

share/fsmon/fanotify_linux.go
jayhuang-suse: My local lint complaints that it was unreachable "return" in this function.

agent/probe/faccess_linux.go
jayhuang-suse: Updated.

yasker (approved):

agent/probe/faccess_linux.go
```diff
@@ -857,7 +870,10 @@ func (fa *FileAccessCtrl) monitorFilePermissionEvents() {
 		}
 
 		if (pfd[0].Revents & unix.POLLIN) != 0 {
-			fa.handleEvents()
+			if err := fa.handleEvents(); err != nil && err != unix.EINTR {
+				log.WithFields(log.Fields{"err": err}).Error("FA: handle")
+				break
+			}
```
holyspectral, line 876: In the original logic, the error from `fa.handleEvents()` will not stop the for loop.  In this new version, when error is not EINTR, the flow will leave the for loop as well as `fa.monitorFilePermissionEvents()` go routine.  Is it expected?

holyspectral (commented):
Generally looks good to me.  Just a question about error handling.

agent/probe/faccess_linux.go
jayhuang-suse: It was tested by the file monitor section and is okay if we need an error handler as Sheng requested.

holyspectral (approved):

PR #1671, opened 2024-11-19
Remove unused property remote_permissions in Swagger documentation of RESTUser

talavis:
Based on the API definition in `controller/api/apis.go`, `remote_permissions` is not a property of `RESTUser`.

As it does not exist in the actual API, I think it makes sense to remove it from the Swagger documentation.

### DIFF
--- a/controller/api/apis.yaml
+++ b/controller/api/apis.yaml
@@ -11702,11 +11702,6 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/RESTRolePermission'
-      remote_permissions:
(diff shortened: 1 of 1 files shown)


williamlin-suse:
remote_permissions is not a property of RESTUser.
it's remote_role_permissions which is defined in RESTUser.

talavis:
As already noted that line defines `remote_role_permissions`, not `remote_permissions`. There is a separate `remote_role_permissions` property in the Swagger file.

holyspectral (approved):
LGTM

PR #1675, opened 2024-11-21
NVSHAS-9678:excessive errors after linter change

jayhuang-suse:
(1) Ignore error traces when the container exited. 
(2) Ignore the supplemental information errors if the file does not exist.

### DIFF
--- a/agent/probe/faccess_linux.go
+++ b/agent/probe/faccess_linux.go
@@ -2,6 +2,7 @@ package probe
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -200,7 +201,9 @@ func (fa *FileAccessCtrl) addDirMarks(pid int, dirs []string) (bool, int) {
 		path := ppath + dir
 		err := fa.fanfd.Mark(unix.FAN_MARK_ADD, fa.cflag|unix.FAN_EVENT_ON_CHILD, unix.AT_FDCWD, path)
 		if err != nil {
-			log.WithFields(log.Fields{"path": path, "error": err}).Error("FA: ")
+			if !os.IsNotExist(errors.Unwrap(err)) {
+				log.WithFields(log.Fields{"path": path, "error": err}).Error("FA: ")
+			}
 		} else {
 			log.WithFields(log.Fields{"path": path}).Debug("FA: ")
 		}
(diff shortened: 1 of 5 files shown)


agent/probe/faccess_linux.go
```diff
@@ -215,8 +218,8 @@ func (fa *FileAccessCtrl) removeDirMarks(pid int, dirs []string) int {
 	ppath := fmt.Sprintf(procRootMountPoint, pid)
 	for _, dir := range dirs {
 		path := ppath + dir
-		if dbgError := fa.fanfd.Mark(unix.FAN_MARK_REMOVE, fa.cflag|unix.FAN_EVENT_ON_CHILD, unix.AT_FDCWD, path); dbgError != nil {
-			log.WithFields(log.Fields{"dbgError": dbgError}).Debug()
+		if err := fa.fanfd.Mark(unix.FAN_MARK_REMOVE, fa.cflag|unix.FAN_EVENT_ON_CHILD, unix.AT_FDCWD, path); err != nil && !os.IsNotExist(errors.Unwrap(err)) {
```
williamlin-suse, line 221: why use
`errors.Unwrap(err)`
here?

agent/probe/faccess_linux.go
```diff
@@ -225,27 +228,27 @@ func (fa *FileAccessCtrl) removeDirMarks(pid int, dirs []string) int {
...
-		log.WithFields(log.Fields{"dbgError": dbgError}).Debug()
-	}
-	if err != nil {
-		log.WithFields(log.Fields{"error": err}).Debug("FA: not supported")
+	if err := fa.fanfd.Mark(unix.FAN_MARK_ADD, unix.FAN_OPEN_PERM, unix.AT_FDCWD, path); err != nil {
+		log.WithFields(log.Fields{"error": err}).Info("FA: not supported")
 		return false
 	}
+
+	if err := fa.fanfd.Mark(unix.FAN_MARK_REMOVE, unix.FAN_OPEN_PERM, unix.AT_FDCWD, path); err != nil && !os.IsNotExist(errors.Unwrap(err)) {
```
williamlin-suse, line 236: why use
`errors.Unwrap(err)`
here?

agent/probe/faccess_linux.go
```diff
@@ -225,27 +228,27 @@ func (fa *FileAccessCtrl) removeDirMarks(pid int, dirs []string) int {
...
-		log.WithFields(log.Fields{"dbgError": dbgError}).Debug()
-	}
-	if err != nil {
-		log.WithFields(log.Fields{"error": err}).Debug("FA: not supported")
+	if err := fa.fanfd.Mark(unix.FAN_MARK_ADD, unix.FAN_OPEN_EXEC_PERM, unix.AT_FDCWD, path); err != nil {
+		log.WithFields(log.Fields{"error": err}).Info("FA: not supported")
 		return false
 	}
+
+	if err := fa.fanfd.Mark(unix.FAN_MARK_REMOVE, unix.FAN_OPEN_EXEC_PERM, unix.AT_FDCWD, path); err != nil && !os.IsNotExist(errors.Unwrap(err)) {
```
williamlin-suse, line 249: why use
`errors.Unwrap(err)`
here?

agent/probe/faccess_linux.go
jayhuang-suse: The error source is built by vendor, like 
  ` fmt.Errorf("fanotify: failed to close Fd: %w", err)`
It needs to be unwrapped before matching the real error code.

agent/probe/faccess_linux.go
jayhuang-suse: Same above reason

agent/probe/faccess_linux.go
jayhuang-suse: Same above reason

yasker (approved):

A side note: it’s better if we comment it to explain why we skip nonexist errors but there are too many of them. The good thing is we’ve commented on the commit so the reader should be able to find out the reason later.

williamlin-suse (approved):

PR #1676, opened 2024-11-22
fix: NVSHAS-9443 ArgoCD issue

holyspectral:
Do not check the lease object's presence, so it can be created correctly on ArgoCD.


jayhuang-suse (approved):
LGTM

PR #1678, opened 2024-11-26
NVSHAS-9667: Setting CTRL_PATH_DEBUG env to error in controller deplo…

williamlin-suse:
…yment is not working

For connection/mutex logging, enable debug_level only when it is specified by cli


jeffhuang4704 (approved):
looks good to me.

holyspectral (approved):

PR #1679, opened 2024-11-27
[NVSHAS-9683] docker has longer timeout

pohanhuangtw:
### Summary

- docker has longer timeout, thus removing the shorter timeout


jayhuang-suse (approved):
LGTM

PR #1691, opened 2024-12-09
NVSHAS-9669: Overall security score through REST API

williamlin-suse:
It sets the required permission for calling GET(v1/system/score/metrics)


jayhuang-suse (approved):
Good

PR #1692, opened 2024-12-10
feat: NVSHAS-9490 support controller SLSA L3

holyspectral:
1. Provide cross platform/standalone Dockerfile for controller and enforcer.
2. Release.yml to publish SLSA-capable artifacts
3. Make sure third party checksum is checked.
4. Provide build target, test-image, build-image and push-image to sync with rancher.
5. Switch to golang:1.22 as its base image


package/build_controller.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+set -e
+
+STAGE_DIR=stage
+
+machine=$(uname -m)
+echo "Machine hardware architecture is \"$machine\""
+
+if [ "$machine" == "x86_64" ]; then
```
pohanhuangtw, line 9: NV only run unit-test in x86_64 env? how about arm?

holyspectral: In SLSA we use emulated ARM64 build, in which unit test is very slow, so we only run unit test on amd64.  We will revisit once GitHub's arm64 node becomes generally available.

pohanhuangtw (approved):
LGTM

package/build_controller.sh
pjbgf: We should decouple the testing from the building script - that way it is not cross-emulated.
A new workflow for tests can be created, which simply run tests at both `push`/`pull_request` events.

pjbgf (approved):

BinX-Suse (approved):

PR #1693, opened 2024-12-11
NVSHAS-9653: learning wrong process rules

jayhuang-suse:
The reported process incident should not be learned even it changes its process name on the same pid.


holyspectral (commented):
Generally looks good to me.  Does this happen when a suspicious process is detected in learning mode and `allowSuspicious == false` ?

jayhuang-suse:
It covers several situations. Basically, it will override the same PID cases during Monitor and Discover modes. The same PID has been validated as a violated incident and we should treat its next caller as an incident, too.

holyspectral (approved):

PR #1694, opened 2024-12-12
[NVSHAS-9681] Scanner Registration Timeout Issue: Increased database slots from 256 to 512.

pohanhuangtw:
### Summary

- The previous size of 128 was insufficient, causing timeouts during scanner registration with the controller.

### DIFF
--- a/controller/grpc.go
+++ b/controller/grpc.go
@@ -35,6 +35,10 @@ import (
 
 // const scanImageDataTimeout = time.Second * 45
 const repoScanTimeout = time.Minute * 20
+const (
+	dbSlotsBase = 256
+	dbSlotsMax  = 512
+)
 
 type ScanService struct {
 }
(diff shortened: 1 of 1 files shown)


jayhuang-suse (approved):
LGTM

controller/grpc.go
```diff
@@ -66,7 +70,7 @@ func (ss *ScanService) preprocessDB(data *share.ScannerRegisterData) map[string]
 
 func (ss *ScanService) prepareDBSlots(data *share.ScannerRegisterData, cvedb map[string]*share.ScanVulnerability) ([][]byte, error) {
 	// As of now, Feb. 2019, the compressed db size is 3M, while max kv value size is 512K.
```
holyspectral, line 72: Should we also update the comment here?

pohanhuangtw: Good catch, let me update the comment here

PR #1709, opened 2024-12-16
NVSHAS-9696: Inconsistent colour indication of assets on risk page

jeffhuang4704:
The asset policy mode data could become stale when accessing the risk page.
This change ensures the latest data is fetched from the cache while composing the risk page.


jayhuang-suse (approved):
LGTM

PR #1711, opened 2024-12-17
NVSHAS-9649: Container link produces 404 response code in security event

jeffhuang4704:
Enhance /v1/workload endpoint to lookup both wlCacheMap and nvwlCacheMap caches when fetching workload information.


williamlin-suse (approved):
LGTM

PR #1722, opened 2024-12-19
NVSHAS-9584: Update tar.go with using upstream package tarutil

kyledong-suse:
1. Remove unmaintained functions in tar.go
2. Import upstream package tarutil in tar.go
3. Add upstream package tarutil in vendor

### DIFF
--- a/share/scan/scan_utils.go
+++ b/share/scan/scan_utils.go
@@ -19,6 +19,7 @@ import (
 	rpmdb "github.com/neuvector/go-rpmdb/pkg"
 	log "github.com/sirupsen/logrus"
 
+	"github.com/coreos/clair/pkg/tarutil"
 	"github.com/neuvector/neuvector/share"
 	"github.com/neuvector/neuvector/share/system"
 	"github.com/neuvector/neuvector/share/utils"
@@ -111,8 +112,8 @@ func NewScanUtil(sys *system.SystemTools) *ScanUtil {
 	return s
 }
 
-func (s *ScanUtil) readRunningPackages(id string, pid int, prefix, kernel string, pidHost bool) ([]utils.TarFileInfo, bool) {
-	var files []utils.TarFileInfo
+func (s *ScanUtil) readRunningPackages(pid int, prefix, kernel string, pidHost bool) (tarutil.FilesMap, bool) {
+	files := make(tarutil.FilesMap)
 	var hasPackage bool
 	for itr := range OSPkgFiles.Iter() {
 		var data []byte
@@ -140,7 +141,7 @@ func (s *ScanUtil) readRunningPackages(id string, pid int, prefix, kernel string
 					continue
 				}
 				name := fmt.Sprintf("%s%s", DpkgStatusDir, file.Name())
-				files = append(files, utils.TarFileInfo{Name: name, Body: filedata})
+				files[name] = filedata
 			}
 			hasPackage = true
 			continue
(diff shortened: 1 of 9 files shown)


share/scan/scan_utils.go
```diff
@@ -623,3 +626,24 @@ func GetAwsFuncPackages(fileName string) ([]*share.ScanAppPackage, error) {
...
 }
+
+func MakeTar(files tarutil.FilesMap) (*bytes.Buffer, error) {
+	buf := new(bytes.Buffer)
+	tw := tar.NewWriter(buf)
+	defer tw.Close()
+	for name, body := range files {
+		hdr := &tar.Header{
+			Name:     name,
+			Mode:     0655,
```
holyspectral, line 637: The file mode here is unusual.  Does it have to be 0655?

share/scan/scan_utils.go
```diff
@@ -623,3 +626,24 @@ func GetAwsFuncPackages(fileName string) ([]*share.ScanAppPackage, error) {
...
+			Typeflag: '0',
```
holyspectral, line 638:
```suggestion
			Typeflag: tar.TypeReg,
```

share/scan/scan_utils.go
```diff
@@ -111,8 +113,8 @@ func NewScanUtil(sys *system.SystemTools) *ScanUtil {
 	return s
 }
 
-func (s *ScanUtil) readRunningPackages(id string, pid int, prefix, kernel string, pidHost bool) ([]utils.TarFileInfo, bool) {
-	var files []utils.TarFileInfo
+func (s *ScanUtil) readRunningPackages(pid int, prefix, kernel string, pidHost bool) (tarutil.FilesMap, bool) {
+	files := make(tarutil.FilesMap)
```
holyspectral, line 117: [tarutil.FilesMap](https://pkg.go.dev/github.com/quay/clair/v3/pkg/tarutil#FilesMap) is just `type FilesMap map[string][]byte`

Do we need `github.com/coreos/clair/pkg/tarutil` for it?

holyspectral (commented):
Generally looks good to me.  Some questions about its implementation detail.

share/scan/scan_utils.go
kyledong-suse: I agree that we don't need `github.com/coreos/clair/pkg/tarutil`. I'll create `type FilesMap map[string][]byte` for tar processing

share/scan/scan_utils.go
kyledong-suse: good suggestion, fixed.

share/scan/scan_utils.go
kyledong-suse: that's from the previous implementation, may want to keep it for now

share/scan/scan_utils.go
kyledong-suse: After discussed with Jay, I aware that scanner uses most of the tar functions. So it's better to keep this tar.go and update it with using the upstream tarutil.

holyspectral: There seems to be some confusion here.  Let's talk through this.

holyspectral (changes requested):
I have concern about tar.go because its copyright from third party is being removed.  Please ensure the change is done correctly before merging this PR.

kyledong-suse:
Need to separate this PR into 2 part,
- Remove unmaintained tar.go
- Add NV implemented tar related functions

PR #1726, opened 2024-12-20
NVSHAS-9700/NVSHAS-9699 Retool logic for parseDotNetPackage

Acmarr:
to fix an issue where module count and contents could be inconsistent. Add/update related unit tests.

removed unnecessary debug logging.


jayhuang-suse (approved):
LGTM

PR #1734, opened 2025-01-07
NVSHAS-9751: add not-listed family process rule alert

jayhuang-suse:
(1) During Monitor mode, set alerts on not listed family processes. 
(2) Add process violation messages on above incidents


holyspectral (approved):
Adding a unit-test to cover `IsAllowedShieldProcess(()` would be advisable, even with only a very simple happy case.  Other than that it looks good to me.

jayhuang-suse:
The "IsAllowedShieldProcess()" needs a lot of local filesystem supports and is not applicable for a test case.

PR #1736, opened 2025-01-08
NVSHAS-9247 / 9326 / 9441 - Only scan cached images in harbor proxy cache projects

angelkestin:
In the Harbor Registry, there is a feature that allows users to set up a "proxy cache project". Essentially, this is registry that can sit between a source registry (such as Dockerhub, or Quay) and a client. When a client pulls images from the proxy cache project, those images will be "cached" in Harbor's memory.

Users of this feature have come to expect that operations on proxy cache projects, such as scanning, ONLY take into consideration these cached images. Meaning if a user sets up a proxy cache project to forward all images from DockerHub, and a client pulls `debian:latest`, our users expect that our scanner only scans `debian:latest`.

However, this was not the case, because when our scanner asks Harbor to list its available images, it uses the OCI Registry endpoints (i.e. `v2/_catalog`). In the case of proxy cache projects, this will return _**every available debian image in dockerhub**_. To avoid this, using the Harbor API to query for images is necessary during scanning, since it allows for the differentiation between cached, and non-cached images.

This PR adds a registry driver implementation for Harbor Registries, with an override for the `GetAllImages` method. This method will reach out to the Harbor API to gather image information during scanning, and ONLY return images that exist within the Harbor registry. Previously, when `GetAllImages` returned `nil`, the methods `GetRepoList` and `GetTags` would be used to gather image information for Harbor registries in `Registry.getScanImages()`. Those methods, by default, use the OCI Registry `_catalog` endpoints, which is what gets us into trouble when dealing with Harbor proxy cache projects.

### DIFF
--- a/controller/scan/harbor.go
+++ b/controller/scan/harbor.go
@@ -9,13 +9,11 @@ import (
 	"strings"
 
 	"github.com/neuvector/neuvector/share"
-	"github.com/neuvector/neuvector/share/httpclient"
 	log "github.com/sirupsen/logrus"
 )
 
 type harbor struct {
 	base
-	client http.Client
 }
 
 type HarborApiProject struct {
@@ -30,24 +28,6 @@ func (proj *HarborApiProject) IsProxyCacheProject() bool {
 type HarborApiRepository struct {
 	FullName  string `json:"name"`
 	ProjectId int    `json:"project_id"`
-}
-
-func newHarbor(baseDriver base) *harbor {
-	client := &http.Client{
-		Timeout: gitTimeout,
-	}
-
-	t, err := httpclient.GetTransport(baseDriver.proxy)
-	if err != nil {
-		log.WithError(err).Warn("failed to get transport")
-	} else {
-		client.Transport = t
-	}
-
-	return &harbor{
-		base:   baseDriver,
-		client: *client,
-	}
 }
 
 // TODO: the splits in projectName() and repoName() are duplicated, perform only once
@@ -96,14 +76,17 @@ func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
 
 // TODO: deal with large registries, implement pagination/chunked responses?
 func (h *harbor) getAllRepositories() ([]HarborApiRepository, error) {
-	req, err := http.NewRequest("GET", fmt.Sprintf("%s/api/v2.0/repositories", h.base.regURL), nil)
+	reqUrl, err := url.JoinPath(h.base.regURL, "api/v2.0/repositories")
+	if err != nil {
+		return nil, fmt.Errorf("could not generate repository request url: %w", err)
+	}
+	req, err := http.NewRequest("GET", reqUrl, nil)
 	if err != nil {
 		return nil, fmt.Errorf("could not make request object: %w", err)
 	}
 	req.SetBasicAuth(h.base.username, h.base.password)
 	req.Header.Add("accept", "application/json")
-	client := http.DefaultClient
-	resp, err := client.Do(req)
+	resp, err := h.rc.Client.Do(req)
 	if err != nil {
 		return nil, fmt.Errorf("could not do request to get all repositories: %w", err)
 	}
@@ -145,15 +128,18 @@ func (h *harbor) getTagsForRepository(repository HarborApiRepository) ([]string,
 	// the harbor api requires the repo name (which tends to contain a slash) to be url encoded twice
 	// example: a/b -> a%2Fb -> a%252Fb
 	encodedRepoName := url.PathEscape(url.PathEscape(repository.repoName()))
-	artifactEndpoint := fmt.Sprintf("api/v2.0/projects/%s/repositories/%s/artifacts?with_tag=true", projectName, encodedRepoName)
-	req, err := http.NewRequest("GET", fmt.Sprintf("%s/%s", h.base.regURL, artifactEndpoint), nil)
+	reqUrl, err := url.JoinPath(h.base.regURL, "api/v2.0/projects", projectName, "repositories", encodedRepoName, "artifacts")
+	if err != nil {
+		return nil, fmt.Errorf("could not generate artifact request url: %w", err)
+	}
+	req, err := http.NewRequest("GET", reqUrl, nil)
 	if err != nil {
 		return nil, fmt.Errorf("could not make request object: %w", err)
 	}
+	req.URL.RawQuery = "with_tag=true" // required to include image tags in response
 	req.SetBasicAuth(h.base.username, h.base.password)
 	req.Header.Add("accept", "application/json")
-	client := http.DefaultClient
-	resp, err := client.Do(req)
+	resp, err := h.rc.Client.Do(req)
 	if err != nil {
 		return nil, fmt.Errorf("could not do request to get all artifacts for repo %s: %w", repository.FullName, err)
 	}
--- a/controller/scan/registry.go
+++ b/controller/scan/registry.go
@@ -695,7 +695,7 @@ func newRegistryDriver(cfg *share.CLUSRegistryConfig, public bool, tracer httptr
 	} else if cfg.Type == share.RegistryTypeIBMCloud {
 		return &ibmcloud{base: baseDriver}
 	} else if cfg.Type == share.RegistryTypeHarbor {
-		return newHarbor(baseDriver)
+		return &harbor{base: baseDriver}
 	} else {
 		return &baseDriver
 	}


controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+func (h *harbor) getAllRepositories() ([]HarborApiRepository, error) {
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/api/v2.0/repositories", h.base.regURL), nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not make request object: %s", err.Error())
+	}
+	basicEncoding := base64.StdEncoding
+	basicAuthToken := basicEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", h.base.username, h.base.password)))
+	req.Header.Add("authorization", fmt.Sprintf("Basic %s", basicAuthToken))
+	req.Header.Add("accept", "application/json")
+	client := http.DefaultClient
```
holyspectral, line 108: `http.DefaultClient` doesn't consider proxy setting.  Use `base.newRegClient()` instead?

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+		log.WithError(err).Warn("failed to get transport")
+	} else {
+		client.Transport = t
+	}
+
+	return &harbor{
+		base:   baseDriver,
+		client: *client,
+	}
+}
```
holyspectral, line 52: Could we use `base.newRegClient()` because it considers proxy?

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+
+// TODO: the splits in projectName() and repoName() are duplicated, perform only once
+func (harborRepo *HarborApiRepository) projectName() string {
+	splitRepoName := strings.Split(harborRepo.FullName, "/")
+	if len(splitRepoName) < 2 {
+		// a harbor repo should always have the project name
+		// prefixed which means this should never happen
+		// but handling for the sake of robustness
+		return ""
+	}
```
holyspectral, line 62: Can we add some error handling here instead of returning an empty string?

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+
+	return &harbor{
+		base:   baseDriver,
+		client: *client,
+	}
+}
+
+// TODO: the splits in projectName() and repoName() are duplicated, perform only once
+func (harborRepo *HarborApiRepository) projectName() string {
```
holyspectral, line 55: It would be nice to have a unit test for this function, so folks know what is a valid project name and what is not without having to install harbor.

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	if len(splitRepoName) < 2 {
+		return harborRepo.FullName
+	}
+	return strings.Join(splitRepoName[1:], "/")
+}
+
+func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
+	repositories, err := h.getAllRepositories()
+	if err != nil {
+		return nil, fmt.Errorf("could not get repositories from harbor api: %s", err.Error())
```
holyspectral, line 80: It would be better to wrap the error instead of converting it into a string.  Some material can be found here: https://rollbar.com/blog/golang-wrap-and-unwrap-error/
```suggestion
		return nil, fmt.Errorf("could not get repositories from harbor api: %w", err)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	images := map[share.CLUSImage][]string{}
+	for _, repository := range repositories {
+		image := share.CLUSImage{
+			Repo: repository.FullName,
+		}
+		tags, err := h.getTagsForRepository(repository)
+		if err != nil {
+			return nil, fmt.Errorf("could not get tags for repository %s from harbor api: %s", repository.FullName, err.Error())
```
holyspectral, line 90:
```suggestion
			return nil, fmt.Errorf("could not get tags for repository %s from harbor api: %w", repository.FullName, err)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		images[image] = tags
+	}
+
+	return images, nil
+}
+
+// TODO: deal with large registries, implement pagination/chunked responses?
+func (h *harbor) getAllRepositories() ([]HarborApiRepository, error) {
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/api/v2.0/repositories", h.base.regURL), nil)
```
holyspectral, line 100: Consider using [url.JoinPath](https://pkg.go.dev/net/url#JoinPath)?  It's generally a safer choice because it cleans the URL and removes elements that can be used for path traversal.

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not make request object: %s", err.Error())
+	}
+	basicEncoding := base64.StdEncoding
+	basicAuthToken := basicEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", h.base.username, h.base.password)))
+	req.Header.Add("authorization", fmt.Sprintf("Basic %s", basicAuthToken))
```
holyspectral, line 106:
```suggestion
	req.SetBasicAuth(h.base.username, h.base.password)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	return images, nil
+}
+
+// TODO: deal with large registries, implement pagination/chunked responses?
+func (h *harbor) getAllRepositories() ([]HarborApiRepository, error) {
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/api/v2.0/repositories", h.base.regURL), nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not make request object: %s", err.Error())
```
holyspectral, line 102:
```suggestion
		return nil, fmt.Errorf("could not make request object: %w", err)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	basicEncoding := base64.StdEncoding
+	basicAuthToken := basicEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", h.base.username, h.base.password)))
+	req.Header.Add("authorization", fmt.Sprintf("Basic %s", basicAuthToken))
+	req.Header.Add("accept", "application/json")
+	client := http.DefaultClient
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("could not do request to get all repositories: %s", err.Error())
```
holyspectral, line 111:
```suggestion
		return nil, fmt.Errorf("could not do request to get all repositories: %w", err)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	}
```
holyspectral, line 112:
```suggestion
	}
	defer resp.Body.Close()
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+
+	basicEncoding := base64.StdEncoding
+	basicAuthToken := basicEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", h.base.username, h.base.password)))
+	req.Header.Add("authorization", fmt.Sprintf("Basic %s", basicAuthToken))
+	req.Header.Add("accept", "application/json")
+	client := http.DefaultClient
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("could not do request to get all artifacts for repo %s: %s", repository.FullName, err.Error())
+	}
```
holyspectral, line 161:
```suggestion
	}
	defer resp.Body.Close()
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	harborRepositories := []HarborApiRepository{}
+	// TODO: more efficiently deal with large responses, instead of reading all into a byte slice
+	respBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("could not read response body: %s", err.Error())
+	}
+	err = json.Unmarshal(respBytes, &harborRepositories)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshall response body json: %s", err.Error())
```
holyspectral, line 124:
```suggestion
		return nil, fmt.Errorf("could not unmarshall response body json: %w", err)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, fmt.Errorf("could not do request to get all repositories: %s", err.Error())
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("received error code from harbor api: %d", resp.StatusCode)
+	}
+	harborRepositories := []HarborApiRepository{}
+	// TODO: more efficiently deal with large responses, instead of reading all into a byte slice
+	respBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("could not read response body: %s", err.Error())
```
holyspectral, line 120:
```suggestion
		return nil, fmt.Errorf("could not read response body: %w", err)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	encodedRepoName := url.PathEscape(url.PathEscape(repository.repoName()))
+	artifactEndpoint := fmt.Sprintf("api/v2.0/projects/%s/repositories/%s/artifacts?with_tag=true", repository.projectName(), encodedRepoName)
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/%s", h.base.regURL, artifactEndpoint), nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not make request object: %s", err.Error())
+	}
+
+	basicEncoding := base64.StdEncoding
+	basicAuthToken := basicEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", h.base.username, h.base.password)))
+	req.Header.Add("authorization", fmt.Sprintf("Basic %s", basicAuthToken))
```
holyspectral, line 155:
```suggestion
	req.SetBasicAuth(h.base.username, h.base.password)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, fmt.Errorf("cannot parse project name for project/repository: %s", repository.FullName)
+	}
+
+	// the harbor api requires the repo name (which tends to contain a slash) to be url encoded twice
+	// example: a/b -> a%2Fb -> a%252Fb
+	encodedRepoName := url.PathEscape(url.PathEscape(repository.repoName()))
+	artifactEndpoint := fmt.Sprintf("api/v2.0/projects/%s/repositories/%s/artifacts?with_tag=true", repository.projectName(), encodedRepoName)
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/%s", h.base.regURL, artifactEndpoint), nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not make request object: %s", err.Error())
```
holyspectral, line 150:
```suggestion
		return nil, fmt.Errorf("could not make request object: %w", err)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+type HarborApiArtifact struct {
+	Tags []HarborApiTag `json:"tags"`
+}
+
+type HarborApiTag struct {
+	Name string `json:"name"`
+}
+
+func (h *harbor) getTagsForRepository(repository HarborApiRepository) ([]string, error) {
+	fmt.Printf("getting tags for repository: %s\n", repository.FullName)
```
holyspectral, line 138:
```suggestion
	log.Infof("getting tags for repository: %s", repository.FullName)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	// example: a/b -> a%2Fb -> a%252Fb
+	encodedRepoName := url.PathEscape(url.PathEscape(repository.repoName()))
+	artifactEndpoint := fmt.Sprintf("api/v2.0/projects/%s/repositories/%s/artifacts?with_tag=true", repository.projectName(), encodedRepoName)
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/%s", h.base.regURL, artifactEndpoint), nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not make request object: %s", err.Error())
+	}
+
+	basicEncoding := base64.StdEncoding
+	basicAuthToken := basicEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", h.base.username, h.base.password)))
```
holyspectral, line 154: Consider using [url.JoinPath](https://pkg.go.dev/net/url#JoinPath)?

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	req.Header.Add("authorization", fmt.Sprintf("Basic %s", basicAuthToken))
+	req.Header.Add("accept", "application/json")
+	client := http.DefaultClient
```
holyspectral, line 157: Use base.newRegClient() instead?

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("received error code from harbor api: %d", resp.StatusCode)
+	}
+
+	harborArtifacts := []HarborApiArtifact{}
+	// TODO: more efficiently deal with large responses, instead of reading all into a byte slice
+	respBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("could not read response body: %s", err.Error())
```
holyspectral, line 170:
```suggestion
		return nil, fmt.Errorf("could not read response body: %w", err)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	fmt.Println(string(respBytes))
+	err = json.Unmarshal(respBytes, &harborArtifacts)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshall response body json: %s", err.Error())
```
holyspectral, line 175:
```suggestion
		return nil, fmt.Errorf("could not unmarshall response body json: %w", err)
```

controller/scan/harbor.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, fmt.Errorf("received error code from harbor api: %d", resp.StatusCode)
+	}
+
+	harborArtifacts := []HarborApiArtifact{}
+	// TODO: more efficiently deal with large responses, instead of reading all into a byte slice
+	respBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("could not read response body: %s", err.Error())
+	}
+	fmt.Println(string(respBytes))
```
holyspectral, line 172:
```suggestion

```

holyspectral (changes requested):
Generally looks good to me.  Thanks for making NV support this!

There are two issues that have to be fixed before merging this:
- Close resp.Body.
- Use `base.newRegClient()` so proxy settings and global TLS verification settings can be honored.

controller/scan/harbor_test.go
```diff
@@ -0,0 +1,41 @@
...
+		got, err := repo.projectName()
+		if err != nil {
+			if testCase.shouldError {
+				continue
+			}
+			t.Errorf("received error for valid name \"%s\": %s", testCase.fullName, err.Error())
+		}
+		if got != testCase.expectedRepoName {
+			t.Errorf("unexpected repo name for %s: got %s, expected %s", testCase.fullName, got, testCase.expectedRepoName)
+		}
```
holyspectral, line 39: Not a blocker for this PR, but these checks can be simplified into one line. :-)
```suggestion
                assert.Equal(t, testCase.expectedRepoName, got)
```

holyspectral (approved):
LGTM.  Thanks for working on this!

jipduo:
Interesting! We might need something similar for a Sonatype Nexus Proxy Repository for Docker. https://github.com/neuvector/neuvector/issues/1978

PR #1743, opened 2025-01-10
NVSHAS-9756: disable file monitor

jayhuang-suse:
Add disable file monitor env flag

The env ENF_NO_FILE_PROFILE will disable file monitor.


holyspectral (approved):
LGTM

PR #1745, opened 2025-01-14
NVSHAS-9248: record network policy with match counters and last matched timestamp

gfsuse:


controller/api/apis.go
```diff
@@ -1314,6 +1314,8 @@ type RESTPolicyRule struct {
 	LastModTS    int64    `json:"last_modified_timestamp"`
 	CfgType      string   `json:"cfg_type"` // CfgTypeLearned / CfgTypeUserCreated / CfgTypeGround / CfgTypeFederal (see above)
 	Priority     uint32   `json:"priority"`
+	MatchCntr    uint64   `json:"match_counter"`
```
williamlin-suse, line 1317: api/apis.yaml needs to be updated about these 2 new fields

controller/cache/connect.go
```diff
@@ -720,6 +720,42 @@ func CalculateGroupMetric(conn *share.CLUSConnection) {
...
+		if exist, ok := policyCache.ruleMap[polid]; ok {
+			exist.MatchCntr = polmet.MatchCntr
+			exist.LastMatchAt = time.Unix(int64(polmet.LastMatchAt), 0).UTC()
+		} else {
+			delete(policyMetricMap, polid)
+		}
+	}
+}
+
+func calNetPolicyMet(conn *share.CLUSConnection) {
```
williamlin-suse, line 736: Is it ok that policyMetricMap access is protected by cacheMutexLock in UpdatePolicyMetric() but not in calNetPolicyMet() ?

controller/cache/cache.go
```diff
@@ -1687,6 +1689,10 @@ func startWorkerThread(ctx *Context) {
 				if isLeader() {
 					CheckGroupMetric()
 				}
+			case <-policyMetricCheckTicker.C:
+				if isLeader() {
```
williamlin-suse, line 1693: Why is UpdatePolicyMetric() called only on lead?

controller/cache/connect.go
```diff
@@ -720,6 +720,42 @@ func CalculateGroupMetric(conn *share.CLUSConnection) {
...
+	if conn.PolicyId > 0 && (conn.PolicyAction == C.DP_POLICY_ACTION_ALLOW ||
+		conn.PolicyAction == C.DP_POLICY_ACTION_VIOLATE || conn.PolicyAction == C.DP_POLICY_ACTION_DENY) {
+		polid := conn.PolicyId
+		if policyMetricMap == nil {
+			policyMetricMap = make(map[uint32]*share.CLUSNetPolicyMetric)
+		}
+		if polMet, ok := policyMetricMap[polid]; ok {
+			polMet.MatchCntr = polMet.MatchCntr + 1
```
williamlin-suse, line 744: Is cache.UpdateConnections() called for new connections or all connections(i.e. including existing connections)?
If it's for all connections, isn't a connection counted once every time cache.UpdateConnections() is called?
If it's for new connections only, because UpdatePolicyMetric() is called on lead only, how/when are the obsolete entries in policyMetricMap deleted?

controller/api/apis.go
gfsuse: ok

controller/cache/connect.go
gfsuse: I will update

controller/cache/cache.go
gfsuse: same connection report are used by all controllers, but only leader learn policy and push it to consul, so let leader count policy matching number is good enough.

controller/cache/connect.go
gfsuse: connection report is sent by dp when packet are processed which include policy matching, connection report can be new connection or existing connection, in each connection report there is matching policy id, when connection is sent to controller we can record how many time a policy is hit, when policy is deleted the obsolete entry can be deleted.

controller/cache/connect.go
```diff
@@ -720,6 +720,45 @@ func CalculateGroupMetric(conn *share.CLUSConnection) {
 	}
 }
 
+func UpdatePolicyMetric() {
+	cacheMutexLock()
+	defer cacheMutexUnlock()
+	for polid, polmet := range policyMetricMap {
```
kyledong-suse, line 726: Maybe change polid => polId, polmet => polMet to follow Camel Case to make it consistent as line 746

controller/cache/connect.go
```diff
@@ -720,6 +720,45 @@ func CalculateGroupMetric(conn *share.CLUSConnection) {
...
+	}
+}
+
+func calNetPolicyMet(conn *share.CLUSConnection) {
+	if conn.PolicyId > 0 && (conn.PolicyAction == C.DP_POLICY_ACTION_ALLOW ||
+		conn.PolicyAction == C.DP_POLICY_ACTION_VIOLATE || conn.PolicyAction == C.DP_POLICY_ACTION_DENY) {
+		cacheMutexLock()
+		defer cacheMutexUnlock()
+
+		polid := conn.PolicyId
```
kyledong-suse, line 742: same as above comment, maybe change polid => polId?

controller/cache/connect.go
gfsuse: ok

controller/cache/connect.go
gfsuse: ok

BinX-Suse (approved):

PR #1753, opened 2025-01-21
NVSHAS-9753: Prevent Rancher user disable Authentication of OpenShift…

williamlin-suse:
… or Rancher's RBAC


jeffhuang4704 (approved):
looks good to me.

PR #1756, opened 2025-01-21
NVSHAS-9777: Webhook JSON with duplicated 'level' keys

williamlin-suse:
Root cause: 'level' field already exists in the embedded type LogCommon


jeffhuang4704 (approved):
looks good to me.

PR #1759, opened 2025-01-22
NVSHAS-9325: NV Protect harden improvement

jayhuang-suse:
1. Apply zero-drift into NV pods
2. Apply fine file permission controls
3. Remove NV Protect process rules
4. The enforcement will become "deny" instead of "kill" operations


agent/probe/faccess_linux.go
```diff
@@ -772,6 +796,17 @@ func (fa *FileAccessCtrl) processEvent(ev *fanotify.EventMetadata) (bool, string
 				}
 			}
 
+			if nvRole != "" && res < rule_allowed {
+				if fa.prober != nil && fa.prober.disableNvProtect {
+					res = rule_allowed
+				}
+
+				if rule_uuid == share.CLUSReservedUuidNotAlllowed {
```
holyspectral, line 804: minor: Can you also help to fix the typo in `CLUSReservedUuidNotAlllowed`?

agent/probe/process.go
```diff
@@ -2964,6 +2908,79 @@ func (p *Probe) PatchContainerProcess(pid int, bEval bool) bool {
...
+		if path, err = global.SYS.GetFilePath(pid); err != nil {
+			log.WithFields(log.Fields{"err": err, "nvRole": nvRole, "pid": pid}).Debug()
+		}
+	}
+
+	log.WithFields(log.Fields{"path": path, "ppath": ppath, "nvRole": nvRole, "ppid": ppid}).Debug()
+	if global.RT.IsRuntimeProcess(filepath.Base(ppath), nil) {
+		switch path {
+		case "/usr/bin/cat": // k8s readiness probe
+			return p.bKubePlatform && isControllerType(nvRole)
```
holyspectral, line 2942: `cat` can be used for many usage.  Is it possible to move readiness probe to enforcer's gRPC port for example?

agent/probe/process.go
```diff
@@ -2964,6 +2908,79 @@ func (p *Probe) PatchContainerProcess(pid int, bEval bool) bool {
...
+		case "/usr/bin/bash": // cli checks
+			return isManagerType(nvRole)
+		}
+	} else if isManagerType(nvRole) && ppath == "/usr/bin/bash" && path == "/usr/bin/python3.12" {
+		return true /// only for cli case
+	}
```
holyspectral, line 2948: Can we move these white list into a yaml config file and include it in enforcer's container image?  This way we can override it using docker or k8s volume mount without having to rebuild enforcer.

holyspectral (commented):
Looks good in high level.  I like that we apply zero draft policy to NV containers!
Two questions:
- Could we move to a different readinessProbe for controller, so we can avoid `cat` to be run?
- Could we make the white list configurable via a yaml file in enforcer image?

agent/probe/process.go
jayhuang-suse: We only use "cat" on the controller, not other nv pods, because it represents the readiness of the consul server.

agent/probe/process.go
jayhuang-suse: It could be another project if requested.

holyspectral (changes requested):
Looks like the change is causing `cat` from controller's readinessProbe being reported as blocked, while they're not. Let's figure this out before merging this.

jayhuang-suse:
the k8s probe issue is fixed.

agent/probe/process.go
```diff
@@ -2964,6 +2908,95 @@ func (p *Probe) PatchContainerProcess(pid int, bEval bool) bool {
...
+		if global.RT.IsRuntimeProcess(filepath.Base(ppath), nil) || bRuncInit {
+			mLog.WithFields(log.Fields{"path": path, "ppath": ppath, "nvRole": nvRole, "k8s": p.bKubePlatform, "bRuncInit": bRuncInit}).Debug()
+			switch path {
+			case "/usr/bin/cat": // k8s readiness probe
+				if p.bKubePlatform && isControllerType(nvRole) {
+					if bRuncInit {
+						return true
+					} else if len(proc.cmds) > 1 && proc.cmds[1] == "/tmp/ready" {
+						return true
+					}
```
holyspectral, line 2959: Can you elaborate a little more about this newly added code?  It looks like `cat /tmp/ready /etc/passwd` would bypass the check, so the `else if` check would be a little useless.  Can we double confirm that?

agent/probe/process.go
```diff
@@ -2964,6 +2908,95 @@ func (p *Probe) PatchContainerProcess(pid int, bEval bool) bool {
...
+			case "/usr/bin/bash": // cli checks
+				return isManagerType(nvRole)
+			}
+			break
+		}
+
+		if filepath.Dir(ppath) == "/usr/local/bin" {
+			log.WithFields(log.Fields{"caller": ppath}).Debug()
+			return true
+		}
```
holyspectral, line 2971: This would allow pip3 to run in manager pod, is it as expected? 
```
$ kubectl exec -it neuvector-manager-pod-5b9ccbd8c7-jgcxh -- bash
bash-4.4$ pip3 install helloworld 
WARNING: The directory '/nonexistent/.cache/pip' or its parent directory is not owned or is 
...
```
In 5.4.2, pip3 will get killed. 
```
$ kubectl exec -it neuvector-manager-pod-5b9ccbd8c7-jgcxh -- bash
bash-4.4$ pip3
Killed
bash-4.4$ 
```

agent/probe/process.go
```diff
@@ -2964,6 +2908,95 @@ func (p *Probe) PatchContainerProcess(pid int, bEval bool) bool {
...
+func isControllerType(role string) bool {
+	return role == "controller" || role == "controller+enforcer+manager" || role == "allinone"
+}
+
+func isManagerType(role string) bool {
+	return role == "manager" || role == "controller+enforcer+manager" || role == "allinone"
+}
+
+// protected by lock
+func (p *Probe) isNVChildProcess(nvRole string, proc *procInternal) bool {
```
holyspectral, line 2920: It doesn't have to be in this PR, but please consider refactoring this function so it can be tested with unit tests.  This will make sure that it has no unexpected behavior as well as provides a good baseline for the future discussion, i.e., what the correct behavior is.

holyspectral (commented):
Sorry for the wait.  Here is the second round.

Overall, like I mentioned last time it looks great, but there are some behavior inconsistency when compare with 5.4.2.

I'm sure QA will find them eventually.  So do you consider fixing them now? 
If you prefer merging this PR first, let's get manager CLI fixed first, as I think it's probably just a small change.

Thanks for fixing this!

agent/probe/process.go
```diff
@@ -2964,6 +2908,95 @@ func (p *Probe) PatchContainerProcess(pid int, bEval bool) bool {
...
+	if path == "" && pid > 0 {
+		if path, err = global.SYS.GetFilePath(pid); err != nil {
+			log.WithFields(log.Fields{"err": err, "nvRole": nvRole, "pid": pid}).Debug()
+		}
+	}
+
+	mLog.WithFields(log.Fields{"path": path, "ppath": ppath, "nvRole": nvRole, "ppid": ppid}).Debug()
+	if isManagerType(nvRole) && ppath == "/usr/bin/bash" && path == "/usr/bin/python3.12" {
+		return true /// only for cli case
+	}
```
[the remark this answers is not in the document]
jayhuang-suse: It might be the time to change the cli script since the default shell of our pods are "bash" now.

agent/probe/process.go
jayhuang-suse: It should raise an alert. Basically, it will only allow the health check command. Others will be raised an alert.

agent/probe/process.go
jayhuang-suse: It is required for the enforcement because of some of the fine-tuning operations.

agent/probe/process.go
holyspectral: That's exactly my point.  `cat /etc/passwd` would trigger an alert, but `cat /tmp/ready /etc/passwd` won't.  :-)
Would this work slightly better? 
```suggestion
					if bRuncInit {
						return true
					} else if len(proc.cmds) == 2 && proc.cmds[1] == "/tmp/ready" {
						return true
					}
```

jayhuang-suse: Fixed.

agent/probe/process.go
jayhuang-suse: Fixed.

holyspectral (approved):
LGTM thanks!

PR #1761, opened 2025-01-24
Nvshas 9326 - Handle paginated repository lists from harbor api

alopez-suse:
Previously, only a single request would be made to the Harbor "get all repositories" endpoint. By default, without any provide query parameters, the responses are paginated to 10 repositories maximum. In order to retrieve every repository in Harbor, the request will need to be repeated until the retrieved number of repositories matches the `X-Total-Count` header in the response.

This implementation also handles the edge case where the number of repositories may have changed during this operation.

It _is possible_ that during our requests to get all repositories, the number of repositories stays the same, BUT the repositories themselves change. (For example, deleting and pushing an image at the exact same time while we are performing these requests). In this instance this implementation could return an out of date representation of the registry. However, I did not see any way to confirm whether this has happened during harbor (such as a "transaction id" value anywhere). Thus, in the rare case that this happens, it would need to be corrected in a subsequent scan (assuming it doesn't happen _again_).

### DIFF
--- a/controller/scan/harbor.go
+++ b/controller/scan/harbor.go
@@ -15,6 +15,7 @@ import (
 
 const totalCounterHeaderCanonicalForm string = "X-Total-Count"
 const defaultRepositoryPageSize int = 10
+const maxGetRepoRequeries int = 3
 
 type harbor struct {
 	base
@@ -97,7 +98,7 @@ func (h *harbor) getAllRepositories() ([]HarborApiRepository, error) {
 		if totalReposInRegistry == -1 {
 			totalReposInRegistry = respTotalCount
 		} else if totalReposInRegistry != respTotalCount {
-			if timesTotalCountChanged < 3 {
+			if timesTotalCountChanged < maxGetRepoRequeries {
 				// number of repos changed while we were querying registry
 				// rerun query for all previous pages as well
 				pageWhereTotalCountChanged = pageNum


controller/scan/harbor.go
```diff
@@ -74,37 +78,93 @@ func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
...
+			// number of repos changed while we were querying registry
+			// rerun query for all previous pages
+			pageWhereTotalCountChanged = pageNum
+			pageNum = 0
+			repositoryPages = map[int][]HarborApiRepository{}
+			totalReposInRegistry = totalCount
+			totalFetchedRepositories = 0
+		}
+
+		repositoryPages[pageNum] = repositories
```
pohanhuangtw, line 108: Does the pageNum key always exist in the repositoryPages object?

controller/scan/harbor.go
```diff
@@ -74,37 +78,93 @@ func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
...
+		totalFetchedRepositories += len(repositories)
+		if totalFetchedRepositories >= totalReposInRegistry {
```
pohanhuangtw, line 110: Why do we check if totalFetchedRepositories >= totalReposInRegistry before adding all repositories? Why not simply add them directly? Is there a specific reason for this condition?

controller/scan/harbor.go
```diff
@@ -74,37 +78,87 @@ func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
...
+			// we've already appended the repos for this page in a previous iteration
+			continue
+		}
+		repositories, totalCount, err := h.getPageOfRepositories(pageNum)
+		if err != nil {
+			return nil, fmt.Errorf("could not get page %d of harbor repositories: %w", pageNum, err)
+		}
+		if totalReposInRegistry == -1 {
+			totalReposInRegistry = totalCount
+		} else if totalReposInRegistry != totalCount {
```
williamlin-suse, line 97: Let's say totalReposInRegistry=50 from pageNum 1~4
In the round that pageNum is 5, totalCount returned from h.getPageOfRepositories(5) is 52.
So line 97 is totalReposInRegistry(50) != totalCount(52) & 
1. pageWhereTotalCountChanged is set to 5
2. pageNum is set to 0
3. fetchedRepositories is reset to empty slice
4. totalReposInRegistry is set to be 52
5. repositories for pageNum 5 are appended to fetchedRepositories
6. pageNum is set to 1
Then the for loop starts over from pageNum=1 again. (repositories for pageNum 1~4 are appended)
When pageNum reaches 5, line 87 is
```
               if pageNum(5) == pageWhereTotalCountChanged(5) {
			// we've already appended the repos for this page in a previous iteration
			continue  // go back to for loop with the same pageNum(5) value & stuck in an infinite for loop ?
		}

```

controller/scan/harbor.go
```diff
@@ -74,37 +78,90 @@ func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
...
+	pageWhereTotalCountChanged := -1
+	fetchedRepositories := []HarborApiRepository{}
+	totalReposInRegistry := -1
+	for {
+		if pageNum == pageWhereTotalCountChanged {
+			// we've already appended the repos for this page in a previous iteration
+			pageNum++
+			continue
+		}
+		repositories, totalCount, err := h.getPageOfRepositories(pageNum)
```
williamlin-suse, line 92: what's returned if a pageNum value bigger than the last page is used to call h.getPageOfRepositories?
Does it return
1. []HarborApiRepository{}, totalCount, {our of page boundary error} , or
2. []HarborApiRepository{}l, totalCount, nil
?

controller/scan/harbor.go
```diff
@@ -74,37 +78,90 @@ func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
...
+			pageNum++
+			continue
+		}
+		repositoriesForPage, totalCount, err := h.getPageOfRepositories(pageNum)
+		if err != nil {
+			return nil, fmt.Errorf("could not get page %d of harbor repositories: %w", pageNum, err)
+		}
+		if totalReposInRegistry == -1 {
+			totalReposInRegistry = totalCount
+		} else if totalReposInRegistry != totalCount {
```
pohanhuangtw, line 98: Can this potentially lead to indefine loop?

controller/scan/harbor.go
```diff
@@ -74,37 +78,90 @@ func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
...
+	pageWhereTotalCountChanged := -1
+	allFetchedRepositories := []HarborApiRepository{}
+	totalReposInRegistry := -1
+	for {
+		if pageNum == pageWhereTotalCountChanged {
+			// we've already appended the repos for this page in a previous iteration
+			pageNum++
+			continue
+		}
+		repositoriesForPage, totalCount, err := h.getPageOfRepositories(pageNum)
```
pohanhuangtw, line 92: Should we add som re-try here? the harbor api could fait at the moment, but works after couple of retry.

controller/scan/harbor.go
```diff
@@ -74,37 +78,90 @@ func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
...
+		return nil, 0, fmt.Errorf("could not make request object: %w", err)
 	}
 	req.SetBasicAuth(h.base.username, h.base.password)
 	req.Header.Add("accept", "application/json")
 	resp, err := h.rc.Client.Do(req)
 	if err != nil {
-		return nil, fmt.Errorf("could not do request to get all repositories: %w", err)
+		return nil, 0, fmt.Errorf("could not do request to get all repositories: %w", err)
 	}
 	defer resp.Body.Close()
```
pohanhuangtw, line 142: Hi @alopez-suse ,
I noticed that if the error occurs on line 140, resp.Body.Close() will not be executed, which could lead to a resource leak. Could you move this line to 138 to ensure proper cleanup?

Also, regarding the retry mechanism—was it already included in this PR?

Thanks!

controller/scan/harbor.go
pohanhuangtw: I think we should allow an incomplete or inaccurate list of images and log the issues accordingly. At least this way, the user can still utilize the available data.

Also, adding a maximum rescan limit is a great idea—please include it.

What do you think? @williamlin-suse

controller/scan/harbor.go
pohanhuangtw: Has the retry mechanism been added in the current changes?

controller/scan/harbor.go
```diff
@@ -135,7 +144,15 @@ func (h *harbor) getPageOfRepositories(pageNum int) ([]HarborApiRepository, int,
...
+
+	var resp *http.Response
+	for tries := 1; tries <= 3; tries++ {
+		resp, err = h.rc.Client.Do(req)
+		if err == nil {
+			break
+		}
+		time.Sleep(time.Second * time.Duration(tries*3)) // backoff
+	}
 	if err != nil {
```
pohanhuangtw, line 156: Please move `defer resp.Body.Close() `here to avoid leaking due to early return.

controller/scan/harbor.go
```diff
@@ -74,37 +79,106 @@ func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
 	return images, nil
 }
 
-// TODO: deal with large registries, implement pagination/chunked responses?
 func (h *harbor) getAllRepositories() ([]HarborApiRepository, error) {
-	reqUrl, err := url.JoinPath(h.base.regURL, "api/v2.0/repositories")
+	pageNum := 1
+	pageWhereTotalCountChanged := -1
```
pohanhuangtw, line 84: minor: pageWhereTotalCountChanged -> lastChangedPage will be more readable

controller/scan/harbor.go
```diff
@@ -74,37 +79,106 @@ func (h *harbor) GetAllImages() (map[share.CLUSImage][]string, error) {
...
+			continue
+		}
+		repositoriesForPage, respTotalCount, err := h.getPageOfRepositories(pageNum)
+		if err != nil {
+			return nil, fmt.Errorf("could not get page %d of harbor repositories: %w", pageNum, err)
+		}
+		if totalReposInRegistry == -1 {
+			totalReposInRegistry = respTotalCount
+		} else if totalReposInRegistry != respTotalCount {
+			if timesTotalCountChanged < 3 {
```
pohanhuangtw, line 101: minor: 3, can be defined as a variable above

controller/scan/harbor.go
```diff
@@ -139,10 +213,18 @@ func (h *harbor) getTagsForRepository(repository HarborApiRepository) ([]string,
 	req.URL.RawQuery = "with_tag=true" // required to include image tags in response
 	req.SetBasicAuth(h.base.username, h.base.password)
 	req.Header.Add("accept", "application/json")
-	resp, err := h.rc.Client.Do(req)
+	var resp *http.Response
+	for tries := 1; tries <= 3; tries++ {
```
pohanhuangtw, line 217: Minor: The value 3 can be defined as a variable for better readability and maintainability.
Since lines 217-223 and 149-155 contain identical logic, consider extracting them into a reusable function to improve maintainability and reduce duplication.

pohanhuangtw: like

```go
func (h *harbor) doRequestWithRetry(req *http.Request, maxRetries int) (*http.Response, error) {
    var resp *http.Response
    var err error
    for tries := 1; tries <= maxRetries; tries++ {
        resp, err = h.rc.Client.Do(req)
        if err == nil {
            return resp, nil
        }
        time.Sleep(time.Second * time.Duration(tries*3))
    }
    return nil, fmt.Errorf("request failed after %d retries: %w", maxRetries, err)
}

```

pohanhuangtw:
Overall, the code looks good to me! Just a couple of areas where improvements could be made.

pohanhuangtw (changes requested):
Overall, the code looks good to me! Just a couple of areas where improvements could be made.

pohanhuangtw (approved):
Overall the code looks good to me, just couple of small improvement can be done.
consider using a named constant to improve clarity and maintainability: 
```go
h.rc.Client.DoWithRetry(req, maxGetRepoRequeries)
```

PR #1762, opened 2025-01-24
[NVSHAS-9784] NV returning 404 during Jfrog image repository scanning

pohanhuangtw:
### Summary
- Append header to support the OCI format of a manifest. 

### Root cause
The current implementation will call the following for any jfrog build when scanning

```bash
curl -u usr:pwd\
     -H "Accept: application/vnd.docker.distribution.manifest.v2+json" \
     https://nvdebug.jfrog.io/v2/{repo}/{image}/manifests/{tag}
 ```

But the header is always, this header is compatible with Docker V2 image manifests.

```bash
application/vnd.docker.distribution.manifest.v2+json 
```

However, to support OCI-compliant images (e.g., OCI build images), the header must also accept:
```bash
application/vnd.oci.image.manifest.v1+json 
```

Thus we use consigned-image as fallback function to work with it

### DIFF
--- a/share/scan/registry/manifest.go
+++ b/share/scan/registry/manifest.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


BinX-Suse (approved):
LGTM.

gfsuse (approved):
LGTM

pohanhuangtw:
@jaimeyu-suse @holyspectral I append the header to fetch the oci format as well, please help review

jayhuang-suse (approved):
LGTM

PR #1764, opened 2025-01-28
NVSHAS-9756: File Monitor Performance Optimization

jayhuang-suse:
This change aims to improve the performance of the file monitor by reducing its CPU and I/O overhead

(1) Initial Flat Scan: Will perform a one-time "flat scan" to identify all relevant files within the workload's scope. This eliminates the need for constant directory traversal, significantly reducing I/O operations.

(2) Limited the number of concurrent "worklets" used by the file monitor will be limited to two. This controlled concurrency prevents excessive resource consumption and potential performance bottlenecks.

(3) Optimized "worklet" Scope:  Worklet efficiency will be improved by defining specific folders for each worklet to monitor.

### DIFF
--- a/agent/probe/fsn.go
+++ b/agent/probe/fsn.go
@@ -295,7 +295,9 @@ func (fsn *FileNotificationCtr) updateFileInfo(index string, file, path string) 
 
 	// reporting events
 	// log.WithFields(log.Fields{"id": root.id, "file": file, "finfo": finfo}).Debug("FSN:")
-	go fsn.prober.ProcessFsnEvent(root.id, []string{file}, *finfo)
+	if bExec || bJavaPkg {
+		go fsn.prober.ProcessFsnEvent(root.id, []string{file}, *finfo)
+	}
 }
 
 // already locked
@@ -461,7 +463,7 @@ func (fsn *FileNotificationCtr) skipPathByRole(role, path string) bool {
 			return true
 		}
 		fallthrough
-	case "enforcer", "scanner":
+	case "enforcer", "scanner", "controller", "manager":
 		if strings.HasPrefix(path+"/", "/tmp/") {
 			return true
 		}
--- a/agent/workerlet/pathWalker/pathWalker.go
+++ b/agent/workerlet/pathWalker/pathWalker.go
@@ -179,7 +179,7 @@ func (tm *taskMain) WalkPathTask(req workerlet.WalkPathRequest) {
 		Files: make([]*workerlet.FileData, 0),
 	}
 
-	log.WithFields(log.Fields{"req": req}).Debug()
+	//log.WithFields(log.Fields{"req": req}).Debug()
 	rootPath := filepath.Join(fmt.Sprintf(procRootMountPoint, req.Pid), req.Path)
 	rootPathLen := len(rootPath)
 	rootPath += "/"
@@ -220,6 +220,18 @@ func (tm *taskMain) WalkPathTask(req workerlet.WalkPathRequest) {
 			if path != rootPath {
 				if utils.IsMountPoint(path) {
 					log.WithFields(log.Fields{"path": path}).Debug("skip dir")
+					return filepath.SkipDir
+				}
+
+				bSkip := true
+				ldir := path[rootPathLen:]
+				for _, rdir := range req.Dirs {
+					if strings.HasPrefix(ldir, rdir) {
+						bSkip = false
+						break
+					}
+				}
+				if bSkip {
 					return filepath.SkipDir
 				}
 			}
--- a/share/fsmon/monitor.go
+++ b/share/fsmon/monitor.go
@@ -1,6 +1,7 @@
 package fsmon
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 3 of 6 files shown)


jayhuang-suse:
The initial comparison is in its original JIRA case.

share/fsmon/monitor.go
```diff
@@ -588,9 +575,62 @@ func (w *FileWatch) addDir(bIncInotify bool, finfo *osutil.FileInfoExt, files ma
...
+
+	req := workerlet.WalkPathRequest{
+		Pid:     pid,
+		Path:    "/",
+		Timeout: rootIterTimeout,
+		Dirs:    dirs,
+	}
+
+	var err error
+	if err = w.walkerLimiter.Acquire(context.Background(), 1); err == nil {
```
holyspectral, line 618: According to [docs](https://pkg.go.dev/golang.org/x/sync/semaphore#Weighted.Acquire), if a walker takes a lot of time or gets stuck, this go routine could wait forever in this `semaphore.Acquire()` call.  

Please utilize a timeout or use cancel(), so in the worst case it can give up and show error message to us.  

```suggestion
        ctx, cancel := context.WithTimeout(context.Background, req.Timeout)
        defer cancel()
	if err = w.walkerLimiter.Acquire(ctx, 1); err == nil {
```

share/fsmon/monitor.go
```diff
@@ -990,187 +1012,74 @@ func (w *FileWatch) SetMonitorTrace(bEnable bool, logLevel string) {
...
+				continue
 			}
 		}
-		dirs.Remove(any)
 
-		if !recur { // only 1st layer of directory
-			break
+		fpath = global.SYS.ContainerFilePath(pid, f.File)
+		file := &osutil.FileInfoExt{
+			FileMode:  f.Info.Mode, // ??
```
holyspectral, line 1068: Is it possible to remove the old `// ??` comment here? It doesn't seem very useful.

holyspectral (changes requested):
Generally looks good to me.  The `semaphore.Acquire()` would be a blocker.  Let's fix it before merging this.

agent/workerlet/pathWalker/pathWalker.go
```diff
@@ -222,6 +222,18 @@ func (tm *taskMain) WalkPathTask(req workerlet.WalkPathRequest) {
...
+				ldir := path[rootPathLen:]
+				for _, rdir := range req.Dirs {
+					if strings.HasPrefix(ldir, rdir) {
+						bSkip = false
+						break
+					}
+				}
+				if bSkip {
+					return filepath.SkipDir
+				}
```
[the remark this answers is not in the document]
jayhuang-suse: It is required because the worker is a passive receiver, not a necessary bottleneck.

share/fsmon/monitor.go
jayhuang-suse: All tasks need to execute and can not be failed calls. The timeout is already applied on the worker.

agent/probe/fsn.go
```diff
@@ -295,7 +295,9 @@ func (fsn *FileNotificationCtr) updateFileInfo(index string, file, path string)
 
 	// reporting events
 	// log.WithFields(log.Fields{"id": root.id, "file": file, "finfo": finfo}).Debug("FSN:")
-	go fsn.prober.ProcessFsnEvent(root.id, []string{file}, *finfo)
+	if bExec || bJavaPkg {
```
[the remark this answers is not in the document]
jayhuang-suse: NV Protect takes care of those executables so it is okay.

share/fsmon/monitor.go
jayhuang-suse: It is good to keep it as future hint...

share/fsmon/monitor.go
holyspectral: If it's expected that there shouldn't be timeout when acquiring a semaphore, could you give it a long timeout, e.g., 1 hour, so in the rare case it happens, we can still see the error?

share/fsmon/monitor.go
holyspectral: yeah, or we should make it more meaningful...

agent/workerlet/pathWalker/pathWalker.go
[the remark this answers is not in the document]
holyspectral: yeah that's fine. Just wanted to point out that there are some algorithms that can be used to improve this part, IF this becomes a bottleneck.  It looks good enough for me right now.

share/fsmon/monitor.go
jayhuang-suse: As I mentioned, the timeouts were applied at their following workers. The semaphore timeout is not required there.

share/fsmon/monitor.go
```diff
@@ -588,9 +575,62 @@ func (w *FileWatch) addDir(bIncInotify bool, finfo *osutil.FileInfoExt, files ma
...
+func getBaseDirPrefix(filter share.CLUSFileMonitorFilter) string {
+	dir := strings.Replace(filter.Path, "\\.", ".", -1)
+	if index := strings.Index(dir, ".*"); index > 0 {
+		dir = dir[:(index - 1)]
+	}
+	if filter.Regex == "" {
+		dir = filepath.Dir(dir)
+	}
+	return dir
+}
```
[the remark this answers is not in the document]
jayhuang-suse: Added

share/fsmon/monitor.go
holyspectral: For sure.  Any concern if we change it like this?  We should be very careful about every lock mechanism because when it happens, it's very difficult to debug.  I don't think printing an error when it times out would be harmful.  Any different thoughts? 

```suggestion
        ctx, cancel := context.WithTimeout(context.Background, time.Hour * 1)
        defer cancel()
	if err = w.walkerLimiter.Acquire(ctx, 1); err == nil {
```

jayhuang-suse: Please read further the codes before make a suggestion. It is a double effort and will make disaster damages since there is not a failure handler at the other sides because those requests are must-executed operations.

IF you do think the extra timeout is required, please elaborate a scenario which could cause a deadlock.

holyspectral: Every syscall could hang if there is a kernel hooking mechanism from a competitor or a kernel bug.  Every I/O call can hang if there is an I/O issue.  There are just too many possibilities that these calls could fail, that's why we want more logs for the unlikely.  The worst errors are those silent ones.

holyspectral: Having said that, I understand your point - the code will work in most of time.  I just don't quite understand adding an error handling using context could be disastrous.  Could you elaborate more about it?

holyspectral: Discussed offline and we agree this is not blocking this PR.

holyspectral (approved):
LGTM.  Thanks for adding the unit test!

PR #1772, opened 2025-01-31
NVSHAS-9325 follow-up: misc fix

jayhuang-suse:
1. patch polymorphic case on controller's healthchecks.
2. remove unnecessary debug traces.


kyledong-suse (approved):
LGTM

PR #1791, opened 2025-02-04
chore: cleanup not used file

holyspectral:
This is to address https://github.com/neuvector/neuvector/pull/1787.  We don't really need github.com/hashicorp/serf.


pohanhuangtw (approved):
LGTM

PR #1821, opened 2025-02-12
NVSHAS-9793: Allow Fed Global roles in ldap/userinitcfg when deployin…

williamlin-suse:
…g NV through ConfigMap and Secret


holyspectral (approved):
LGTM.  Thanks.  The lint issue will be addressed in https://github.com/neuvector/neuvector/pull/1822

PR #1823, opened 2025-02-13
NVSHAS-9810: dispatcher's caller should not hold cacher's mutex.

jayhuang-suse:
(1) A deadlock situation happened during the controller's leadership changes. 
(2) The dispatcher refers cacher's data and its caller should hold the cacher mutex 
(3) Make sure that callers of the dispatcher are free of cacher's mutex.


holyspectral (approved):
LGTM.  Thanks!

PR #1825, opened 2025-02-13
NVSHAS-9824: reduce error messages when the file monitor exited

jayhuang-suse:
It ignores the errors when the target files were removed.


holyspectral (approved):
LGTM.  `os.IsNotExist(err)` can be replaced with `errors.Is(os.ErrNotExist)` for consistency but it's completely optional.

PR #1828, opened 2025-02-19
[NVSHAS-9826] Support auto scan with fine-grained leverl control.

pohanhuangtw:
### Summary
- Support auto scan with fine-grained level control.
- Auto-scan can be configured globally or with fine-grained control for workloads, hosts, or platforms. However, if any of the detailed control fields (enable_auto_scan_workload, enable_auto_scan_host, or enable_auto_scan_platform) is set to false, auto-scan will be disabled overall, because these flags collectively represent a "scan all" setting.

### Test Perform
- set auto-scan through patch /v1/scan/config, will not affect enable_auto_scan_workload, enable_auto_scan_host, or enable_auto_scan_platform
- if set any of enable_auto_scan_workload, enable_auto_scan_host, or enable_auto_scan_platform, it will follow the following _**if enable_auto_scan_workload, enable_auto_scan_host, or enable_auto_scan_platform is set to false, auto-scan will be disabled overall, because these flags collectively represent a "scan all" setting.**_

### DIFF
--- a/controller/api/apis.go
+++ b/controller/api/apis.go
@@ -2244,14 +2244,26 @@ type RESTServiceBatchConfigData struct {
 }
 
 type RESTScanConfig struct {
-	AutoScan bool `json:"auto_scan"`
+	AutoScan bool `json:"auto_scan"` // Deprecated, keeps for backward compatibility
+	// New fields for auto scan, to separate from the old unified auto scan
+	EnableAutoScanWorkload bool `json:"enable_auto_scan_workload"`
+	EnableAutoScanHost     bool `json:"enable_auto_scan_host"`
+	EnableAutoScanPlatform bool `json:"enable_auto_scan_platform"`
 }
 
 type RESTScanConfigConfig struct {
-	AutoScan *bool `json:"auto_scan"`
+	AutoScan *bool `json:"auto_scan"` // Deprecated, keeps for backward compatibility
+	// New fields for auto scan, to separate from the old unified auto scan
+	EnableAutoScanWorkload *bool `json:"enable_auto_scan_workload"`
+	EnableAutoScanHost     *bool `json:"enable_auto_scan_host"`
+	EnableAutoScanPlatform *bool `json:"enable_auto_scan_platform"`
 }
 
 type RESTScanConfigData struct {
+	Config *RESTScanConfigConfig `json:"config"`
+}
+
+type RESTScanConfigResp struct {
 	Config *RESTScanConfig `json:"config"`
 }
 
--- a/controller/api/apis.go
+++ b/controller/api/apis.go
@@ -2248,7 +2248,6 @@ type RESTScanConfig struct {
 	// New fields for auto scan, to separate from the old unified auto scan
 	EnableAutoScanWorkload bool `json:"enable_auto_scan_workload"`
 	EnableAutoScanHost     bool `json:"enable_auto_scan_host"`
-	EnableAutoScanPlatform bool `json:"enable_auto_scan_platform"`
 }
 
 type RESTScanConfigConfig struct {
@@ -2256,7 +2255,6 @@ type RESTScanConfigConfig struct {
 	// New fields for auto scan, to separate from the old unified auto scan
 	EnableAutoScanWorkload *bool `json:"enable_auto_scan_workload"`
 	EnableAutoScanHost     *bool `json:"enable_auto_scan_host"`
-	EnableAutoScanPlatform *bool `json:"enable_auto_scan_platform"`
 }
 
 type RESTScanConfigData struct {
--- a/controller/api/apis.yaml
+++ b/controller/api/apis.yaml
@@ -9147,11 +9147,7 @@ definitions:
         example: false
   RESTScanConfigConfig:
     type: object
-    required:
-      - auto_scan
-      - enable_auto_scan_workload
-      - enable_auto_scan_host
-      - enable_auto_scan_platform
+    minProperties: 1
     properties:
       auto_scan:
         description: set to true to enable all kind of auto-scan
@@ -9168,7 +9164,7 @@ definitions:
       enable_auto_scan_platform:
         description: set to true to enable auto-scan for platform
         type: boolean
-        example: false        
+        example: false
   RESTScanConfigData:
     type: object
     required:
--- a/controller/api/apis.yaml
+++ b/controller/api/apis.yaml
@@ -2367,7 +2367,7 @@ paths:
     patch:
       tags:
         - Scan
-      summary: Update scan configure, Auto-scan can be configured globally or with fine-grained control for workloads, hosts, or platforms. However, if any of the detailed control fields (enable_auto_scan_workload, enable_auto_scan_host, or enable_auto_scan_platform) is set to false, auto-scan will be disabled overall, because these flags collectively represent a "scan all" setting.
+      summary: Update scan configure, Auto-scan can be configured globally or with fine-grained control for workloads, hosts, or platforms. However, if any of the detailed control fields (enable_auto_scan_workload, enable_auto_scan_host) is set to false, auto-scan will be disabled overall, because these flags collectively represent a "scan all" setting.
       security:
         - ApiKeyAuth: []
         - TokenAuth: []
@@ -9127,7 +9127,6 @@ definitions:
       - auto_scan
       - enable_auto_scan_workload
       - enable_auto_scan_host
-      - enable_auto_scan_platform
     properties:
       auto_scan:
         description: set to true to enable all kind of auto-scan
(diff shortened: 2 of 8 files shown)


controller/api/apis.yaml
```diff
@@ -9125,17 +9125,64 @@ definitions:
...
+        description: set to true to enable auto-scan for platform
+        type: boolean
+        example: false
+  RESTScanConfigConfig:
+    type: object
+    required:
+      - auto_scan
+      - enable_auto_scan_workload
+      - enable_auto_scan_host
+      - enable_auto_scan_platform
```
williamlin-suse, line 9154: RESTScanConfigConfig is for PATCH that auto_scan / enable_auto_scan_workload / enable_auto_scan_host / enable_auto_scan_platform should be optional instead of required, correct?

pohanhuangtw: yes, in fact, you can patch any of those.

controller/rest/scanner.go
```diff
@@ -61,13 +126,18 @@ func handlerScanConfig(w http.ResponseWriter, r *http.Request, ps httprouter.Par
 		return
 	}
 
-	cconf := &share.CLUSScanConfig{AutoScan: sconf.Config.AutoScan}
-
+	cconf := &share.CLUSScanConfig{}
```
williamlin-suse, line 129: shouldn't cconf be read from kv first so that the PATCH handler can opt to patch just some of the settings?

pohanhuangtw: you are right, let me work on that

controller/cache/scan.go
```diff
@@ -688,7 +705,7 @@ func scanMapAdd(taskId string, agentId string, idns []api.RESTIDName, objType sh
 			task := &scanTask{id: taskId, priority: scheduler.PriorityHigh}
 			scanScher.AddTask(task, true)
 			updateScanState(taskId, info.objType, api.ScanStatusScheduled)
-		} else if scanCfg.AutoScan {
+		} else if isInfoAutoScanEnabled(info, &scanCfg) {
```
williamlin-suse, line 708: in line 701 there is this comment:
`// Always scan the platform even auto-scan is disabled`
if this comment is still valid, why do we add EnableAutoScanPlatform ?

controller/cache/scan.go
```diff
@@ -784,7 +801,28 @@ func scanAgentDelete(id string, param interface{}) {
...
+		targetEnabled bool
+	}{
+		{src.EnableAutoScanWorkload, target.EnableAutoScanWorkload},
+		{src.EnableAutoScanHost, target.EnableAutoScanHost},
+		{src.EnableAutoScanPlatform, target.EnableAutoScanPlatform},
+		{src.AutoScan, target.AutoScan},
+	}
+
+	// Track if any scan option was enabled or disabled
+	for _, pair := range scanPairs {
```
williamlin-suse, line 817: In the PR conversation it says:
if any of the detailed control fields (enable_auto_scan_workload, enable_auto_scan_host, or enable_auto_scan_platform) is set to false, auto-scan will be disabled overall,
But the logic from lines 817~820 has different result from the conversation:
src means the object in memory cache.
target means the object unmarshalled from kv.
Let's say kv is 
```
{
    AutoScan: true,
}
```
in 5.4.2.
Right after upgrade to 5.4.3, the fist time scanConfigUpdate() is called, src is

```
{
    AutoScan: false,
    EnableAutoScanWorkload: false,
    EnableAutoScanHost: false,
    EnableAutoScanPlatform: false, 
}
```
target is 
```
{
    AutoScan: true,
    EnableAutoScanWorkload: false,
    EnableAutoScanHost: false,
    EnableAutoScanPlatform: false, 
}
```
After the logic from lines 817~820, shouldEnable is true & shouldDisable is false.
It's conflict from the conversation,

pohanhuangtw: Hi @williamlin-suse ,
Thanks for the review, about this. the main logic is actually handle by applyScanConfigUpdates. This guarantee the result is correct that the value store into the kv.

and the scanCfg represent the current config (init kv scan config).

pohanhuangtw: So there is no conflict here.

controller/api/apis.yaml
williamlin-suse: if user can patch any of those, they should not be "required"

controller/cache/scan.go
pohanhuangtw: Hi @williamlin-suse,
This only represents the initial scan task. However, considering auto-scan, we still need it.

pohanhuangtw:
Hi @williamlin-suse ,
The PR is updated, with update api.yaml / applyScanConfigUpdates.

About the doubt

1. update EnableAutoScanWorkload or ... issues, it resolved in applyScanConfigUpdates
2. Always scan the platform even auto-scan => this will not affect the auto-scan platform, so we still needs this
3. api yaml => use minproperty to resolve the issue.

controller/api/apis.go
```diff
@@ -2244,14 +2244,26 @@ type RESTServiceBatchConfigData struct {
...
+	EnableAutoScanPlatform bool `json:"enable_auto_scan_platform"`
 }
 
 type RESTScanConfigConfig struct {
-	AutoScan *bool `json:"auto_scan"`
+	AutoScan *bool `json:"auto_scan"` // Deprecated, keeps for backward compatibility
+	// New fields for auto scan, to separate from the old unified auto scan
+	EnableAutoScanWorkload *bool `json:"enable_auto_scan_workload"`
+	EnableAutoScanHost     *bool `json:"enable_auto_scan_host"`
+	EnableAutoScanPlatform *bool `json:"enable_auto_scan_platform"`
```
williamlin-suse, line 2259: EnableAutoScanPlatform is actually non-configurable in the PATCH api.
If so, RESTScanConfigConfig containing EnableAutoScanPlatform could confuse REST clients that EnableAutoScanPlatform is configurable (also in apis.yaml)

pohanhuangtw: Hi @williamlin-suse,

I’m a bit confused about why EnableAutoScanPlatform is non-configurable in the PATCH API.

Could you elaborate on this? In the issue you mentioned above, it only states that the scan platform will always run when the controller starts. However, this doesn’t seem directly related to the auto scan functionality, as it could still be triggered in previous versions.

williamlin-suse: It looks like in 5.4.2(-) platform scan is always executed on k8s no matter AutoScan is enabled or not, correct?
(i.e. platform scan has nothing to do with AutoScan, correct?)

pohanhuangtw: Correct. After discussing with @jayhuang-suse, he suggested not enabling EnableAutoScanPlatform. Given this, I think we should remove it.

controller/cache/scan.go
```diff
@@ -125,6 +125,20 @@ func scanMutexRUnlock() {
...
+// If objType is not recognized, returns base AutoScan setting
+func isInfoAutoScanEnabled(info *scanInfo, cfg *share.CLUSScanConfig) bool {
+	enableAutoScanBool := cfg.AutoScan
+	switch info.objType {
+	case share.ScanObjectType_CONTAINER:
+		enableAutoScanBool = enableAutoScanBool || cfg.EnableAutoScanWorkload
+	case share.ScanObjectType_HOST:
+		enableAutoScanBool = enableAutoScanBool || cfg.EnableAutoScanHost
+	case share.ScanObjectType_PLATFORM:
+		enableAutoScanBool = false // platform is not supported for auto scan
```
williamlin-suse, line 137: I wonder why keep line 136/137 when platform is not supported for auto scan

hhcs9527:
It's because, when disable or enable auto scan, we need to make sure the platform is not affected

controller/cache/scan.go
pohanhuangtw: This has been tested and it works

controller/api/apis.yaml
```diff
@@ -2363,11 +2363,11 @@ paths:
         '200':
           description: Success
           schema:
-            $ref: '#/definitions/RESTScanConfigData'
+            $ref: '#/definitions/RESTScanConfigResp'
     patch:
       tags:
         - Scan
-      summary: Update scan configure
+      summary: Update scan configure, Auto-scan can be configured globally or with fine-grained control for workloads, hosts, or platforms. However, if any of the detailed control fields (enable_auto_scan_workload, enable_auto_scan_host) is set to false, auto-scan will be disabled overall, because these flags collectively represent a "scan all" setting.
```
williamlin-suse, line 2370: "platforms" should be removed from the summary because it's not configurable

williamlin-suse (approved):

PR #1832, opened 2025-02-21
NVSHAS-9817: Creating NvClusterSecurityRule CRD shows successful crea…

williamlin-suse:
…tion, but it was not actually created due to duplicated process rule entries

When configuring process rules thru REST API, the API handler does merge the rules so that duplicate rules are not treated as error.
CRD should have the same behavior.


jeffhuang4704 (approved):
looks good to me.

PR #1834, opened 2025-02-24
NVSHAS-7997: GitHub Registry Type

alopez-suse:
This PR implements a fixed registry type for the GitHub Container Registry, which allows a user to set namespace/* filters when scanning.

The generic * filter is not allowed when scanning the GitHub container registry, as it would scan every package visible to the scanning user on the GitHub Container Registry. While I could have limited the * filter to use the given username as a namespace, I prefer the unambiguous behavior of just disallowing the filter. Instead, the user can use a filter such as rancher/* to scan every package visible to the user under the Rancher namespace.

The GitHub registry API seems to differentiate between "users" and "organizations" endpoints (referred to here as namespaces) even though the behavior is exactly the same for both of them. Also, a "user" and an "organization" cannot share the same name, so in my implementation, instead of requiring the user to specify whether they are scanning an "organization" or "user" I just try the organization endpoint first, and fallback to the user endpoint in the case of a "not found" error.

### DIFF
--- a/controller/rest/registry.go
+++ b/controller/rest/registry.go
@@ -62,6 +62,7 @@ var registryTypeList []string = []string{
 	share.RegistryTypeGitlab,
 	share.RegistryTypeIBMCloud,
 	share.RegistryTypeHarbor,
+	share.RegistryTypeGitHub,
 }
 var registryTypeSet utils.Set = utils.NewSetFromSliceKind(registryTypeList)
 


controller/scan/github.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return nil, 0, fmt.Errorf("could not read response body: %w", err)
+	}
+	err = json.Unmarshal(respBytes, &packages)
+	if err != nil {
+		return nil, 0, fmt.Errorf("could not unmarshall response body json: %w", err)
+	}
+
+	nextPage := 0
+	if strings.Contains(resp.Header.Get("link"), "rel=\"next\"") {
```
jeffhuang4704, line 140: Is it possible that "link" header can have multiple rel attributes, not just next?
If so, using this way to check might have problem.

controller/scan/github.go
```diff
@@ -0,0 +1,145 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return nil, 0, fmt.Errorf("received error code from github api: %d", resp.StatusCode)
+	}
+
+	packages := []GithubApiPackage{}
+	respBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, 0, fmt.Errorf("could not read response body: %w", err)
+	}
+	err = json.Unmarshal(respBytes, &packages)
```
jeffhuang4704, line 134: We define the `GithubApiPackage` struct to unmarshal the response.
I'm wondering if there is an official document that defines this structure, or if we determined it through observation?

jeffhuang4704 (approved):

holyspectral (approved):

PR #1836, opened 2025-02-24
NVSHAS-9227: skip in toto attestation images when scanning

alopez-suse:
Some customers have these attestation artifacts in their registries, and similar to `.sig` images, these should not be scanned otherwise they will cause errors.


jeffhuang4704 (approved):
looks good to me.

PR #1838, opened 2025-02-25
[NVSHAS-9826] fix cold start with auto-scan true bug

pohanhuangtw:
### Summary
- when the auto scan is true, the others needs to be true.

### Related PR
- https://github.com/neuvector/neuvector/pull/1828

### DIFF
--- a/controller/cache/scan.go
+++ b/controller/cache/scan.go
@@ -130,9 +130,13 @@ func isInfoAutoScanEnabled(info *scanInfo, cfg *share.CLUSScanConfig) bool {
 	enableAutoScanBool := cfg.AutoScan
 	switch info.objType {
 	case share.ScanObjectType_CONTAINER:
-		enableAutoScanBool = enableAutoScanBool || cfg.EnableAutoScanWorkload
+		if cfg.EnableAutoScanWorkload != nil {
+			enableAutoScanBool = *cfg.EnableAutoScanWorkload
+		}
 	case share.ScanObjectType_HOST:
-		enableAutoScanBool = enableAutoScanBool || cfg.EnableAutoScanHost
+		if cfg.EnableAutoScanHost != nil {
+			enableAutoScanBool = *cfg.EnableAutoScanHost
+		}
 	}
 	return enableAutoScanBool
 }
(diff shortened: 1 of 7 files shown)


pohanhuangtw:
HI @alopez-suse ,
Thanks for helping out reviewing this case, it would not, but after I discuss with @williamlin-suse .
We decide to go with different approach.

controller/cache/scan.go
```diff
@@ -625,9 +629,15 @@ func scannerDBChange(newVer string) {
...
-		!scanCfg.AutoScan {
+	enableAutoScanWorkload := false
+	if scanCfg.EnableAutoScanWorkload != nil {
+		enableAutoScanWorkload = *scanCfg.EnableAutoScanWorkload
+	}
+	enableAutoScanHost := false
+	if scanCfg.EnableAutoScanHost != nil {
+		enableAutoScanHost = *scanCfg.EnableAutoScanHost
+	}
+	if !enableAutoScanWorkload && !enableAutoScanHost && !scanCfg.AutoScan {
```
williamlin-suse, line 640: This logic looks like incorrect. 
Let's say the operation sequence is:
1. scanCfg.AutoScan is set to true in 5.4.2
2. NV controller is upgraded to 5.4.3
3. Both enableAutoScanWorkload / enableAutoScanHost are explicitly set to false
It means "do not auto scan for workloads & hosts" but line 640 will be "if !false && !false && !true" and skip line 641 to trigger autoscan!
------
The initial value of local var enableAutoScanWorkloa / enableAutoScanHost should be from scanCfg.AutoScan.
(like in controller/cache/log.go)
When scanCfg.EnableAutoScanWorkload / scanCfg.EnableAutoScanHost is non-nil, their value is set accordingly. 
Then line 640~642 can simply be
```
if !enableAutoScanWorkload && !enableAutoScanHost {
    return
}
```

controller/cache/scan.go
```diff
@@ -798,21 +808,26 @@ func scanAgentDelete(id string, param interface{}) {
 	}
 }
 
-func CompareScanConfig(src, target *share.CLUSScanConfig) (shouldEnable, shouldDisable bool) {
+func CompareScanConfig(currentCfg, updateCfg *share.CLUSScanConfig) (shouldEnable, shouldDisable bool) {
```
williamlin-suse, line 811: I think this function has the same issue as line 640 in this file.

pohanhuangtw:
Hi @williamlin-suse ,
I have been update the compare function and cover it by the unit-test, please take a look

williamlin-suse (approved):
LGTM

holyspectral (approved):

PR #1840, opened 2025-02-26
NVSHAS-9792: Federation policy syncing failing due to body size excee…

williamlin-suse:
…ding Consul max


jeffhuang4704 (approved):
looks good to me.

PR #1848, opened 2025-03-03
NVSHAS-9764: add azure devops repo remote export

alopez-suse:


williamlin-suse (approved):
LGTM
Please update api/apis.yaml about the new types added in api/apis.go before 5.4.3 is released

PR #1850, opened 2025-03-04
NVSHAS-9828: improvement for short-lived pods

jayhuang-suse:
(1) Patch missing learning short-lived pods by extending active container life cycle. 
(2) Maintain enforcing process groups (mode monitor/protect) at nodes 
(3) It is improving to block processes with a valid enforcing process profile.

### DIFF
--- a/agent/policy/process.go
+++ b/agent/policy/process.go
@@ -37,7 +37,6 @@ var k8sGrpProbe utils.Set = utils.NewSet()
 func (e *Engine) UpdateProcessPolicy(name string, profile *share.CLUSProcessProfile) (bool, *share.CLUSProcessProfile) {
 	e.Mutex.Lock()
 	defer e.Mutex.Unlock()
-
 	exist, ok := e.ProcessPolicy[name]
 	if !ok || !reflect.DeepEqual(exist, profile) {
 		e.ProcessPolicy[name] = profile
@@ -57,7 +56,18 @@ func (e *Engine) ObtainProcessPolicy(name, id string) (*share.CLUSProcessProfile
 	e.Mutex.Lock()
 	profile := e.ProcessPolicy[name]
 	e.Mutex.Unlock()
-	if profile != nil { // the process policy per group has been fetched
+	if profile == nil { // a temporary learned group
+		profile = &share.CLUSProcessProfile{
+			Group:    name,
+			Mode:     share.PolicyModeLearn,
+			Process:  make([]*share.CLUSProcessProfileEntry, 0),
+			CfgType:  share.Learned,
+			Baseline: share.ProfileBasic, // avoid false positive events
+		}
+		e.Mutex.Lock()
+		e.ProcessPolicy[name] = profile
+		e.Mutex.Unlock()
+	} else { // the process policy per group has been fetched
 		if grp_profile, ok := e.getGroupRule(id); ok {
 			if grp_profile == nil { // neuvector pods only
 				return profile, true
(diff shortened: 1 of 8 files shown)


agent/policy/process.go
```diff
@@ -73,6 +83,7 @@ func (e *Engine) ObtainProcessPolicy(name, id string) (*share.CLUSProcessProfile
 			}
 			return grp_profile, true
 		}
+		return profile, true
```
williamlin-suse, line 86: Line 68 & 86 make line 90 never reached

jayhuang-suse: correct. let me update it.

jayhuang-suse: Done

kyledong-suse (approved):
LGTM

controller/kv/dispatcher.go
```diff
@@ -481,6 +546,13 @@ func (dpt *kvDispatcher) PutProfile(group, subkey string, value []byte, txn *clu
 			log.WithFields(log.Fields{"group": group, "key": subkey, "error": err, "bPutIfNotExist": bPutIfNotExist}).Error("DPT:")
 		}
 	}
+
+	switch mode {
+	case share.PolicyModeEvaluate, share.PolicyModeEnforce:
+		dpt.setEnforcingGroups(group, true, true)
+	case share.PolicyModeLearn:
+		dpt.setEnforcingGroups(group, false, true)
```
williamlin-suse, line 554: Why lines 552 & 554 both always pass true as the 3rd parameter(bLeader).

jayhuang-suse: The PutProfile function has already has the right as a leader to update the KV entries.

williamlin-suse: So every controller pod eventually calls dpt.copyProfileKeys()/dpt.removeProfileKeys() no matter it's lead or not? If so, why bother always pass the 3rd parameter as true?

jayhuang-suse: It makes sense to remove it. This flag was working for the initial implementation. Let me remove it.

jayhuang-suse: Correction. This flag is required for GC when its workload left and all controller should be acknowledged the workload status.

controller/kv/dispatcher.go
```diff
@@ -317,11 +374,16 @@ func (dpt *kvDispatcher) WorkloadJoin(node, group, id string, customGrps utils.S
...
 //	(1) (current) purge the custom groups immediately(it needs the validatation of all other groups/workloads).
 //	(2) a 30-minutes timer to purge the unaffected custom groups
-func (dpt *kvDispatcher) WorkloadLeave(node, group, id string, customGrps utils.Set, bLeader bool) {
-	// log.WithFields(log.Fields{"node": node, "group": group, "id": id, "customs": customGrps.String()}).Debug("DPT:")
-
+func (dpt *kvDispatcher) WorkloadLeave(node, group, id, mode string, customGrps utils.Set, bLeader bool) {
 	dpt.lock()
 	defer dpt.unlock()
+
+	if mode == share.PolicyModeEvaluate || mode == share.PolicyModeEnforce {
```
williamlin-suse, line 381: So in learning mode dpt.setEnforcingGroups() is never called in func (dpt *kvDispatcher) WorkloadLeave() ?

jayhuang-suse: Yes, the PutProfile() will update it.

williamlin-suse (approved):
LGTM regarding to the controller side changes

PR #1851, opened 2025-03-04
NVSHAS-9856 Update OPA to v1.2.0 and run it in backward-compatible mode

jeffhuang4704:


kyledong-suse (approved):
LGTM
OPA v0 Backwards Compatibility reference, https://www.openpolicyagent.org/docs/latest/v0-compatibility/

holyspectral (approved):
LGTM

PR #1862, opened 2025-03-06
Review golanci-lint settings

pjbgf:


.github/workflows/golangci-lint.yml
```diff
@@ -1,28 +1,29 @@
...
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v6
+        uses: golangci/golangci-lint-action@2226d7cb06a077cd73e56eedd38eecad18e5d837 # v6.5.0
         with:
-          version: v1.60
-          args: --timeout=30m --new-from-rev=${{ github.event.pull_request.base.sha }}
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          version: latest
+          install-mode: binary
+          args: --timeout=30m
```
holyspectral, line 29: The code here was there to deal with a large PR with a lot of diff.  (For example, dependencies update or introducing new packages.) https://github.com/neuvector/neuvector/pull/1642  Would the new version of golangci-lint be able to support this scenario?

pjbgf: I believe so, as it will be running the checks on the current cloned version - based on the point in time state on that PR. Instead of running it on the diff only.

holyspectral (approved):

PR #1871, opened 2025-03-10
fix: staticcheck and gosimple findings

holyspectral:
So far we've fixed most of issues reported by golangci-lint.  These are the remaining ones.  

```
S1009: should omit nil check; len() for nil slices is defined as zero (gosimple)
SA1015: using time.Tick leaks the underlying ticker, consider using it only in endless functions, tests and the main package, and use time.NewTicker here (staticcheck)
SA1019: grpc.DialContext is deprecated: use NewClient instead.  Will be supported throughout 1.x. (staticcheck)
```

For SA1015, our usage is reviewed to make sure our usage is alright - no ticker leak.
For SA1019, because gRPC won't deprecate those APIs until its next major version, these findings are marked as nolint. 

After this PR, we don't have to run golangci-lint against only diff anymore.  This will help us to discover more issues using golangci-lint.

### DIFF
--- a/agent/dp/ctrl.go
+++ b/agent/dp/ctrl.go
@@ -273,7 +273,7 @@ func DPCtrlAddMAC(iface string, mac, ucmac, bcmac, oldmac, pmac net.HardwareAddr
 			PIPS:   tpips,
 		},
 	}
-	if pips == nil || len(pips) <= 0 {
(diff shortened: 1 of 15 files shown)


kyledong-suse (approved):
LGTM from agent perspective

PR #1874, opened 2025-03-11
NVSHAS-9847: fix broken dockerhub urls

alopez-suse:
A previous change to use `url.JoinPath` was causing the query parameters to get escaped, meaning they would be interpreted as the repo part of the request path instead of actual query parameters.

This change reverts the previous behavior, where the query parameters are literally appended to the url, instead of escaped during `JoinPath`. Ideally this would be changed in a more _semantic_ way but this is the simplest way to fix this without affecting other parts of the codebase.


williamlin-suse (approved):
LGTM

PR #1881, opened 2025-03-12
NSHAS-9875: add default commit comment for azure devops export

alopez-suse:
It is possible that user does not enter a commit comment, however an empty `comment` results in a `400` from the Azure Devops repo API. A default comment is necessary.


williamlin-suse (approved):
LGTM

PR #1883, opened 2025-03-12
fix: incorrect version format

holyspectral:
In 5.4.2, the version format in golang space missed a v prefix.  This commit fixed the issue.


williamlin-suse (approved):
The fix for "non-release builds do not send telemetry data (because of "interim/" prefix in Version)" is not covered in this PR

williamlin-suse (approved):
LGTM

BinX-Suse (approved):

PR #1890, opened 2025-03-17
NVSHAS-9828: fix false-positive asynchronous processes checkup

jayhuang-suse:
(1) Avoid the false-positive runtime processes prevention on nv pods. 
(2) Identify correct nv pod roles
(3) Reduce file events on the not-modified files


holyspectral (approved):
LGTM

PR #1896, opened 2025-03-18
fix: NVSHAS-9884 prevent half-baked agent info

holyspectral:
Prevent half-baked agent info from sending to controller, which would lead to connection issue in scanner.


gfsuse (approved):
LGTM

PR #1897, opened 2025-03-19
Potential race condition

jayhuang-suse:
The situation is unlikey occured because of the downstream profile data is at least 5-7 sec apart. However, it good to make the prevetion.


kyledong-suse (approved):
LGTM

PR #1924, opened 2025-04-14
NV9905: k8s workload scan returns image metadata

jayhuang-suse:
K8s only: add returning image repo into the scan data. Later on, the scanner can utilize the information to match its cve database.

Add new module type and its associated structure:

```
Kubernetes = "kubernetes"

const (
	TypeK8sRepo = "k8sRepo"
)

type KubernetesResource struct {
	ResourceType string `json:"type"`
	Name         string `json:"name"`
}
```


pohanhuangtw (approved):
LGTM

PR #1928, opened 2025-04-18
NV9919: Enforcer keep crashing because monitor getting killed

jayhuang-suse:
k8s docker-specific: the parent of the monitor becomes systemd


kyledong-suse (approved):
LGTM

PR #1940, opened 2025-04-30
NV9899: Process Profile Alerts contain sensitive data

jayhuang-suse:
Add mask mechanism on the command line when detecting keyword "password"


agent/probe/process.go
```diff
@@ -1963,7 +1963,28 @@ func (p *Probe) inspectNewProcesses(bInit bool) {
 	}
 }
 
+func maskPasswordString(cmds []string) {
+	maskNextCmd := false
+	// Iterate using index to modify the slice directly
+	for i, cmd := range cmds {
```
holyspectral, line 1969: Maybe we should start from index=1 as cmds[0] is the executable path?

holyspectral (commented):
Thanks for adding unit-tests, which make it very easy to understand.  Just a little comment about which argument to check.

agent/probe/process.go
jayhuang-suse: If there is a command like below, it will become false-negative case and miss the masking. 

/bin/InputPassword "secret" .... 

Overalls speaking, the cmds fields are reference-only and will not affect the detection. The overwritten will not cause a problem in the incident reports.

holyspectral: yeah totally agree that it's hard to make it right and cover all cases... @pjbgf has shared his thoughts in rst channel.  Maybe we should move the discussion there.

PR #1958, opened 2025-05-06
fix: improve the error logging

pohanhuangtw:
### Summary
- The new logging is confusing (from customer), so I make some change to improve the logging

### DIFF
--- a/controller/rest/longpoll.go
+++ b/controller/rest/longpoll.go
@@ -59,7 +59,9 @@ func NewJobError(code int, err error, detail interface{}) *JobError {
 
 func (e *JobError) Error() string {
 	if e.Err != nil {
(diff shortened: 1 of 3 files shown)


controller/rest/longpoll.go
```diff
@@ -59,7 +59,9 @@ func NewJobError(code int, err error, detail interface{}) *JobError {
 
 func (e *JobError) Error() string {
 	if e.Err != nil {
-		return fmt.Sprintf("JobError Code: %d, Error: %v", e.Code, e.Err)
+		return e.Err.Error()
```
williamlin-suse, line 62: should `e.Code` be printed out as before?

pohanhuangtw: Already addressed in the wrapper function.

PR #1960, opened 2025-05-07
NVSHAS-9791: read runtime dependency not dev dependency

pohanhuangtw:
### Summary
- In .NET applications, multiple *.deps.json files may exist, but only the runtime dependencies are relevant for identifying actual vulnerabilities — not the development-time dependencies.
  - Currently, NV including all dependencies by default, regardless of whether they’re actually used at runtime.
  - Incorporating the `runtime` section into our logic would make this more accurate.
- add `os.Getenv("SCAN_DOTNET_RUNTIME")` to ensure backward compatible
Here are some `.deps.json` snippets for context:

```json
"Microsoft.AspNetCore.Authentication.OpenIdConnect/7.0.0": {
  "dependencies": {
    "Microsoft.IdentityModel.Protocols.OpenIdConnect": "6.15.1"
  },
  "runtime": {
    "lib/net7.0/Microsoft.AspNetCore.Authentication.OpenIdConnect.dll": {
      "assemblyVersion": "7.0.0.0",
      "fileVersion": "7.0.22.51819"
    }
  }
},
"Microsoft.AspNetCore.Http/2.2.2": {
  "dependencies": {
    "Microsoft.AspNetCore.Http.Abstractions": "2.2.0",
    "Microsoft.AspNetCore.WebUtilities": "2.2.0",
    "Microsoft.Extensions.ObjectPool": "2.2.0",
    "Microsoft.Extensions.Options": "8.0.2",
    "Microsoft.Net.Http.Headers": "2.2.0"
  }
},
"Microsoft.AspNetCore.Http.Abstractions/2.2.0": {
  "dependencies": {
    "Microsoft.AspNetCore.Http.Features": "2.2.0",
    "System.Text.Encodings.Web": "4.5.0"
  }
}
```

### DIFF
--- a/share/scan/apps.go
+++ b/share/scan/apps.go
@@ -106,8 +106,13 @@ type mvnDependency struct {
 }
 */
 
+type dotnetRuntimeDetail struct {
+	AssemblyVersion string `json:"assemblyVersion"`
+	FileVersion     string `json:"fileVersion"`
+}
 type dotnetDependency struct {
-	Deps map[string]string `json:"dependencies"`
+	Deps    map[string]string              `json:"dependencies"`
+	Runtime map[string]dotnetRuntimeDetail `json:"runtime"`
 }
 
 type dotnetRuntime struct {
@@ -557,7 +562,7 @@ func isRuby(filename string) bool {
 }
 
 func isDotNet(filename string) bool {
-	return strings.HasSuffix(filename, ".App.deps.json")
+	return strings.HasSuffix(filename, ".deps.json")
 }
 
 func isWordpress(filename string) bool {
@@ -733,7 +738,10 @@ func parseDotNetJsonData(filename string, fullpath string, dotnet dotnetPackage)
 
 	if targets, ok := dotnet.Targets[dotnet.Runtime.Name]; ok {
 		for target, dep := range targets {
-
+			// Skip if the target has no runtime; it means the dependency is not actually installed.
+			if dep.Runtime == nil && os.Getenv("SCAN_DOTNET_RUNTIME") == "true" {
+				continue
+			}
 			//Get dependencies of individual targets
 			for app, v := range dep.Deps {
 				key := fmt.Sprintf("%s-%s", ".NET:"+app, v)
--- a/share/scan/apps.go
+++ b/share/scan/apps.go
@@ -557,7 +557,7 @@ func isRuby(filename string) bool {
 }
 
 func isDotNet(filename string) bool {
-	return strings.HasSuffix(filename, ".deps.json")
+	return strings.HasSuffix(filename, ".App.deps.json")
 }
 
 func isWordpress(filename string) bool {
--- a/share/scan/apps.go
+++ b/share/scan/apps.go
@@ -739,7 +739,8 @@ func parseDotNetJsonData(filename string, fullpath string, dotnet dotnetPackage)
 	if targets, ok := dotnet.Targets[dotnet.Runtime.Name]; ok {
 		for target, dep := range targets {
 			// Skip if the target has no runtime; it means the dependency is not actually installed.
-			if dep.Runtime == nil && os.Getenv("SCAN_DOTNET_RUNTIME") == "true" {
+			// if SCAN_DOTNET_RUNTIME is not set, we don't need to skip
+			if dep.Runtime == nil && os.Getenv("SCAN_DOTNET_RUNTIME") != "" {
 				continue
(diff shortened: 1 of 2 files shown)


jayhuang-suse (commented):
Can you provide an online link to differ its runtime dependency from other dependencies?

pohanhuangtw:
Hi @jayhuang-suse ,
Thanks for the review.
Actually, I haven’t found an official way to distinguish runtime dependencies from development ones through a direct link or online tool.

However, when I inspect the .deps.json files using grep, I notice a pattern like this:

```bash
./<hash>/usr/share/dotnet/shared/Microsoft.AspNetCore.App/8.0.11/Microsoft.AspNetCore.App.deps.json  
./<hash>/app/Swimlane.SecurityIntelligence.deps.json  
./<hash>/usr/share/dotnet/shared/Microsoft.NETCore.App/8.0.11/Microsoft.NETCore.App.deps.json  
Based on this, I assume the files under usr/share/dotnet/shared/ are runtime framework dependencies provided by the .NET runtime, while the one under /app/ is specific to the application and includes both runtime and compile-time dependencies.
```
Also, Trivy’s documentation gives a bit more context here:
https://trivy.dev/latest/docs/coverage/language/dotnet/

It seems Trivy prioritizes .deps.json found under the app directory, which likely reflects actual runtime usage.

pohanhuangtw:
Hi @jayhuang-suse,

**TL;DR**:
I propose to **only consider dependencies that have a corresponding `runtime` entry**.
If a dependency doesn't include a `runtime` section, we’ll ignore it.

Here are some `.deps.json` snippets for context:

```json
"Microsoft.AspNetCore.Authentication.OpenIdConnect/7.0.0": {
  "dependencies": {
    "Microsoft.IdentityModel.Protocols.OpenIdConnect": "6.15.1"
  },
  "runtime": {
    "lib/net7.0/Microsoft.AspNetCore.Authentication.OpenIdConnect.dll": {
      "assemblyVersion": "7.0.0.0",
      "fileVersion": "7.0.22.51819"
    }
  }
},
"Microsoft.AspNetCore.Http/2.2.2": {
  "dependencies": {
    "Microsoft.AspNetCore.Http.Abstractions": "2.2.0",
    "Microsoft.AspNetCore.WebUtilities": "2.2.0",
    "Microsoft.Extensions.ObjectPool": "2.2.0",
    "Microsoft.Extensions.Options": "8.0.2",
    "Microsoft.Net.Http.Headers": "2.2.0"
  }
},
"Microsoft.AspNetCore.Http.Abstractions/2.2.0": {
  "dependencies": {
    "Microsoft.AspNetCore.Http.Features": "2.2.0",
    "System.Text.Encodings.Web": "4.5.0"
  }
}
```

Currently, we’re including all dependencies by default, regardless of whether they’re actually used at runtime.
Incorporating the `runtime` section into our logic would make this more accurate.
It would help us focus on what’s truly executed, avoid false positives, and skip compile-time or transitive dependencies that don’t end up in the final application.

Let me know if you foresee any concerns with this approach.

cc: @holyspectral

pohanhuangtw:
not sure why is the format fail here, I use the gofmt -w. try to fix, but still not working.
@holyspectral any good suggestions?

jayhuang-suse:
The test scanner image will manually be tested against nightly tests. Let hear the feedbacks of the results. It is about the dotNet packages and expect some dropped cases and please give an example why those missing CVEs are false-positive cases.

pohanhuangtw:
Hi @jayhuang-suse , @holyspectral 
This has been test by QA, I think we can merge it.

jayhuang-suse:
I understand this package shared by enforcer and scanner but I am not sure that our code standard can accept taking any environment variable in a normal function call. @holyspectral 

"os.Getenv("SCAN_DOTNET_RUNTIME")"

holyspectral:
@jayhuang-suse sorry I don't really get what you mean.  Can you elaborate more?

holyspectral:
@jayhuang-suse that's a good question.  In terms of security, an environment variable is actually better than a command line parameter because you need extra permission to access its environment variables.  That's why generally command line parameters should be avoided.  Having said that, for an on/off switch I think both are acceptable.

I don't think NeuVector has a strict guideline about which to use at the moment.  It's worth noting that helm charts doesn't support overriding command line parameters, so it's fine with me if we want to use an environment variable instead, especially for a non user-facing configuration.

holyspectral (approved):

PR #1968, opened 2025-05-09
docs: fix api docs for the admission control rules test endpoint

lentus:
The API documentation for the endpoint `/v1/assess/admission/rule` seems to be outdated/incorrect: the documentation claims this endpoint consumes an `application/json` body using the `RESTAdmissionRuleConfigData` schema, but actually expects a yaml string containing a kubernetes resource.

I have changed the required type to a UTF8-encoded plain text string (like the `/v1/file/compliance/profile/config` endpoint), and added a description and example.

### DIFF
--- a/controller/api/apis.yaml
+++ b/controller/api/apis.yaml
@@ -298,19 +298,42 @@ paths:
     post:
       tags:
         - Admission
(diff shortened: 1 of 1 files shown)


williamlin-suse (approved):
LGTM

Issue #1969, opened 2025-05-09
Admission control rules testing endpoint requires Admission Control Modify permission


lentus:
**Environment**
Platform: all
Kubernetes/Platform Version(s): -

**Describe the bug**
During a CI pipeline, before deploying to an environment, I would like to test whether a deployment resource violates any admission control rules. This is possible by using the admission control rules testing endpoint (`/v1/assess/admission/rule`), which gives a great summary of what rules were used and which would block deployment. However, this endpoint currently requires modify permission for admission control configuration, despite not modifying any admission control rules or configuration. This allows API keys created to _test_ admission control rules to _change_ admission control rules as well, which is not something I want my CI pipeline to be able to do.

The current implementation does not explicitly specify whether this endpoint performs a read or write operation. Because of this the access operation is determined by checking the default for the used request method, which results in write since this is a `POST` endpoint.

**To Reproduce**
1. Create a custom role with "Admission Control (V)" permission
2. Create an API key with the new custom role as "global role"
3. Use the API key (in the "X-Auth-Apikey" header) to call the `/v1/assess/admission/rule` endpoint using some kind of kubernetes resource yaml file as body
4. See access denied error

When creating an API key based on a custom role with "Admission Control (M)" permission, I get the expected admission control assessment result.

**Expected behavior**
Since this endpoint does not change any admission control configuration I would like to be able to call it using "Admission Control (V) permission". I have implemented and tested this in #1970.

fivemserverhostings:
The performance optimizations here are quite clever. Nice engineering work!

lentus:
Closing this issue since my PR has been merged!

lentus:
@williamlin-suse Any idea when this change will be included in a release?

williamlin-suse:
It will be available in NV 5.4.5 (~ mid July/2025)

PR #1970, opened 2025-05-09
fix(controller): make admission control rules test endpoint accessible with read permission

lentus:
See #1969

The current implementation does not explicitly specify whether this endpoint performs a read or write operation. Because of this the access operation is determined by checking the default for the used request method, which results in write since this is a `POST` endpoint. Because of this, you need an API key for a role with `Admission Control (M)` permission to call this endpoint.

By specifying the access operation as `access.AccessOPRead`, an API key using a role with the `Admission Control (V)` permission can be used to test admission control rules.

### DIFF
--- a/controller/rest/assessment.go
+++ b/controller/rest/assessment.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


williamlin-suse (approved):
LGTM

PR #1983, opened 2025-05-22
NVSHAS-9960: Scanner not working.

pohanhuangtw:
### Root cause
- wlCacheMap not sync with scanMap
- if enforcer is down, the scanMap will delete the item created by the agent, but wlCacheMap doesn't, and it lead to the in-consistent and unable to scan workload properly.

### Solution
Sync both map in deletion.
- When the [agent is removed](https://github.com/neuvector/neuvector/blob/7f856d6e0935bb9d62bf167cd9e2aa4228fefb74/controller/cache/scan.go#L808), remove related item in both scanMap and wlCacheMap.
- When delete the [wlCacheMap](https://github.com/neuvector/neuvector/blob/7f856d6e0935bb9d62bf167cd9e2aa4228fefb74/controller/cache/object.go#L1563), remove the related scanMap item as well
- Add `scanWorkloadAdd(id, cache)` to ensure the item in scanMap exist, because these two map should sync.

### DIFF
--- a/controller/cache/object.go
+++ b/controller/cache/object.go
@@ -1560,6 +1560,7 @@ func workloadUpdate(nType cluster.ClusterNotifyType, key string, value []byte) {
 		cacheMutexLock()
 
 		if wlCache, ok = wlCacheMap[id]; ok {
+			scanMapDelete(id) // Remove workload from scanMap
 			delete(wlCacheMap, id)
 
 			// Update parent's children list.
--- a/controller/cache/scan.go
+++ b/controller/cache/scan.go
@@ -368,12 +368,15 @@ func scanObject(id string) {
 }
 
 func (m CacheMethod) ScanWorkload(id string, acc *access.AccessControl) error {
-	if cache := getWorkloadCache(id); cache == nil {
+	var cache *workloadCache
+	if cache = getWorkloadCache(id); cache == nil {
 		return common.ErrObjectNotFound
 	} else if !acc.Authorize(&share.CLUSWorkloadScanDummy{Domain: cache.workload.Domain}, nil) {
 		return common.ErrObjectAccessDenied
 	}
 
+	// Ensure the workload object is in scanMap
+	scanWorkloadAdd(id, cache)
 	scanObject(id)
 	return nil
 }
@@ -797,13 +800,15 @@ func scanHostDelete(id string, param interface{}) {
 func scanAgentDelete(id string, param interface{}) {
 	// purge incomplete scanning jobs but keep completed scans
 	scanMutexLock()
+	cacheMutexRLock()
+	defer cacheMutexRUnlock()
 	defer scanMutexUnlock()
-	for task, info := range scanMap {
+
+	for workloadID, info := range scanMap {
 		if info.agentId == id {
-			if info.cveDBCreateTime == "" { // incompleted, not done yet
-				log.WithFields(log.Fields{"task": task, "info": info}).Info()
-				delete(scanMap, task)
-			}
+			log.WithFields(log.Fields{"workloadID": workloadID, "info": info}).Info()
+			delete(scanMap, workloadID)
+			delete(wlCacheMap, workloadID)
 		}
 	}
 }


controller/cache/object.go
```diff
@@ -1560,6 +1560,7 @@ func workloadUpdate(nType cluster.ClusterNotifyType, key string, value []byte) {
 		cacheMutexLock()
 
 		if wlCache, ok = wlCacheMap[id]; ok {
+			scanMapDelete(id) // Remove workload from scanMap
```
williamlin-suse, line 1563: Is line 1563 necessary?
This code already exists:
```
	evhdls.Register(EV_WORKLOAD_STOP, []eventHandlerFunc{
		hostWorkloadStop,
		addrWorkloadStop,
		connectWorkloadDelete,
		groupWorkloadLeave,
		scanWorkloadDelete,
	})
```
scanWorkloadDelete() calls scanMapDelete().

controller/cache/scan.go
```diff
@@ -797,13 +800,15 @@ func scanHostDelete(id string, param interface{}) {
...
+
+	for workloadID, info := range scanMap {
 		if info.agentId == id {
-			if info.cveDBCreateTime == "" { // incompleted, not done yet
-				log.WithFields(log.Fields{"task": task, "info": info}).Info()
-				delete(scanMap, task)
-			}
+			log.WithFields(log.Fields{"workloadID": workloadID, "info": info}).Info()
+			delete(scanMap, workloadID)
+			delete(wlCacheMap, workloadID)
```
williamlin-suse, line 811: Line 811 means when an agent is deleted, all known pods' info on that agent's host are gone as well.
Do we really want to have this behavior?

controller/cache/scan.go
```diff
@@ -368,12 +368,15 @@ func scanObject(id string) {
...
-	if cache := getWorkloadCache(id); cache == nil {
+	var cache *workloadCache
+	if cache = getWorkloadCache(id); cache == nil {
 		return common.ErrObjectNotFound
 	} else if !acc.Authorize(&share.CLUSWorkloadScanDummy{Domain: cache.workload.Domain}, nil) {
 		return common.ErrObjectAccessDenied
 	}
 
+	// Ensure the workload object is in scanMap
+	scanWorkloadAdd(id, cache)
```
williamlin-suse, line 379: If the agent is gone on the host of the workload, line 379 could add the workload to scanMap but there is still no agent for scanning the workload & the caller of ScanWorkload() still gets a nil error. 
Why not just return an error to the caller when the agent is unavailable ?

pohanhuangtw: I think you are right about this.
However, if the agent is gone, the scan object should encounter the error.

This line just to ensure the sync of scanMap and wlcacheMap.

What do you think?

controller/cache/object.go
pohanhuangtw: Does the workload stop means the same as workload delete? I haven't tried it, if that's the case, I should remove it

controller/cache/scan.go
williamlin-suse: I think we should not call scanWorkloadAdd(id, cache) unconditionally.
Could we check the agent of the workload(by `cache`) is in agentCacheMap or not before calling scanWorkloadAdd(id, cache)?
(If not found, return error instead of keep calling scanObject(id) so that caller knows the `scan workload` fails with reason.)

controller/cache/scan.go
williamlin-suse: If you delete the enforcer pod & go to UI -> Assets -> Containers page, you will see it like:
image
This UI design looks like workload entries in wlcacheMap should not be deleted immediately after the agent stops.

controller/cache/scan.go
gfsuse: when agent is down there is a 10 minute wait before agent is really removed from cache, the logic is complex, pls check this function deleteAgentFromCache for reference

controller/cache/scan.go
pohanhuangtw: I agree with what you said to some extent, but when the agent starts, it adds the node’s workload into the cache.

The wlcacheMap uses the image ID as the key, which stays the same even if the agent goes down.

So, if the old data isn’t removed, the cache may still hold outdated information from the previous agent, which could lead the scan container to connect to the wrong enforcer.
What do you think?

jayhuang-suse:
@pohanhuangtw William found the loophole and make another PR which will include your fix as well. Please close this PR and you can follow up his PR (https://github.com/neuvector/neuvector/pull/1986) later.

williamlin-suse:
This PR is replaced by https://github.com/neuvector/neuvector/pull/1986

PR #1991, opened 2025-05-31
NVSHAS-9969: concurrent map writes results in enforcer component restart

williamlin-suse:
enforcer issue has been fixed in prior version.
this PR is for improving logging & events aggregation found in customer's env

### DIFF
--- a/controller/rest/admwebhook.go
+++ b/controller/rest/admwebhook.go
@@ -590,9 +590,6 @@ func getOwnerUserGroupMetadataFromK8s(kind, name, ns string) (string, utils.Set,
 					replicas = *rsObj.Spec.Replicas
 				}
 				ownerReferences = rsObj.OwnerReferences
-				if rsObj.Spec.Replicas != nil {
-					replicas = *rsObj.Spec.Replicas
-				}
 			}
 		case resource.RscTypeDeployment:
 			if deployObj, ok := obj.(*appsv1.Deployment); ok {
--- a/controller/rest/admwebhook.go
+++ b/controller/rest/admwebhook.go
@@ -132,16 +132,14 @@ func checkAggrLogsCache(alwaysFlush bool) {
 		if alog != nil {
 			if alwaysFlush || time.Now().Unix() > alog.ReportedAt.Unix() {
 				delete(aggrLogsCache, key)
-				if alog.Count > 0 {
-					// use the last log's time as the aggregated log's ReportedAt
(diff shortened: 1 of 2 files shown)


controller/rest/admwebhook.go
```diff
@@ -569,26 +567,41 @@ func mergeMaps(labels1, labels2 map[string]string) map[string]string {
...
+					return "", utils.NewSet(), mergeMaps(ssObj.Labels, ssObj.Spec.Template.Labels),
+						mergeMaps(ssObj.Annotations, ssObj.Spec.Template.Annotations), replicas, true
 				}
 			}
 		case resource.RscTypeReplicaSet:
 			// support pod -> replicaset -> deployment for now
 			if rsObj, ok := obj.(*appsv1.ReplicaSet); ok {
+				if rsObj.Spec.Replicas != nil {
+					replicas = *rsObj.Spec.Replicas
+				}
```
holyspectral, line 591: Why do we have to do this part the second time at Ln 593-595?

controller/rest/admwebhook.go
```diff
@@ -599,32 +612,34 @@ func getOwnerUserGroupMetadataFromK8s(kind, name, ns string) (string, utils.Set,
...
-				if userName, groups, labels, annotations, found := getOwnerUserGroupMetadataFromK8s(strings.ToLower(ownerRef.Kind), ownerRef.Name, ns); found {
-					return userName, groups, labels, annotations, true
+				if userName, groups, labels, annotations, replicas, found :=
+					getOwnerUserGroupMetadataFromK8s(strings.ToLower(ownerRef.Kind), ownerRef.Name, ns); found {
+					return userName, groups, labels, annotations, replicas, true
 				}
 			}
 		}
-	} else {
+	} else if err != resource.ErrResourceNotSupported {
```
holyspectral, line 624: Shouldn't we log any unknown errors?

williamlin-suse: This is one of the reasons that a large amount of ERROR logs appear in customer's logs in a short period of time.
When a k8s deployment(or daemonset, ...) is managed by some framework, the root owning resource is custom resource, not k8s standard resource, that NV doesn't have access for.
It means every time NV controller handles an admission control request in such kind of env, an ERROR log is generated just because NV controller cannot access that custom resource.
That's why we do not want to log error for unsupported resource(types).
If it's unknown error for supported (k8s standard) resources, it's still logged.

controller/rest/admwebhook.go
williamlin-suse: You are correct. It's duplicate & should be removed

holyspectral (approved):
LGTM

PR #2006, opened 2025-06-12
fix: quay.io is unable to use wildcard properly

pohanhuangtw:
### Summary
- Wildcard not work properly in quay.io, because quay.io use other api to fetch the repo.
- Only consider the cases for explicitly write quay, since there may not much user using this, plus it requires extra UI change if we make this as a new type.

### DIFF
--- a/controller/scan/image.go
+++ b/controller/scan/image.go
@@ -256,7 +256,14 @@ func (r *base) GetRepoList(org, name string, limit int) ([]*share.CLUSImage, err
 		}
 	}
 
-	repos, err := r.rc.Repositories()
+	var repos []string
+	var err error
+	if strings.Contains(r.regURL, "quay.io") {
+		repos, err = r.rc.QuayRepositories(org)
+	} else {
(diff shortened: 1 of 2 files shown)


controller/scan/image.go
```diff
@@ -256,7 +256,14 @@ func (r *base) GetRepoList(org, name string, limit int) ([]*share.CLUSImage, err
 		}
 	}
 
-	repos, err := r.rc.Repositories()
+	var repos []string
+	var err error
+	if strings.Contains(r.regURL, "quay") {
```
holyspectral: Could we make this check stricter, e.g., checking `quay.io` instead?

pohanhuangtw: Hi @holyspectral ,
Thanks for the review. Yes, I think we can do so, please check the update PR.
Thanks

holyspectral (approved):
LGTM.  Please make sure you squash the commits.

pohanhuangtw:
Just did it, can you help confirm? thanks

PR #2013, opened 2025-06-16
NVSHAS-9964: Scan fails after upgrading NeuVector to 5.4.1 due to rep…

williamlin-suse:
…ort size


controller/cache/scan.go
```diff
@@ -843,7 +845,7 @@ func putGZipKVChunks(key string, data []byte) error {
 	return nil
 }
 
-func getGZipKVChunks(key string) ([]byte, error) {
+func getGZipKVChunks(key, id string, objType share.ScanObjectType) {
```
holyspectral: Given that you've moved the handling of the original return values into this function, maybe you should give it another name, e.g., `extractScanReportChunks()`?

PR #2015, opened 2025-06-17
NVSHAS-9993: Replace md5 by sha256

williamlin-suse:


share/utils/utils.go
```diff
@@ -150,19 +150,8 @@ func GzipBytes(buf []byte) []byte {
...
-}
-
 func GetGuid() (string, error) {
-	b := make([]byte, 48)
-	if _, err := io.ReadFull(rand.Reader, b); err != nil {
-		return "", err
-	}
-
-	return GetMd5(base64.StdEncoding.EncodeToString(b)), nil
+	return strings.Replace(uuid.NewString(), "-", "", -1), nil
```
pohanhuangtw, line 154: Should we change this to use a UUID instead?

However, using a UUID would make the ID non-deterministic and unpredictable.
I think it’s better to switch to SHA (e.g., SHA-256) so the ID remains consistent for the same input, while still being more secure than MD5.

pohanhuangtw, line 154: Should we change this to use a UUID instead?

However, using a UUID would make the ID non-deterministic and unpredictable.
I think it’s better to switch to SHA (e.g., SHA-256) so the ID remains consistent for the same input, while still being more secure than MD5.

pohanhuangtw (commented):
Overall, LGTM
I just have a minor question regarding the use of GetGuid. Would you mind taking a quick look to see if any adjustment might be needed?

share/utils/utils.go
williamlin-suse: UUID is 128-bit number(typically represented as a 36-character string) & sha256 is 256-bit number.
Using sha256, instead of md5, to generate a UUID is not more secure.
There is no input parameter for GetGuid(). What does "ID remains consistent for the same input" mean?
GetGuid() should not return the same string

pohanhuangtw: you are right, I misunderstand the caller.

holyspectral (commented):
Generally it looks good to me.  Would this affect multi-cluster scenario, given that we are removing some fields from its API response?

PR #2021, opened 2025-06-24
NVSHAS-9940: NV scan JFrog Subdomain mode issue

williamlin-suse:


controller/scan/jfrog.go
```diff
@@ -219,9 +272,18 @@ func getSubdomainFromRepo(repo string) (string, string) {
 // GetArtifactoryTags is designed to work with a new API endpoint provided by JFrog.
 func (r *jfrog) GetArtifactoryTags(repositoryStr string, rc *scanUtils.RegClient) ([]string, error) {
 	tags := make([]string, 0)
-	key, repository, found := strings.Cut(repositoryStr, "/")
-	if !found {
-		return tags, fmt.Errorf("invalid repository format: %v", repositoryStr)
+	key := r.subdomainInRegURL
```
pohanhuangtw: If JFrog is not using the sub-domain style right now, will this branch still work?
In that case the key will be empty.

**Suggestion**: We could branch the logic explicitly by mode:

```go
if r.mode == share.JFrogModeSubdomain {
    // handle sub-domain scenario
} else {
    // existing path for path-style or default mode
}
```
This makes the intent clear and prevents nil/empty-key issues when the registry is in the “path-style” mode.

what do you think?

pohanhuangtw (commented):
LGTM overall  — I just have one small question.
Thanks—let me know what you think! 🙏

controller/scan/jfrog.go
williamlin-suse: ok. changed code as advised

pohanhuangtw (approved):
LGTM

PR #2029, opened 2025-07-02
chore: switch to obs opa

holyspectral:
Now OPA is available on OBS, so switch to the OBS version for better maintenance.


holyspectral:
Note this will also bring OPA to 1.6.0, which was released two days ago.

BinX-Suse (approved):

kyledong-suse (approved):
LGTM

PR #2040, opened 2025-07-10
New step to automate update upgrade-responder

Anarkis:
# What
This PR brings an automated way to generate an intermediate file, which will be used to release a new `update-upgrade-responder` in `EIO` processes 

# Why
We want to remove manual actions in order to release a new `update-upgrade-responder`

# How
There is a new step inside [release-workflow](https://github.com/neuvector/neuvector/blob/main/.github/workflows/release.yml)  called `update-upgrade-responder`, which will generate a file `.github/upgrade-responder.json`  located at [.github/](https://github.com/neuvector/neuvector/tree/main/.github) with the same content as `Neuvector` team creates the requests to `EIO`. 
For example:
```json
{
  "Versions": [
    {
      "Name": "v5.4.4",
      "ReleaseDate": "2025-05-13T01:31:29.99Z",
      "Tags": [
        "latest", "5.4.4"
      ]
    }
  ]
}
```
This new file, will be queried by internal `EIO` processed which will run once per day. If it finds a difference, it will populate the latest version found on `neuvector` repo over the `EIO` infra, which will lead to a deployment with the new version.

The file creation will be triggered by a `tag` creation from `neuvector` team.

As we can see https://github.com/neuvector/neuvector/tags We have to review which `tags` are the one to be released ( for example, only `v[0-9]+.[0-9]+.[0-9]+` avoiding `-p*` or `-rc*` @williamlin-suse 

Background: https://github.com/rancher-eio/internal-tickets/issues/6

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,9 +45,22 @@ jobs:
           secret/data/github/repo/${{ github.repository }}/rancher-prime-registry/credentials username | PRIME_REGISTRY_USERNAME ;
           secret/data/github/repo/${{ github.repository }}/rancher-prime-registry/credentials password | PRIME_REGISTRY_PASSWORD
     - name: Parse target tag
+      id: parse_tag_step
       run: |
         TARGET=${{ github.ref_name }}
         echo "TAG=${TARGET#v}" >> $GITHUB_ENV
+
+        PARSED_TAG="" # Initialize as empty
+
+        if [[ "$TARGET" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+          PARSED_TAG="${TARGET#v}"
+          echo "Tag '$PARSED_TAG' successfully validated as '$PARSED_TAG'."
+        else
+          echo "Tag '$PARSED_TAG' does NOT match the required 'vX.Y.Z' format. Skipping."
+        fi
+
+        echo "tag_value=$PARSED_TAG" >> $GITHUB_OUTPUT
+
     - name: Get controller
       env:
         GH_TOKEN: ${{ secrets.ACCESS_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -93,3 +106,54 @@ jobs:
         prime-repo: rancher
         prime-username: ${{ env.PRIME_REGISTRY_USERNAME }}
         prime-password: ${{ env.PRIME_REGISTRY_PASSWORD }}
+    outputs:
+      parsed_tag: ${{ steps.parse_tag_step.outputs.tag_value }}
+
+  update-upgrade-responder:
```
holyspectral, line 112: Like mentioned in DMs, have you considered to make the code into a shared GHA, like `rancher-eio/read-vault-secrets`?  This way you have more control about the action and other products can benefit from it too.

.github/workflows/release.yml
```diff
@@ -45,9 +45,22 @@ jobs:
...
+
+        if [[ "$TARGET" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+          PARSED_TAG="${TARGET#v}"
+          echo "Tag '$PARSED_TAG' successfully validated as '$PARSED_TAG'."
+        else
+          echo "Tag '$PARSED_TAG' does NOT match the required 'vX.Y.Z' format. Skipping."
+        fi
+
+        echo "tag_value=$PARSED_TAG" >> $GITHUB_OUTPUT
+
```
holyspectral, line 63: It's actually fine with me if it creates a PR for all kinds of tags, including RC or patch releases.  We can still decide whether to merge the PR or not, but this version is good too.

.github/workflows/release.yml
```diff
@@ -93,3 +106,54 @@ jobs:
...
+  update-upgrade-responder:
+      needs: publish
+      runs-on: ubuntu-latest
+      permissions:
+        contents: write
+        pull-requests: write
+      env:
+        TAG_FROM_PUBLISH: ${{ needs.publish.outputs.parsed_tag }}
+        COMMIT_MESSAGE: "Automated commit by GitHub Actions"
```
holyspectral: There is no strong requirement in this repo, but generally it's a good idea to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) and provide more detail, e.g., `chore: update upgrade-responder.json`

.github/workflows/release.yml
```diff
@@ -93,3 +106,54 @@ jobs:
...
+        
+        - name: Creating upgrade-responder file
+          run: |
+            python -c '
+            import json, os
+            from datetime import datetime, timezone
+            tag_value = os.environ.get("TAG_FROM_PUBLISH", "unknown")
+            current_date = datetime.now(timezone.utc).isoformat(timespec="milliseconds")
+            data = {"Versions": [{"Name": f"v{tag_value}", "ReleaseDate": current_date, "Tags": ["latest", tag_value]}]}
+            with open(".github/upgrade-responder.json", "w") as f: json.dump(data, f, indent=2); f.write("\n")
```
holyspectral, line 140: The logic here will overwrite the previous upgrade-responder.json.

```
$ cat .github/upgrade-responder.json 
{
  "Versions": [
    {
      "Name": "v1.2.3",
      "ReleaseDate": "2025-07-10T21:23:35.639+00:00",
      "Tags": [
        "latest",
        "1.2.3"
      ]
    }
  ]
}
$ TAG_FROM_PUBLISH=4.5.6 python3 ./1.py 
$ cat .github/upgrade-responder.json 
{
  "Versions": [
    {
      "Name": "v4.5.6",
      "ReleaseDate": "2025-07-10T21:23:44.520+00:00",
      "Tags": [
        "latest",
        "4.5.6"
      ]
    }
  ]
}
```
I'm not 100% sure about the upgrade-responder's format.  Could you confirm if this is the expected behavior?

holyspectral (commented):
I think generally it looks good to me.  Thanks for working on it!  I will leave it open for a little longer to gather feedback from others.  Just a little comments.

.github/workflows/release.yml
```diff
@@ -3,7 +3,7 @@ name: Release
 on:
   push:
     tags:
-      - 'v*'
+      - 'v*.*.*'
```
pjbgf: I don't think this is particularly strict. I'd rather leave as it was or go for something that is quite specific (e.g. `'v[0-9]+.[0-9]+.[0-9]+'`).

.github/workflows/release.yml
pjbgf: +1, I'm not sure about the format. How many times we need to keep there, etc. It would be nice to have a comment somewhere to either call it out or a link to refer back to.

pjbgf (commented):
Overall the approach LGTM.

.github/workflows/release.yml
holyspectral: @Anarkis i saw you’ve implemented this suggestion in your latest version.  However, while whether rc builds should go through release.yml is arguable, we do need the suffix to be allowed.  This way we can build patch release like 5.5.5-p1.

Could you please revert this part back to your original version?

@pjbgf please feel free to let us know if you have different thoughts.

.github/workflows/release.yml
holyspectral: @Anarkis could you please confirm if the behavior here is as expected?

PR #2054, opened 2025-07-26
NVSHAS-10018: Neuvector is not scanning all images in GitLab Registry

williamlin-suse:


controller/scan/gitlab.go
```diff
@@ -110,118 +112,234 @@ func (r *gitlab) Login(cfg *share.CLUSRegistryConfig) (error, string) {
...
+			}
+		}
+		if nextURL == noNextPageURL {
+			if len(all) > 0 {
+				err = nil
+			}
+			return all, err
+		}
+		ur = nextURL
 	}
```
pohanhuangtw, line 177: It looks like we’re duplicating the same pagination+unmarshal loop in multiple places. Could we extract that into a single, generic helper (e.g. FetchAllPaginated[T]) and call it from each of these methods?
```go
func (c *Client) FetchAllPaginated(startURL string) ([]T, error) {
    var (
        all     []T
        nextURL = startURL
    )

    for {
        c.log.WithFields(log.Fields{"url": nextURL}).Debug("fetching page")
        raw, next, err := c.getData(nextURL)
        if err != nil {
            return nil, err
        }

        var page []T
        if err := json.Unmarshal(raw, &page); err != nil {
            return nil, err
        }

        if len(page) == 0 {
            break
        }
        all = append(all, page...)

        if next == noNextPageURL {
            break
        }
        nextURL = next
    }

    return all, nil
}
```**

File: controller/scan/gitlab.go [hunk 2]
```diff
@@ -110,118 +112,234 @@ func (r *gitlab) Login(cfg *share.CLUSRegistryConfig) (error, string) {
...
+				}
+			}
+		}
+		if nextURL == noNextPageURL {
+			if len(all) > 0 {
+				err = nil
+			}
+			return all, err
+		}
+		ur = nextURL
```
pohanhuangtw: same here

controller/scan/gitlab.go
```diff
@@ -110,118 +112,234 @@ func (r *gitlab) Login(cfg *share.CLUSRegistryConfig) (error, string) {
...
+				}
+			}
+		}
+		if nextURL == noNextPageURL {
+			if len(all) == 0 && err != nil {
+				e1 = err
+			}
+			break
+		}
+		ur = nextURL
```
pohanhuangtw, line 212: same here

controller/scan/gitlab.go
```diff
@@ -110,118 +112,234 @@ func (r *gitlab) Login(cfg *share.CLUSRegistryConfig) (error, string) {
...
+						}
+					}
 				}
+				if nextURL == noNextPageURL {
+					if len(all) == 0 && err != nil {
+						e2 = err
+					}
+					break
+				}
+				ur = nextURL
```
pohanhuangtw: same here

controller/scan/gitlab.go
```diff
@@ -110,118 +112,234 @@ func (r *gitlab) Login(cfg *share.CLUSRegistryConfig) (error, string) {
...
-	} else {
-		e2 = err
+		if nextURL == noNextPageURL {
+			if len(all) == 0 && err != nil {
+				e2 = err
+			}
+			break
+		}
+		ur = nextURL
 	}
```
pohanhuangtw, line 242: same here

controller/scan/gitlab.go
```diff
@@ -110,118 +112,234 @@ func (r *gitlab) Login(cfg *share.CLUSRegistryConfig) (error, string) {
...
+						}
+					}
+				}
+				if nextURL == noNextPageURL {
+					if len(all) == 0 && err != nil {
+						e3 = err
+					}
+					break
 				}
+				ur = nextURL
```
pohanhuangtw: same here

controller/scan/gitlab.go
```diff
@@ -232,35 +350,57 @@ func (r *gitlab) getProjects() ([]gitProject, error) {
...
+				}
+			}
+		}
+		if nextURL == noNextPageURL {
+			if len(all) > 0 {
+				err = nil
+			}
+			return all, err
+		}
+		ur = nextURL
```
pohanhuangtw: same here

controller/scan/gitlab.go
```diff
@@ -232,35 +350,57 @@ func (r *gitlab) getProjects() ([]gitProject, error) {
...
```
pohanhuangtw: same here

controller/scan/gitlab.go
```diff
@@ -110,118 +112,234 @@ func (r *gitlab) Login(cfg *share.CLUSRegistryConfig) (error, string) {
...
 	}
-
-	return all, nil
 }
 
 func (r *gitlab) getProjects() ([]gitProject, error) {
 	var all []gitProject
-	var projects []gitProject
 	var e1, e2, e3 error
```
pohanhuangtw: Let’s rename the error variables for clarity:
```suggestion
	var projectErr, userProjectErr, groupProjectErr error
```

controller/scan/gitlab.go
```diff
@@ -110,118 +112,234 @@ func (r *gitlab) Login(cfg *share.CLUSRegistryConfig) (error, string) {
...
+				relStr := strings.TrimSpace(ss[1])
+				if relStr == `rel="next"` {
+					nextURL = strings.TrimSpace(ss[0])
+					nextURL = strings.Trim(nextURL, "<>")
+					break
+				}
+			}
+		}
+	}
+	if len(data) == 0 {
```
pohanhuangtw, line 148: It looks like we’re duplicating the Link-header pagination logic here. Could we extract this into a helper function (for example, parseNextPageURL(resp *http.Response) string) to improve readability and avoid repetition?

pohanhuangtw (changes requested):
I’ve left a few review comments—could you please take a look and let me know your thoughts?

controller/scan/gitlab.go
williamlin-suse: It's actually a bug. The duplicate parsing code is removed.

controller/scan/gitlab.go
williamlin-suse: use go generics instead

controller/scan/gitlab.go
williamlin-suse: add a common function for projects

controller/scan/gitlab.go
williamlin-suse: use go generics instead

controller/scan/gitlab.go
williamlin-suse: use go generics instead

controller/scan/gitlab.go
williamlin-suse: use common function instead

controller/scan/gitlab.go
williamlin-suse: use go generics instead

controller/scan/gitlab.go
williamlin-suse: use go generics instead

controller/scan/gitlab.go
williamlin-suse: error names changed

controller/scan/gitlab.go
williamlin-suse: project's logic is a little different from others. so go generics is not used for projects to maintain the original logic.

controller/scan/gitlab.go
```diff
@@ -110,118 +112,158 @@ func (r *gitlab) Login(cfg *share.CLUSRegistryConfig) (error, string) {
...
-		return nil, err
-	}
-	err = json.Unmarshal(data, &all)
-	if err != nil {
-		return nil, err
+	return getResources[gitGroup](r, ur)
+}
+
+func (r *gitlab) getProjectResources(ur string, all []gitProject) ([]gitProject, error) {
+	var e error
```
pohanhuangtw:
```suggestion
	var getProjResourceErr error
```

pohanhuangtw (changes requested):
Overall LGTM. Just a small nit

controller/scan/gitlab.go
williamlin-suse: code changed

controller/scan/gitlab.go
```diff
@@ -110,118 +112,158 @@ func (r *gitlab) Login(cfg *share.CLUSRegistryConfig) (error, string) {
...
+					relStr := strings.TrimSpace(ss[1])
+					if relStr == `rel="next"` {
+						nextURL = strings.TrimSpace(ss[0])
+						nextURL = strings.Trim(nextURL, "<>")
+						break outerLoop
+					}
+				}
+			}
+		}
 	}
```
pohanhuangtw, line 147: Hi @williamlin-suse ,
After some thought, I think we should extract the following, and check if the return is noNextPageURL, the keep the loop, what do you think?

```go
			for _, pair := range strings.Split(links, ",") {
				pair = strings.TrimSpace(pair)
				if ss := strings.Split(pair, ";"); len(ss) == 2 {
					relStr := strings.TrimSpace(ss[1])
					if relStr == `rel="next"` {
						nextURL = strings.TrimSpace(ss[0])
						nextURL = strings.Trim(nextURL, "<>")
						break outerLoop
					}
				}
			}
```

williamlin-suse: adding noNextPageURL check around `break outerLoop` should be enough

pohanhuangtw (approved):

PR #2070, opened 2025-08-11
fix: NVSHAS-9981 update zero-drift behavior

holyspectral:
Reworked based on https://github.com/neuvector/neuvector/commit/06eefb0aa33e6942ccb57aaeab774847c6fe83d7

### DIFF
--- a/agent/probe/process.go
+++ b/agent/probe/process.go
@@ -2515,6 +2515,15 @@ func (p *Probe) procProfileEval(id string, proc *procInternal, bKeepAlive bool) 
 				log.WithFields(log.Fields{"name": proc.name, "pid": proc.pid}).Debug("PROC: Denied and killed")
 			}
 		}
+	} else {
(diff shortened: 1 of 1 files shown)


holyspectral:
@gfsuse Yeah so in the original Jay's version, https://github.com/neuvector/neuvector/commit/06eefb0aa33e6942ccb57aaeab774847c6fe83d7, enforcer will override a process's verdict with its parent's verdict regardless it's learning mode or monitoring mode.

However, in monitoring mode, at least before 5.4.2, users were actually allowed to write an allow rule to prevent a process from being reported again. This can be done in UI too. 

Because Jay's change unconditionally use the parent's verdict, the result from process profiles will not be used, which essentially makes the feature broken, so we can't let the logic work in monitor mode.

gfsuse (approved):
LGTM after discussion.

PR #2071, opened 2025-08-12
NVSHAS-10043: Improve link and address subscription with options

kyledong-suse:


agent/engine.go
```diff
@@ -2540,15 +2540,29 @@ func intfHostMonitorLoop(hid string, stopCh chan struct{}) {
 	doneLink := make(chan struct{})
 	defer close(doneLink)
 
-	if err = netlink.LinkSubscribe(chLink, doneLink); err != nil {
+	linkOptions := netlink.LinkSubscribeOptions{
+		ListExisting: false,
```
kyledong-suse, line 2544: Set the `ListExisting` to `false` means
- Prevents the initial dump of all existing links when the subscription starts
- Won't affect ongoing change detection for existing links

PR #2076, opened 2025-08-16
NVSHAS-10054: Do not reset k8s ValidatingWebhookConfiguration to avoid infinite updating loops

williamlin-suse:
…d infinite updating loops


gfsuse (commented):
In reviewing following functions:
refreshK8sAdminWebhookStateCache->SyncAdmCtrlStateToK8s->ConfigK8sAdmissionControl
I noticed we already have a logic to prevent update loop
			if op == K8sResOpUpdate && k8sResInfo.RevertCount != nil && (k8sResInfo.UnexpectedMatchExpr != "" && *k8sResInfo.RevertCount > 0) {
				return true, nil
			} else {
Is the third-party update not reflected in k8sResInfo.UnexpectedMatchExpr?

PR #2079, opened 2025-08-20
NVSHAS-10060: Response rules for the target CRD group are not deleted

williamlin-suse:


gfsuse (approved):
The logic and code looks good.

PR #2084, opened 2025-08-25
NVSHAS-9968-9990: support setting default admin account's default password

williamlin-suse:


BinX-Suse:
Merged since @holyspectral is on PTO.

PR #2087, opened 2025-08-25
NVSHAS-10063: namespace user with runtime_policy(w) permission cannot…

williamlin-suse:
… create response rule


kyledong-suse (approved):
LGTM

PR #2095, opened 2025-09-03
NVSHAS-10073: Add a new field for exported admission control rules in…

williamlin-suse:
… the yaml file


pohanhuangtw (commented):
Overall LGTM, one minor concern.
Should we update the api doc?

williamlin-suse:
> Overall LGTM, one minor concern. Should we update the api doc?

No need to update apis.yaml
The type NvSecurityAdmCtrlRule is an internal type and RESTAdmCtrlRulesExport is not changed in this PR

holyspectral (approved):
I do feel that it would be beneficial to document `NvSecurityAdmCtrlRule` but either way this doesn't block this PR.  Approving it now.

PR #2105, opened 2025-09-13
fix: admission converter dependency

pohanhuangtw:
### Summary
- move export admission control struct to api package.


williamlin-suse (approved):
LGTM

PR #2107, opened 2025-09-18
NVSHAS-9918: add option for strict group mode

gfsuse:


controller/cache/policy.go
```diff
@@ -925,7 +925,91 @@ func getServiceAddrForMode(mode string) []*share.CLUSWorkloadAddr {
...
+				PolicyMode: share.PolicyModeEvaluate,
+				LocalPortApp: []share.CLUSPortApp{
+					{
+						Ports:       "any",
+						Application: C.DP_POLICY_APP_ANY,
+					},
+				},
+			})
+		printOneGroupIPPolicy(&policy)
+		policyList = append(policyList, policy)
```
kyledong-suse: nit: maybe this part can be moved to a helper function to reduce the redundant as the following checks will run the same logic.

controller/cache/policy.go
```diff
@@ -980,7 +1064,126 @@ func getMixedGroupPolicyForIngress() []share.CLUSGroupIPPolicy {
...
+					},
+				},
+			})
+
+		srv := getServiceAddrForMode(share.PolicyModeLearn)
+		policy.To = append(policy.To, srv...)
+		if len(policy.To) != 0 {
+			policyList = append(policyList, policy)
+		}
+		printOneGroupIPPolicy(&policy)
```
kyledong-suse: nit: maybe this part can be moved to a helper function to reduce the redundant as the following checks will run the same logic.

controller/cache/policy.go
gfsuse: ok

controller/cache/policy.go
gfsuse: ok

PR #2108, opened 2025-09-19
NVSHAS-10079: fix goroutine crash in system.go

kyledong-suse:


agent/system.go
```diff
@@ -207,6 +209,8 @@ func initWorkloadPolicyMap() map[string]*policy.WorkloadIPPolicyInfo {
 	policyVerVal++
 	policyVer := uint16(policyVerVal % polVerMax)
 	workloadPolicyMap := make(map[string]*policy.WorkloadIPPolicyInfo)
+
+	gInfoRLock()
```
holyspectral: You can use `defer`.
```suggestion

	gInfoRLock()
	defer gInfoRUnlock()
```

holyspectral (commented):
It might be a little tricky in the `NotifyPolicyChange()` case because it holds another lock, but I think you can use `defer()` in other ~two~ three cases.

agent/system.go
kyledong-suse: good suggestion

kyledong-suse:
Yes, I think for `NotifyPolicyChange` just lock the specific part of iterating `gInfo.activeContainers` will be more safe.

holyspectral (approved):

agent/system.go
```diff
@@ -111,6 +111,9 @@ func systemConfigPolicyMode(mode string) {
 func systemConfigTapProxymesh(tapProxymesh bool) {
 	//proxy mesh status is changed
 	gInfo.tapProxymesh = tapProxymesh
+	gInfoRLock()
+	defer gInfoRUnlock()
+
 	for _, c := range gInfo.activeContainers {
 		if tapProxymesh {
 			//enable proxy mesh
```
gfsuse, line 119: inside function enableTapProxymesh, we need to pass down a correct with(out)lock value to updateProxyMeshMac
func enableTapProxymesh(c *containerData) {
	updateProxyMeshMac(c, false)
func updateProxyMeshMac(c *containerData, withlock bool) {

gfsuse (changes requested):

agent/system.go
kyledong-suse: Thanks for catching it. I have added the `withlock` flag in `enableTapProxymesh()` to prevent the nest lock scenario.

PR #2119, opened 2025-10-09
NVSHAS-10049: Follow up on the NV scan JFrog Subdomain mode issue

williamlin-suse:
Solve the issue when the jfrog(subdomain mode) registry URL starts with tweaked-subdomain in the hostname.
Older NV versions worked well on this special env before NV supports newer versions of JFrog.
However, after NV started supporting newer versions of JFrog, NV doesn't work on this special env anymore.

### DIFF
--- a/controller/rest/registry.go
+++ b/controller/rest/registry.go
@@ -105,11 +105,18 @@ func parseWildcardRegex(s string) (string, error) {
 	return out, nil
 }
 
-func parseFilter(filters []string, regType string) ([]*share.CLUSRegistryFilter, error) {
+func isJFrogSubdomainPlaceholder(regType, jfrogMode, org string) bool {
+	return regType == share.RegistryTypeJFrog && jfrogMode == share.JFrogModeSubdomain &&
+		strings.HasPrefix(org, "<") && strings.HasSuffix(org, ">")
+}
+
+func parseFilter(filters []string, config share.CLUSRegistryConfig) ([]*share.CLUSRegistryFilter, error) {
 	if len(filters) == 0 {
 		return make([]*share.CLUSRegistryFilter, 0), nil
 	}
 
+	regType := config.Type
+	jfrogMode := config.JfrogMode
 	repoFilters := make([]*share.CLUSRegistryFilter, len(filters))
 
 	for n, filter := range filters {
@@ -137,9 +144,22 @@ func parseFilter(filters []string, regType string) ([]*share.CLUSRegistryFilter,
 		}
 
 		// org
-		if (org != "" && !orgRegexp.MatchString(org)) || (org == "" && filter != "*" && regType == share.RegistryTypeOpenShift) {
-			log.WithFields(log.Fields{"org": org, "type": regType}).Error("Failed to parse organization in the filter")
-			return nil, errors.New("Invalid filter format")
+		// Characters '<' & '>' are invalid for org in the register filter configuration
+		// However, on the special case where the registry URL contains "{tweaked_subdomain}." , but not "{subdomain}." , as the prefix of hostname,
+		//  NV cannot tell what's the real {subdomain} value is because "{tweaked_subdomain}" could be totally irrevalent from "{subdomain}" literally.
+		// To solve this issue, characters '<' & '>' are allowed for org in the register filter configuration of JFrog registry server/subdomain mode.
+		// If user's JFrog registry URL must contain "{tweaked_subdomain}." as the prefix of hostname, user can spceify the real {subdomain} value by
+		//  configuring registry filter like "<{subdomain}>/{repo}"
+		// This special prefix "<{subdomain}>/" in the registry filter tells NV what the real {subdomain} value is for JFrog registry server/subdomain mode.
+		invalidOrg := false
+		if org != "" && !orgRegexp.MatchString(org) {
+			invalidOrg = true
+		}
+		if invalidOrg || (org == "" && filter != "*" && regType == share.RegistryTypeOpenShift) {
+			if !isJFrogSubdomainPlaceholder(regType, jfrogMode, org) {
+				log.WithFields(log.Fields{"org": org, "type": regType}).Error("Failed to parse organization in the filter")
+				return nil, errors.New("Invalid filter format")
+			}
 		}
 
 		// repo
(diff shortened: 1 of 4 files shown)


controller/rest/registry.go
```diff
@@ -137,9 +137,16 @@ func parseFilter(filters []string, regType string) ([]*share.CLUSRegistryFilter,
...
-		if (org != "" && !orgRegexp.MatchString(org)) || (org == "" && filter != "*" && regType == share.RegistryTypeOpenShift) {
-			log.WithFields(log.Fields{"org": org, "type": regType}).Error("Failed to parse organization in the filter")
-			return nil, errors.New("Invalid filter format")
+		invalidOrg := false
+		if org != "" && !orgRegexp.MatchString(org) {
+			invalidOrg = true
+		}
+		if invalidOrg || (org == "" && filter != "*" && regType == share.RegistryTypeOpenShift) {
+			if !(regType == share.RegistryTypeJFrog && mode == share.JFrogModeSubdomain &&
+				strings.HasPrefix(org, "<") && strings.HasSuffix(org, ">")) {
```
pohanhuangtw: Maybe create a new function for better readablity

```go
func isJFrogSubdomainPlaceholder(regType, mode, org string) bool {
    return regType == share.RegistryTypeJFrog &&
           mode == share.JFrogModeSubdomain &&
           strings.HasPrefix(org, "<") &&
           strings.HasSuffix(org, ">")
}
```

controller/rest/registry.go
```diff
@@ -105,7 +105,7 @@ func parseWildcardRegex(s string) (string, error) {
 	return out, nil
 }
 
-func parseFilter(filters []string, regType string) ([]*share.CLUSRegistryFilter, error) {
+func parseFilter(filters []string, regType, mode string) ([]*share.CLUSRegistryFilter, error) {
```
pohanhuangtw: -`mode` is JFrog-specific. Renaming to `jfrogMode` would make the scope clearer.
- we can even pass the config into this parseFilter
```suggestion
func parseFilter(filters []string, string, config share.CLUSRegistryConfig) ([]*share.CLUSRegistryFilter, error) {

regType := config.type
jfrogMode := config.JfrogMode
```

controller/scan/jfrog.go
```diff
@@ -78,7 +78,7 @@ func (r *jfrog) Login(cfg *share.CLUSRegistryConfig) (error, string) {
 	return nil, ""
 }
 
-func (r *jfrog) getSubdomainRepoList(jdirs []jfrogDir, org, name string, limit int) ([]*share.CLUSImage, error) {
+func (r *jfrog) getSubdomainRepoList(jdirs []jfrogDir, org, name, keyToUse string, limit int) ([]*share.CLUSImage, error) {
```
pohanhuangtw: how about `repositoryKey` instead of `keyToUse`?

controller/scan/jfrog.go
```diff
@@ -97,13 +98,19 @@ func (r *jfrog) getSubdomainRepoList(jdirs []jfrogDir, org, name string, limit i
...
-			if org != "" && org != dir.Key {
-				continue
+			if r.subdomainInRegURL != "" {
+				if r.subdomainInRegURL != dir.Key {
+					continue
+				}
+			} else {
+				if org != "" && org != dir.Key {
+					continue
+				}
```
pohanhuangtw:
```suggestion
func (r *jfrog) shouldSkipDir(dir jfrogDir, org, keyToUse string) bool {
    if keyToUse != "" {
        return dir.Key != keyToUse
    }
    
    if r.subdomainInRegURL != "" {
        return dir.Key != r.subdomainInRegURL
    }
    
    return org != "" && dir.Key != org
}

```

To simplify the logic

controller/scan/jfrog.go
```diff
@@ -166,7 +173,20 @@ func (r *jfrog) GetRepoList(org, name string, limit int) ([]*share.CLUSImage, er
 		return r.base.GetRepoList(org, name, limit)
 	}
 
-	jdirs, err := r.getJFrogDirUrl()
+	var keyToUse string
```
pohanhuangtw: how about `repositoryKey` instead of `keyToUse`?

controller/scan/jfrog.go
```diff
@@ -178,39 +198,52 @@ func (r *jfrog) GetRepoList(org, name string, limit int) ([]*share.CLUSImage, er
...
 				// registry filter could contain subdomain as prefix or not
 				r.subdomainInRegURL = getSubdomainFromRegURL(r.base.regURL)
 				repoPrefixToRevert := ""
-				if r.subdomainInRegURL != "" && org != "" && org != r.subdomainInRegURL {
-					// in 5.4.4(-), if registry filter is not "*", it must have prefix "{subdomain}/"
-					// so we need to make sure filter has prefix "{subdomain}/" before calling r.getSubdomainRepoList()
-					name = fmt.Sprintf("%s/%s", org, name)
-					repoPrefixToRevert = fmt.Sprintf("%s/", r.subdomainInRegURL)
-					org = r.subdomainInRegURL
+				if r.subdomainInRegURL != "" {
```
pohanhuangtw, line 204: Could we extract lines 216–241 into a helper function and add comments explaining the parsing cases and rationale?

pohanhuangtw (changes requested):
I’ve left some comments. With minor naming tweaks and a small function extraction, overall LGTM.

controller/scan/jfrog.go
williamlin-suse: prefer not to use a helper function for this piece of code because some local variables are changed & affect later logic

controller/rest/registry.go
```diff
@@ -137,9 +144,15 @@ func parseFilter(filters []string, regType string) ([]*share.CLUSRegistryFilter,
...
 
 		// org
-		if (org != "" && !orgRegexp.MatchString(org)) || (org == "" && filter != "*" && regType == share.RegistryTypeOpenShift) {
-			log.WithFields(log.Fields{"org": org, "type": regType}).Error("Failed to parse organization in the filter")
-			return nil, errors.New("Invalid filter format")
+		invalidOrg := false
+		if org != "" && !orgRegexp.MatchString(org) {
+			invalidOrg = true
+		}
+		if invalidOrg || (org == "" && filter != "*" && regType == share.RegistryTypeOpenShift) {
```
pohanhuangtw, line 158: nit: Could you add a brief comment here explaining why we’re doing this?

pohanhuangtw (changes requested):

pohanhuangtw (approved):
LGTM, thanks

holyspectral (approved):

PR #2129, opened 2025-10-22
NVSHAS-10061: move scan credit storage into consul

pohanhuangtw:
## Problem

When scanner autoscaling is enabled, controllers fail to distribute load across all scanners, causing "RPC request fail" errors (NVSHAS-10061). Controllers can connect to zombie scanners or have stale scanner information.

## Root Cause

- **Stale scanner state**: Controllers don't share up-to-date scanner information
- **Zombie scanner connections**: Controllers maintain connections to removed scanners
- **Unfair load distribution**: High concurrent loads amplify the timing window for stale state

### 1. Centralized Scanner State (Consul)
- Persist scan-credit state in Consul
- Controllers read fresh scanner info from cluster instead of local cache
- Eliminates zombie scanner connections

### 2. Fair Scanner Acquisition
- Implement pull-time ordered priority queue (FIFO)
- Prevent request starvation under high concurrency
- Exponential backoff with jitter for failed acquisitions

### 3. Safe Scanner Acquisition Operations
- Use CAS (Compare-And-Swap) for scanner acquisition
- Retry loop (max 10 times) for credit release to handle transient failures

## Testing

- **Concurrent load**: 1000 parallel image scans
- **Autoscaling**: min 3, max 8 scanner pods
- **Infrastructure**: Latest Jenkins build

## Verification

1. Enable scanner autoscaling (min: 3, max: 8)
2. Trigger 1000 parallel repository scans
3. Monitor RPC errors during scale-up
4. Verify all scans complete

### Expected Results

✅ No "RPC request fail" errors during scanner initialization  
✅ Smooth scale-up without dropped requests  
✅ All 1000 scans complete successfully

### DIFF
--- a/controller/rpc/scanner.go
+++ b/controller/rpc/scanner.go
@@ -134,6 +134,7 @@ func (mgr *ScannerAcquisitionManager) acquireScanner(ctx context.Context) (strin
 	backoff.InitialInterval = BaseDelay
 	backoff.MaxInterval = MaxDelay
 	backoff.RandomizationFactor = JitterFactor
+	backoff.Reset()
 
 	req := &acquireScannerRequest{
 		pullTime:     time.Now(),


controller/rpc/scanner.go
```diff
@@ -2,82 +2,93 @@ package rpc
...
-		scannerLoadBalancer: scannerlb.NewScannerLoadBalancer(),
-		maxConns:            maxConns,
-		// scanLBMax*(maxConns+1) ensures that the creditPool channel has sufficient capacity to handle task signals without blocking.
-		creditPool: make(chan struct{}, scannerLBMax*(maxConns+1)),
-	}
-}
+var (
+	ScanCreditMgr         *ScanCreditManager
+	ErrNoScannerAvailable = errors.New("no scanner available")
+	ErrNoScannerFound     = errors.New("no scanner found")
```
williamlin-suse: Aren't ErrNoScannerFound/ErrNoScannerAvailable already defined in share/types.go ?

share/clus_apis.go
```diff
@@ -1941,6 +1941,10 @@ type CLUSScanner struct {
 	RPCServerPort   uint16    `json:"rpc_port"`
 	BuiltIn         bool      `json:"builtin"`
 	CVEDBEntries    int       `json:"db_entries"`
+	// ScanCredit represents the number of scan tasks assigned to this scanner
```
williamlin-suse: This comment is confusing. Which one is more accurate?
1. // ScanCredit represents the number of scan tasks assigned to this scanner
2. // ScanCredit represents the number of scan tasks that can still be assigned to this scanner

monitor/monitor.c
```diff
@@ -80,7 +80,7 @@
 #define ENV_SCANNER_CTRL_PASS   "SCANNER_CTRL_API_PASSWORD"
 
 #define ENV_MAX_SCANNER_TASKS                      "MAX_SCANNER_TASKS"
-#define ENV_SCANNER_LB_MAX                         "SCANNER_LB_MAX"
+#define ENV_MAX_SCANNERS_PER_CONTROLLER            "MAX_SCANNERS_PER_CONTROLLER"
```
williamlin-suse: The var name is confusing.
It's actually for maxExpectedScanners which has nothing to do with "per controller", correct?

controller/rpc/scanner.go
```diff
@@ -92,39 +103,35 @@ func (mgr *ScanCreditManager) signalScanCredit() {
...
-	mgr.mutex.Unlock()
-
+	// Ensure the grpc client is deleted
+	scanner, _, err := mgr.clusterHelper.GetScannerRev(scannerID)
 	if err != nil {
 		log.WithFields(log.Fields{"error": err, "scanner": scannerID}).Error()
 		return err
 	}
+	endpoint := mgr.getScannerEndpoint(scanner)
+	cluster.DeleteGRPCClient(endpoint)
```
williamlin-suse, line 183: After `rpc.ScanCreditMgr.RemoveScanner(sid) -> mgr.clusterHelper.DeleteScanner(scannerID)` is called by the 1st controller, the kv key for that scanner is deleted.
Then how is `cluster.DeleteGRPCClient(endpoint)` called in the 2nd/3rd controllers.
Is there a leak?
The same question also applies to draining availableSlots tokens from creditPool to maintain balance(lines 127~136).
The 2nd/3rd controller won't drain availableSlots tokens from creditPool after the 1st controller deleted the scanner's kv key, correct?

share/clus_apis.go
pohanhuangtw: The second one, thanks
Will use the second one!

controller/rpc/scanner.go
pohanhuangtw: Yes, I should remove this, thanks

controller/rpc/scanner.go
pohanhuangtw: yes, you are right, in this case the controller 2/3 will not able to delete the endpoint and availableSlots, will fix it in next commit

pohanhuangtw:
Hi @williamlin-suse ,
In the latest change I made the following changes
- use remove scanner with info to delete the remaing credit for those controller does not delete scanner in HA, which address https://github.com/neuvector/neuvector/pull/2129#discussion_r2467574750
- update naming in monitor.c, address https://github.com/neuvector/neuvector/pull/2129#discussion_r2467472388
- remove redundant var, address https://github.com/neuvector/neuvector/pull/2129#discussion_r2467282386
- improve comment, address https://github.com/neuvector/neuvector/pull/2129#discussion_r2467459471

controller/cache/cache.go
```diff
@@ -1767,20 +1767,24 @@ func startWorkerThread(ctx *Context) {
 									cache.errCount++
 									if cache.errCount >= scannerClearnupErrorMax {
 										log.WithFields(log.Fields{"scanner": sid}).Info("Remove stalled internal scanner")
-										_ = clusHelper.DeleteScanner(sid)
+										// Remove from cache first to avoid double cleanup when receiving our own delete notification
+										RemoveScannerFromCache(sid)
```
williamlin-suse: In `func RemoveScannerFromCache(scannerID string)` it tries to grab cache lock for write by calling `cacheMutexLock()`.
But in line 1758 the cache lock is already owned for read before it reaches line 1771.
Isn't this a deadlock?

controller/cache/cache.go
```diff
@@ -1767,20 +1767,24 @@ func startWorkerThread(ctx *Context) {
...
 							// Ping external scanner, remove if no response after 3 tries.
-							if err := rpc.Ping(sid, scannerClearnupTimeout); err != nil {
+							if err := rpc.ScanCreditMgr.Ping(sid, scannerClearnupTimeout); err != nil {
 								log.WithFields(log.Fields{"scanner": sid, "error": err}).Error("Failed to ping scanner")
 								cache.errCount++
 								if cache.errCount >= scannerClearnupErrorMax {
 									log.WithFields(log.Fields{"scanner": sid}).Info("Remove stalled external scanner")
-									_ = clusHelper.DeleteScanner(sid)
+									// Remove from cache first to avoid double cleanup when receiving our own delete notification
+									RemoveScannerFromCache(sid)
```
williamlin-suse: same deadlock issue as line 1771

controller/cache/cache.go
pohanhuangtw: Hi @williamlin-suse ,
The latest commit should resolve the issue, please take a look when you are available, thanks

controller/cache/cache.go
```diff
@@ -1768,7 +1768,7 @@ func startWorkerThread(ctx *Context) {
 									if cache.errCount >= scannerClearnupErrorMax {
 										log.WithFields(log.Fields{"scanner": sid}).Info("Remove stalled internal scanner")
 										// Remove from cache first to avoid double cleanup when receiving our own delete notification
-										RemoveScannerFromCache(sid)
+										delete(scannerCacheMap, sid)
```
williamlin-suse: The cache lock is owned for `read` but it updates `scannerCacheMap` here ...

controller/cache/cache.go
```diff
@@ -1783,7 +1783,7 @@ func startWorkerThread(ctx *Context) {
 								if cache.errCount >= scannerClearnupErrorMax {
 									log.WithFields(log.Fields{"scanner": sid}).Info("Remove stalled external scanner")
 									// Remove from cache first to avoid double cleanup when receiving our own delete notification
-									RemoveScannerFromCache(sid)
+									delete(scannerCacheMap, sid)
```
williamlin-suse: same issue as line 1771

controller/cache/cache.go
pohanhuangtw: fix in next commit

controller/cache/cache.go
pohanhuangtw: fix in next commit

controller/cache/cache.go
pohanhuangtw: fix in next commit

williamlin-suse (approved):

controller/kv/helper.go
```diff
@@ -134,6 +134,10 @@ type ClusterHelper interface {
 	PutScannerStats(id string, objType share.ScanObjectType, result *share.ScanResult) error
 	GetScannerDB(store string) []*share.CLUSScannerDB
 
+	GetAvailableScanners() []*share.CLUSScanner
+	PickLeastLoadedScanner(ScannerHealthCheckFunc func(string, time.Duration) error) (*share.CLUSScanner, error)
+	ReleaseScanCredit(scannerId string) error
+	GetScannerRev(id string) (*share.CLUSScanner, uint64, error) // GetScannerRev gets scanner with revision for internal operations (no auth check)
```
holyspectral: Please add comments about what these new functions do and their expected outcome.

controller/kv/helper.go
```diff
@@ -1582,6 +1602,133 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+	// This is critical because losing credits would cause scanner starvation
+	key := share.CLUSScannerKey(scannerId)
+
+	for {
+		// Get current scanner state with revision
+		value, rev, err := m.get(key)
+		if err != nil || value == nil {
+			// Scanner has been deleted - credit is implicitly released with the scanner
+			// This is the only successful exit condition besides successful Put
+			log.WithFields(log.Fields{"scanner": scannerId}).Debug("Scanner not found during credit release, likely deleted")
```
holyspectral: I think you have to check the error before claiming the scanner is not found.  The `err` might indicate it's a network issue or something like that.  For those kind of issues you probably want to retry where possible.

controller/kv/helper.go
```diff
@@ -1582,6 +1602,133 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+			return nil
+		}
+
+		var scanner share.CLUSScanner
+		if err := nvJsonUnmarshal(key, value, &scanner); err != nil {
+			log.WithFields(log.Fields{"error": err, "scanner": scannerId}).Error("Failed to unmarshal scanner")
+			continue
+		}
```
holyspectral: A mal-formed JSON object or wrong encryption key could lead to an infinite loop here if I'm not mistaken.  We probably shouldn't retry for this case.

controller/kv/helper.go
```diff
@@ -1582,6 +1602,133 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+
+		// Increment credit, ensuring it doesn't exceed max
+		scanner.ScanCredit = min(scanner.ScanCredit+1, scanner.MaxConcurrentScansPerScanner)
+
+		// Attempt to update using CAS (PutRev)
+		value, err = json.Marshal(scanner)
+		if err != nil {
+			log.WithFields(log.Fields{"error": err, "scanner": scannerId}).Error("Failed to marshal scanner")
+			continue
+		}
```
holyspectral, line 1753: Similarly we probably shouldn't retry here.

controller/kv/helper.go
```diff
@@ -1582,6 +1602,133 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+
+		// Use PutRev for atomic compare-and-swap
+		if err := cluster.PutRev(key, value, rev); err != nil {
```
holyspectral, line 1756: ❤️  I think this is exactly what we want when we need data consistency across multiple controllers, but you should only retry retryable errors like `cluster.ErrPutCAS`.

controller/kv/helper.go
```diff
@@ -1582,6 +1602,133 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+			// CAS failed - either revision mismatch or scanner was deleted
+			// Both cases are handled by retrying (scanner deletion will be caught in next iteration)
+			log.WithFields(log.Fields{"error": err, "scanner": scannerId}).Debug("Failed to release scanner credit, retrying")
+			continue
+		}
+		return nil
```
holyspectral, line 1771: At the bare minimum I think we should implement a max retry number here.

controller/kv/helper.go
pohanhuangtw: I think we can simply check if the error is not found and retry otherwise.
With a maximum retry limit, this should save us from overcomplicating the logic or worrying too much about transient errors.

```go
		if err != nil {
			// Check if scanner was deleted (object not found)
			if errors.Is(err, common.ErrObjectNotFound) {
				// Scanner has been deleted - credit is implicitly released with the scanner
				log.WithFields(log.Fields{"scanner": scannerId}).Debug("Scanner not found during credit release, likely deleted")
				return nil
			}

			// Network error or other temporary failure - retry with backoff
			log.WithFields(log.Fields{"error": err, "scanner": scannerId}).Error("Failed to get scanner, retrying")
			continue
		}

```
What do you think?

pohanhuangtw:
Hi @holyspectral,

Thanks for the detailed feedback! After thinking more about this, I realize the current design has a fundamental trade-off I should address honestly:

**The Core Issue:**
- Each controller maintains its own local credit pool
- With 3 scanners (2 credits each) and 3 controllers: each gets 6 credits locally
- **Total in system: 6 * 3 = 18, but actual capacity is only 6** ← This is the problem

This "oversupply" happens because controllers don't coordinate when picking scanners. We trade consistency for performance (no need to lock on every request).

**Why this is actually okay for our use case:**
- When a controller tries to acquire a scanner, it validates:
  1. Scanner is healthy (ping check)
  2. Scanner has available credits in Consul
- So even though local pools oversupply, **Consul is still the source of truth**
- Worst case: controller gets a false "available" signal locally, but Consul check fails → retry picks another scanner

**The honest trade-off:**
- We accept local inconsistency for performance
- Consul's CAS operations prevent actual credit loss
- Periodic health checks and retries catch stale data

**Why not queue/Semaphore **
- Consul is not design for Queue
- Semaphore require fix number of scanner, which is not our use case.

Does this make sense? I am not avoiding the problem - just tolerating local oversupply because Consul keeps us safe.

Happy to discuss if you think this trade-off isn't acceptable for the performance goals.
cc: @williamlin-suse

pohanhuangtw:
Hi @williamlin-suse @holyspectral ,
In the latest 2 commit, I make the following changes to ensure job will not expire && no need to sync the credit.
- Implement FIFO priority queue for fair allocation
- Use ScanJob-based reconciliation + CAS to prevent credit imbalance

controller/kv/helper.go
```diff
@@ -335,7 +353,7 @@ func newClusterHelper(id, version string, persist bool, keyRotationDuration time
 }
 
 func GetClusterHelper() ClusterHelper {
-	return clusHelperImpl
+	return clusHelper
```
williamlin-suse: This should not be changed to `return clusHelper`
Member fields of var type in `clusHelper` are all empty which breaks other functions

controller/kv/helper.go
```diff
@@ -1582,6 +1616,148 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+			}
+		}
+
+		selectedScanner, rev, err := m.GetScannerRev(scanner.ID)
+		if err != nil {
+			log.WithFields(log.Fields{"error": err, "scanner": selectedScanner.ID}).Error("Failed to get scanner")
+			continue
+		}
+
+		// If the CAS is failed, try the next least loaded scanner
```
williamlin-suse: Is this line of comment correct?
Why `if selectedScanner.ScanCredit <= 0` has anything to do with CAS error?

controller/kv/helper.go
```diff
@@ -1582,6 +1616,148 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+		if ScannerHealthCheckFunc != nil {
+			if err := ScannerHealthCheckFunc(scanner.ID, time.Second*10); err != nil {
+				log.WithFields(log.Fields{"scanner": scanner.ID, "error": err}).Error("Failed to ping scanner")
+				continue
+			}
+		}
+
+		selectedScanner, rev, err := m.GetScannerRev(scanner.ID)
+		if err != nil {
+			log.WithFields(log.Fields{"error": err, "scanner": selectedScanner.ID}).Error("Failed to get scanner")
```
williamlin-suse: When `err != nil`, most likely `selectedScanner` is nil and this logging line causes panic because it tries to reference selectedScanner.
Should it be `"scanner": scanner.ID` instead?

controller/kv/helper.go
```diff
@@ -1582,6 +1616,148 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+	// Infinite retry loop to ensure credit is never lost
+	// This is critical because losing credits would cause scanner starvation
+	key := share.CLUSScannerKey(scannerId)
+	maxRetries := 10
+	for retry := 0; retry < maxRetries; retry++ {
+		// Get current scanner state with revision
+		value, rev, err := m.get(key)
+		if err != nil {
+			// Check if scanner was deleted (object not found)
+			if errors.Is(err, common.ErrObjectNotFound) {
```
williamlin-suse, line 1734: The calling sequence in line 1708 is m.get(key) -> cluster.GetRev(key) -> (m *consulMethod) GetRev(key).
When the key doesn't exist, the returned error is cluster.ErrKeyNotFound, not common.ErrObjectNotFound

Lines 1708~1732 look similar to what `m.GetScannerRev(scannerId)` does.

controller/rest/repository.go
```diff
@@ -141,12 +132,13 @@ func (r *repoScanTask) Run(arg interface{}) (interface{}, *JobError) {
 		// RPC request failed
 		scanErr = NewJobError(api.RESTErrClusterRPCError, err, nil)
 		log.WithFields(log.Fields{
-			"registry": req.Registry, "image": getImageName(req), "error": rsr.errMsg,
+			"registry": req.Registry, "image": getImageName(req), "error": err,
```
williamlin-suse: Could we log both `rsr.errMsg` and `err`?

controller/rest/repository.go
```diff
@@ -141,12 +132,13 @@ func (r *repoScanTask) Run(arg interface{}) (interface{}, *JobError) {
...
 		}).Error("RPC request fail")
 	} else if result.Error != share.ScanErrorCode_ScanErrNone {
-		scanErr = NewJobError(api.RESTErrFailRepoScan, err, scanUtils.ScanErrorToStr(result.Error))
+		// Include the error code in Detail to enable ShouldRetry logic
+		scanErr = NewJobError(api.RESTErrFailRepoScan, err, result.Error)
 		log.WithFields(log.Fields{
-			"registry": req.Registry, "image": getImageName(req), "error": rsr.errMsg,
+			"registry": req.Registry, "image": getImageName(req), "error": scanUtils.ScanErrorToStr(result.Error),
```
williamlin-suse: Could we log both `rsr.errMsg` and `result.Error`?

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
 
-func (mgr *ScanCreditManager) CountScanners() (busy, idle uint32) {
-	mgr.mutex.RLock()
-	defer mgr.mutex.RUnlock()
-
-	for _, s := range mgr.scannerLoadBalancer.GetActiveScanners() {
-		if s.AvailableScanCredits > 0 {
+func (mgr *ScannerAcquisitionManager) CountScanners() (busy, idle uint32) {
+	for _, scanner := range mgr.clusterHelper.GetAvailableScanners() {
+		if scanner.ScanCredit > 0 {
```
williamlin-suse: scanner.ScanCredit represents the number of scan tasks that can still be assigned to this scanner.
Why `scanner.ScanCredit > 0` gets `busy++` & `scanner.ScanCredit <= 0` gets `idle++`?

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
-
-			// Use a heap to find the scanner with the least active tasks
-			scanner, err := mgr.scannerLoadBalancer.PickLeastLoadedScanner()
-			if err != nil {
-				log.WithFields(log.Fields{"error": err}).Error("PickLeastLoadedScanner")
-				return "", err
+		// If no pending requests, wait for new ones from channel (blocking)
+		if len(pendingRequests) == 0 {
+			select {
+			case <-mgr.shutdownCh:
```
williamlin-suse: When is this shutdownCh select triggered? 
I don't see any code to send data to mgr.shutdownCh or close it ?

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
+			case req := <-mgr.requestChan:
+				pendingRequests = append(pendingRequests, req)
+			}
+		}
+
+		for len(mgr.requestChan) > 0 {
+			select {
+			case req := <-mgr.requestChan:
+				pendingRequests = append(pendingRequests, req)
+			case <-mgr.shutdownCh:
```
williamlin-suse: same question as line 112

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
+			pendingRequests = pendingRequests[1:]
+			continue
+		default:
+		}
+
+		scanner, err := mgr.clusterHelper.PickLeastLoadedScanner(mgr.scannerHealthChecker)
+
+		if err != nil {
+			// Retry - add backoff and keep it in queue for next iteration
+			req.attempts++
```
williamlin-suse, line 121: Is there a upper limit for `attempt`?
Or `attempt` could grow to a huge number like 1,000,000 ?

controller/kv/helper.go
pohanhuangtw: I will remove this line, thanks

controller/kv/helper.go
pohanhuangtw: you are right.

controller/kv/helper.go
pohanhuangtw: you are right, let me clean it up

controller/rpc/scanner.go
pohanhuangtw: you are right, let me clean it up.

controller/rpc/scanner.go
pohanhuangtw: yes, I think this is not needed, will remove it.

controller/rpc/scanner.go
pohanhuangtw: It will keep growiong, but it actually rrturn fail becauase the timeout will be end 20 min, which retry 2xx at most.

So I will not add a max for this.

controller/kv/helper.go
williamlin-suse: Is line 1710 correct for checking common.ErrObjectNotFound ?

pohanhuangtw: I think it is, we should check it before we do the update.

controller/rpc/scanner.go
```diff
@@ -101,31 +103,29 @@ func (mgr *ScannerAcquisitionManager) startWorker() {
...
 	for {
 		// If no pending requests, wait for new ones from channel (blocking)
 		if len(pendingRequests) == 0 {
-			select {
-			case <-mgr.shutdownCh:
-				return
-			case req := <-mgr.requestChan:
-				pendingRequests = append(pendingRequests, req)
-			}
+			pendingRequests = append(pendingRequests, <-mgr.requestChan)
```
williamlin-suse, line 97: Why is line 125 required when line 128 `mgr.collectPendingRequests()` is always called immediately later?

pohanhuangtw: This is because once the code at line 125 unblocks, mgr.collectPendingRequests() ensures that the pending requests are fetched correctly.

pohanhuangtw: Should I add some comment?

controller/rpc/scanner.go
williamlin-suse: where is the `timeout will be end 20 min` logic?

pohanhuangtw: This is triggered here, and the timeout will ensure the operation not process too long. https://github.com/neuvector/neuvector/blob/1b9f775985b066a964824eb8cb1e31a6bfb447ac/controller/rest/repository.go#L93

pohanhuangtw: https://github.com/neuvector/neuvector/blob/1b9f775985b066a964824eb8cb1e31a6bfb447ac/controller/rest/repository.go#L26

controller/cache/scan.go
```diff
@@ -1239,24 +1239,16 @@ func ScannerUpdateHandler(nType cluster.ClusterNotifyType, key string, value []b
...
-					rpc.ScanCreditMgr.AddScanner(&s)
-					if err := scan.AddScanner(s.ID); err != nil {
-						log.WithError(err).Warn("failed to add scanner to reg scheduler")
-					}
-					if err := addScanner(s.ID); err != nil {
-						log.WithError(err).Warn("failed to add scanner to scheduler")
-					}
-				} else if s.ID == localDev.Ctrler.ID {
-					rpc.ScanCreditMgr.AddScanner(&s)
+				if isNewScanner && (!s.BuiltIn || s.ID == localDev.Ctrler.ID) {
```
holyspectral, line 1255: I'd recommend adding comments about these conditions.

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
 
-func (mgr *ScanCreditManager) getScannerEndpoint(scanner *share.CLUSScanner) string {
+// CleanUpScannerResources cleans up scanner resources when the cleanup is performed by consul KV.
+func (mgr *ScannerAcquisitionManager) CleanUpScannerResources(scanner *share.CLUSScanner) {
+	endpoint := mgr.getScannerEndpoint(scanner)
+	cluster.DeleteGRPCClient(endpoint)
+}
+
+func (mgr *ScannerAcquisitionManager) getScannerEndpoint(scanner *share.CLUSScanner) string {
 	return fmt.Sprintf("%s:%v", scanner.RPCServer, scanner.RPCServerPort)
```
holyspectral: I'd recommend using https://pkg.go.dev/net#JoinHostPort

controller/kv/helper.go
```diff
@@ -134,6 +136,22 @@ type ClusterHelper interface {
 	PutScannerStats(id string, objType share.ScanObjectType, result *share.ScanResult) error
 	GetScannerDB(store string) []*share.CLUSScannerDB
 
+	// GetAvailableScanners returns a list of all available scanners.
+	// Expected outcome: []*share.CLUSScanner representing all registered scanners in the cluster.
+	GetAvailableScanners() []*share.CLUSScanner
```
holyspectral: Does `GetAvailableScanners()` have to return a slice of pointers?

controller/kv/helper.go
```diff
@@ -1582,6 +1616,135 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+			log.Error("scanner is nil, skipping")
+			continue
+		}
+
+		if ScannerHealthCheckFunc != nil {
+			if err := ScannerHealthCheckFunc(scanner.ID, time.Second*10); err != nil {
+				log.WithFields(log.Fields{"scanner": scanner.ID, "error": err}).Error("Failed to ping scanner")
+				continue
+			}
+		}
```
holyspectral: Question: Do we really need this?  My understanding is that we've had logics to ping these scanners [here](https://github.com/neuvector/neuvector/blob/52396fc8f4dca37fbc3665d458e412b4d5a29c51/controller/cache/cache.go#L1759).  Adding the health check here might help, but there is still a window that you assign a job to a scanner that repeatedly restarts.

controller/kv/helper.go
```diff
@@ -1582,6 +1616,135 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+
+		return selectedScanner, nil
+	}
+
+	return nil, share.ErrNoScannerAvailable
+}
+
+func (m clusterHelper) ReleaseScanCredit(scannerId string) error {
+	// Infinite retry loop to ensure credit is never lost
+	// This is critical because losing credits would cause scanner starvation
```
holyspectral: I guess the comments here are out of date?

controller/kv/helper.go
```diff
@@ -1582,6 +1616,135 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+		scanner, rev, err := m.GetScannerRev(scannerId)
+		if err != nil {
+			// Check if scanner was deleted (object not found)
+			if errors.Is(err, common.ErrObjectNotFound) {
+				// Scanner has been deleted - credit is implicitly released with the scanner
+				log.WithFields(log.Fields{"scanner": scannerId}).Debug("Scanner not found during credit release, likely deleted")
+				return nil
+			}
+
+			// Network error or other temporary failure - retry with backoff
```
holyspectral, line 1740: Probably I missed it, but where is the backoff implemented?

controller/controller.go
```diff
@@ -630,6 +626,10 @@ func main() {
 		getConfigKvData, evqueue, keyRotationDuration)
 	ruleid.Init()
 
+	// Each scanner can handle multiple requests concurrently, set 2 to avoid OOM.
```
holyspectral: Is this comment out of date?

share/scan/scan_utils.go
```diff
@@ -56,26 +56,27 @@ var OSPkgFiles utils.Set = utils.NewSet(
...
+	share.ScanErrorCode_ScanErrXrayAPI:               "Xray API call error",
+	share.ScanErrorCode_ScanErrContainerExit:         "container exit",
+	share.ScanErrorCode_ScanErrAuthentication:        "authentication error",
+	share.ScanErrorCode_ScanErrCertificate:           "certificate error",
+	share.ScanErrorCode_ScanErrCanceled:              "scan canceled",
+	share.ScanErrorCode_ScanErrDriverAPINotSupport:   "driver API not supported",
+	share.ScanErrorCode_ScanErrImageNotFound:         "Image not found",
+	share.ScanErrorCode_ScanErrAwsDownloadErr:        "Aws Resource download error",
+	share.ScanErrorCode_ScanErrArgument:              "invalid input arguments",
+	share.ScanErrorCode_ScanErrAcquireScannerTimeout: "acquire scanner timeout",
```
holyspectral, line 79: Is there any background why `ScanErrorCode_ScanErrSignatureScanError` gets skipped?

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
+	mgr.SetScannerHealthChecker(mgr.Ping)
+	return mgr
 }
 
-func (mgr *ScanCreditManager) releaseScanCredit(sid string) {
-	mgr.mutex.Lock()
-	defer mgr.mutex.Unlock()
+func (mgr *ScannerAcquisitionManager) SetScannerHealthChecker(healthChecker ScannerHealthCheckFunc) {
+	mgr.scannerHealthChecker = healthChecker
+}
```
holyspectral: It's really up to each one's taste, but since `mgr.scannerHealthChecker` will not change after it's set, I'd declare it along with ScannerAcquisitionManager.

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
-		log.WithFields(log.Fields{"error": err, "scanner": sid}).Error()
-	} else {
-		mgr.signalScanCredit()
+func (mgr *ScannerAcquisitionManager) GetMaxConcurrentScansPerScanner() int {
+	return mgr.maxConcurrentScansPerScanner
+}
+
+func (mgr *ScannerAcquisitionManager) SetClusterHelper(clusterHelper kv.ClusterHelper) {
+	mgr.clusterHelper = clusterHelper
+}
```
holyspectral: Similarly, we don't need to allow this to be changed.

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
+	if exp > float64(MaxDelay) {
+		exp = float64(MaxDelay)
 	}
+
+	// between [1-JitterFactor, 1+JitterFactor]
+	jitter := 1 - JitterFactor + rand.Float64()*(2*JitterFactor)
+	delay := exp * jitter
+
+	return time.Duration(delay)
+}
```
holyspectral: Can we use an existing package to do the exponential retry?  For example, https://github.com/cenkalti/backoff/ or [Backoff.Step()](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#Backoff.Step).

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
+
+// startWorker starts the background worker that processes requests from priority queue
+func (mgr *ScannerAcquisitionManager) startWorker() {
+	mgr.workerStarted.Do(func() {
```
holyspectral: I think the function `startWorker()` here is a little confusing because it doesn't always start worker.  A better alternative is to move this section into `acquireScanner()` and rename `workerStarted` to include `Once` in its naming, so we know that `go mgr.processRequestQueue()` will run only once.

controller/rpc/scanner.go
```diff
@@ -388,9 +441,9 @@ func ScanPackage(ctx context.Context, pkgs []*share.ScanAppPackage) (*share.Scan
 }
 
 func ScanAwsLambdaFunc(ctx context.Context, funcInput *share.CLUSAwsFuncScanInput) (*share.ScanResult, error) {
```
holyspectral: I think we can just remove this function because it's not used anymore.

controller/rpc/scanner.go
```diff
@@ -362,13 +411,17 @@ func ScanImage(scanner string, ctx context.Context, req *share.ScanImageRequest)
 }
 
 func ScanPackage(ctx context.Context, pkgs []*share.ScanAppPackage) (*share.ScanResult, error) {
```
holyspectral: I think we can just remove this function because it's not used anymore.

controller/rpc/scanner.go
```diff
@@ -313,40 +360,42 @@ func ScanPlatform(scanner string, k8sVersion, ocVersion string, timeout time.Dur
...
+	client, scanner, err := ScannerAcquisitionMgr.getScannerServiceClient(scanner)
 	if err != nil {
 		return nil, err
 	}
 
 	result, err := client.ScanImage(ctx, req)
 	if err == nil {
+		if result == nil {
+			return nil, fmt.Errorf("scan image returned nil result")
+		}
```
holyspectral, line 381: Maybe add more information about the failing scanner?

controller/rpc/scanner.go
```diff
@@ -313,40 +360,42 @@ func ScanPlatform(scanner string, k8sVersion, ocVersion string, timeout time.Dur
...
-			// Decrement the active task count for the scanner
-			ScanCreditMgr.decScanningCount(scanner)
-		}
-	}()
+	if hasAcquiredScanner {
+		defer func() {
+			if err := ScannerAcquisitionMgr.releaseScanner(scanner); err != nil {
+				log.WithFields(log.Fields{"error": err, "scanner": scanner}).Error("failed to release scan credit")
+			}
+		}()
```
holyspectral, line 369: Could we move this defer call to where `hasAcquiredScanner` is assigned to true?  Then we don't need this flag anymore.

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
 
-func (mgr *ScanCreditManager) CountScanners() (busy, idle uint32) {
-	mgr.mutex.RLock()
-	defer mgr.mutex.RUnlock()
-
-	for _, s := range mgr.scannerLoadBalancer.GetActiveScanners() {
-		if s.AvailableScanCredits > 0 {
+func (mgr *ScannerAcquisitionManager) CountScanners() (busy, idle uint32) {
+	for _, scanner := range mgr.clusterHelper.GetAvailableScanners() {
+		if scanner.ScanCredit <= 0 {
```
holyspectral, line 230: hmm good eyes.

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
-func (mgr *ScanCreditManager) signalScanCredit() {
-	select {
-	case mgr.creditPool <- struct{}{}:
-		// Singal the pending task
-	default:
-		// Signal already pending, do not block
-	}
-}
+// processRequestQueue processes scanner acquisition requests in pull-time order
+func (mgr *ScannerAcquisitionManager) processRequestQueue() {
```
holyspectral: Given that this function is designed to never return, it should be reflected in its naming, e.g., `requestProcessLoop`.

controller/rpc/scanner.go
holyspectral: @pohanhuangtw I guess I know what you mean, but this is not how golang channel typically works.  It's expected that items are fetched from a channel and processed one by one.  

If you want the behavior of a priority queue, you should get the items one by one and push to a [heap](https://pkg.go.dev/container/heap), instead of getting using sort, and then fetch items from the priority queue using another go routine.

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
+		if err != nil {
+			// Retry - add backoff and keep it in queue for next iteration
+			req.attempts++
+			time.Sleep(mgr.backoffDelay(req.attempts))
+			continue
+		} else {
+			pendingRequests = pendingRequests[1:]
+			req.resultChan <- acquireScannerResult{scannerID: scanner.ID, err: nil}
+			close(req.resultChan)
+		}
```
holyspectral, line 124: nit: we don't need a `else`.

```suggestion
		if err != nil {
			// Retry - add backoff and keep it in queue for next iteration
			req.attempts++
			time.Sleep(mgr.backoffDelay(req.attempts))
			continue
		}
		pendingRequests = pendingRequests[1:]
		req.resultChan <- acquireScannerResult{scannerID: scanner.ID, err: nil}
		close(req.resultChan)
```

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
+
+		// Check if context is still valid
+		select {
+		case <-req.ctx.Done():
+			req.resultChan <- acquireScannerResult{err: req.ctx.Err()}
+			close(req.resultChan)
+			pendingRequests = pendingRequests[1:]
+			continue
+		default:
+		}
```
holyspectral, line 116: Generally it's not a good idea to store context in a struct.  Could you elaborate more about the behavior you want to achieve here?

controller/rpc/scanner.go
```diff
@@ -3,193 +3,242 @@ package rpc
...
-		log.WithFields(log.Fields{"error": err, "scanner": scannerID}).Error()
-		return err
+	mgr.requestChan <- req
+
+	select {
+	case result := <-req.resultChan:
+		return result.scannerID, result.err
+	case <-ctx.Done():
+		return "", ctx.Err()
 	}
```
holyspectral, line 167: There is still potential race condition here.  The steps are:
1. [main] acquireScanner() is called, and a request is pushed into requestChan().
2. [worker] the worker passes the context check and enters `PickLeastLoadedScanner()`.
3. [main] its context expires, so `<-ctx.Done()` returns and this function exits. 
4. [worker] worker sends the scanner information via resultChan.

Then this scan credit will never get released.

controller/rpc/scanner.go
pohanhuangtw: @holyspectral, using a heap sounds reasonable, but since there are only about 6 items, source: https://github.com/neuvector/neuvector/blob/55511a74ce6261b251d18d1a026abf7349cdf16f/controller/rest/longpoll.go#L372
The default is 6, so I think using heap is a bit overkill here.

controller/kv/helper.go
pohanhuangtw: I think we still need this. Even with periodic checks, there’s still a chance we assign a job to a scanner that has just gone down or is restarting, so this extra health check helps reduce that risk.

controller/kv/helper.go
pohanhuangtw: I’m not sure I follow. Do you mean we’re just getting a slice of the object?

controller/kv/helper.go
pohanhuangtw: it retried out side of the caller with https://github.com/pohanhuangtw/neuvector/blob/52396fc8f4dca37fbc3665d458e412b4d5a29c51/controller/rpc/scanner.go#L82

controller/rpc/scanner.go
pohanhuangtw: yes, sounds great.

controller/rpc/scanner.go
pohanhuangtw: Doing this is to check if the request is timeout, to ensure the we are not assign the scanner to the timeout one.

pohanhuangtw: @holyspectral I replace the ctx with a channel like the following, which eliminate the issue you concern, along side with the race issue you mentioned.
```go
func (mgr *ScannerAcquisitionManager) processRequestQueue() {
....
		// Check if context is still valid
		select {
		case <-req.shutdownChan:
			req.resultChan <- acquireScannerResult{err: context.Canceled}
			close(req.resultChan)
			pendingRequests = pendingRequests[1:]
			continue
		default:
		}
		scanner, err := mgr.clusterHelper.PickLeastLoadedScanner(mgr.scannerHealthChecker)
...
}


// acquireScanner acquires a scanner using priority queue to maintain pull order
func (mgr *ScannerAcquisitionManager) acquireScanner(ctx context.Context) (string, error) {
	req := &acquireScannerRequest{
		pullTime:     time.Now(),
		resultChan:   make(chan acquireScannerResult, 1),
		attempts:     0,
		shutdownChan: make(chan struct{}),
	}

	mgr.requestChan <- req

	select {
	case result := <-req.resultChan:
		return result.scannerID, result.err
	case <-ctx.Done():
		close(req.shutdownChan)

		select {
		case result := <-req.resultChan:
			// If the scanner is acquired but timeout, release it.
			if result.scannerID != "" {
				mgr.releaseScanner(result.scannerID)
			}
		default:
		}

		return "", ctx.Err()
	}
}
```

what do you think?

share/scan/scan_utils.go
pohanhuangtw: I think it's simply not implemented by the pervious people, will pick up in the next commits, thanks

controller/rpc/scanner.go
pohanhuangtw: you are right, I make this just for testing purpose.

pohanhuangtw: I will remove it in next commits, replace it with a var can be inject in tests.

controller/rpc/scanner.go
pohanhuangtw: good call

controller/kv/helper.go
holyspectral: Generally if we don't need to modify the object, we don't have to return a pointer, so I wonder if there is any use case for `[]*share.CLUSScanner` instead of `[]share.CLUSScanner`.

controller/kv/helper.go
holyspectral: Of course it's okay to try to ping a scanner before assigning a job to it, but my point is that, [the logic](https://github.com/neuvector/neuvector/blob/52396fc8f4dca37fbc3665d458e412b4d5a29c51/controller/cache/cache.go#L1759) has already run every one minute.  Adding a separate check here is a little redundant, and doesn't really resolve the problem where a scanner can go offline after getting a job assigned.

If this is a real issue, which I think we both agree, then we probably need a different solution than pinging it before assigning a job to it.

controller/kv/helper.go
holyspectral: But it's for `processRequestQueue()`, not for `ReleaseScanCredit()`, am I right?

share/scan/scan_utils.go
holyspectral: Since it's just one line, I'd recommend adding it, but feel free to skip it if we found it's bigger than that.

controller/rpc/scanner.go
holyspectral: If using a heap is overkill, then maybe implementing a scheduling is overkill too? lol

Anyway, I don't mind if we want to have our own implementation, but having a non-standard way to handle items in a channel is an issue.  I'd be open to other options.

controller/rpc/scanner.go
holyspectral: yeah I think it looks good in general.  Basically you have to close and then drain the channel.

However, I think you can't close the channel from `processRequestQueue()` because that would lead to the channel being closed twice.  It's better to have one dedicated go routine to own and close the channel.

controller/rpc/scanner.go
holyspectral: If for test purpose, then maybe we can just NOT expose it?  Of course keeping it in a var works too since we all do white box testing.

controller/rpc/scanner.go
holyspectral: Let me think about a better design and discuss with you.

controller/kv/helper.go
pohanhuangtw: yes, you are right

controller/kv/helper.go
pohanhuangtw: > doesn't really resolve the problem where a scanner can go offline after getting a job assigned.

Honestly, there is nothing we can do about this, just let the job retry, that what we can do.
I will remove this and the related test.

controller/kv/helper.go
pohanhuangtw: not really, will replace with []share.CLUSScanner

controller/rpc/scanner.go
pohanhuangtw: @holyspectral For the double clean up concern, I proposed the following improvement
- close everything from the caller, not in the `processRequestQueue`
- wait for the receive data from req.resultChan then close

This ensure
- single source of close
- any case of timeout, the credit is release properly
```go
func (mgr *ScannerAcquisitionManager) collectPendingRequests() []*acquireScannerRequest {
	pendingRequests := []*acquireScannerRequest{}
	for {
		select {
		case req := <-mgr.requestChan:
			select {
			case <-req.shutdownChan:
				req.resultChan <- acquireScannerResult{err: context.Canceled} // add this to filter the timeout one
			default:
				pendingRequests = append(pendingRequests, req)
			}
		default:
			return pendingRequests
		}
	}
}

// requestProcessLoop processes scanner acquisition requests in pull-time order
func (mgr *ScannerAcquisitionManager) requestProcessLoop() {
	var pendingRequests []*acquireScannerRequest

	for {
		// If no pending requests, wait for new ones from channel (blocking)
		if len(pendingRequests) == 0 {
			pendingRequests = append(pendingRequests, <-mgr.requestChan)
		}

		pendingRequests = append(pendingRequests, mgr.collectPendingRequests()...)

		// Sort by pullTime to get earliest request first
		sort.Slice(pendingRequests, func(i, j int) bool {
			return pendingRequests[i].pullTime.Before(pendingRequests[j].pullTime)
		})
		req := pendingRequests[0]

		// Check if context is still valid
		select {
		case <-req.shutdownChan:
			req.resultChan <- acquireScannerResult{err: context.Canceled}
			pendingRequests = pendingRequests[1:]
			continue
		default:
		}
		scanner, err := mgr.clusterHelper.PickLeastLoadedScanner(mgr.ScannerHealthChecker)

		if err != nil {
			// Retry - add backoff and keep it in queue for next iteration
			req.attempts++
			time.Sleep(req.backoff.NextBackOff())
			continue
		}
		pendingRequests = pendingRequests[1:]
		req.resultChan <- acquireScannerResult{scannerID: scanner.ID, err: nil}
	}
}

// acquireScanner acquires a scanner using priority queue to maintain pull order
func (mgr *ScannerAcquisitionManager) acquireScanner(ctx context.Context) (string, error) {
	// Initialize backoff with default values
	backoff := backoff.NewExponentialBackOff()
	backoff.InitialInterval = BaseDelay
	backoff.MaxInterval = MaxDelay
	backoff.RandomizationFactor = JitterFactor

	req := &acquireScannerRequest{
		pullTime:     time.Now(),
		resultChan:   make(chan acquireScannerResult, 1),
		attempts:     0,
		shutdownChan: make(chan struct{}),
		backoff:      backoff,
	}
	defer close(req.resultChan)

	mgr.requestChan <- req

	select {
	case result := <-req.resultChan:
		close(req.shutdownChan)
		return result.scannerID, result.err
	case <-ctx.Done():
		close(req.shutdownChan)

		result := <-req.resultChan
		// If the scanner is acquired but timeout, release it.
		if result.scannerID != "" {
			err := mgr.releaseScanner(result.scannerID)
			if err != nil {
				log.WithFields(log.Fields{"error": err, "scanner": result.scannerID}).Error("Failed to release scanner credit in Consul")
			}
		}

		return "", ctx.Err()
	}
}
```

what do you think?

controller/cache/scan.go
holyspectral: Aren't these two conditions are the same thing per [here](https://github.com/neuvector/neuvector/blob/177eb7f65257a477473b2f6bbbd24ad0a563ad34/controller/grpc.go#L203)?

controller/kv/helper.go
holyspectral: Looks like this hasn't been addressed yet?

controller/kv/helper.go
holyspectral: Is this addressed?

controller/kv/helper.go
```diff
@@ -1582,6 +1618,202 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+		value, err := json.Marshal(selectedScanner)
+		if err != nil {
+			return nil, err
+		}
+
+		// If the CAS is failed, try the next least loaded scanner
+		if err := cluster.PutRev(key, value, rev); err != nil {
+			log.WithFields(log.Fields{"error": err, "scanner": selectedScanner.ID}).Error("Failed to update scanner credit")
+			continue
+		}
```
holyspectral: The comment and the implementation are not aligned?

controller/kv/helper.go
```diff
@@ -1582,6 +1618,202 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+		return nil
+	}
+
+	return fmt.Errorf("failed to release scanner credit after %d retries", maxRetries)
+}
+
+// TrackCreditAcquisition records or removes a controller's ownership of scanner credits.
+func (m clusterHelper) TrackCreditAcquisition(controllerID, scannerID string, maxConcurrentScansPerScanner int, acquire bool) error {
+	key := share.CLUSScannerCreditOwnerKey(controllerID)
+	maxRetries := 10
```
holyspectral: It's better to make this configurable, so for slower clusters we're able to change.

controller/kv/helper.go
```diff
@@ -1582,6 +1618,202 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+
+	for i := 0; i < maxRetries; i++ {
+		scanners := make(map[string]int)
+		value, rev, err := m.get(key)
+		if err == nil && value != nil {
+			err = json.Unmarshal(value, &scanners)
+			if err != nil {
+				return err
+			}
+		}
```
holyspectral:
```suggestion
		if err != nil {
			// error handling
		}
		if value == nil {
			// error handling
		}
		err = json.Unmarshal(value, &scanners)
		if err != nil {
			return err
		}
```

controller/kv/helper.go
```diff
@@ -1582,6 +1618,202 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+
+func (m clusterHelper) PickLeastLoadedScanner() (*share.CLUSScanner, error) {
+	scanners := m.GetAvailableScanners()
+	if len(scanners) == 0 {
+		return nil, share.ErrNoScannerFound
+	}
+
+	// To minimize CAS conflicts and improve fairness, sort scanners by ScanCredit (descending).
+	// This ensures we try to consume credits from the most available scanner first.
+	sort.Slice(scanners, func(i, j int) bool {
```
holyspectral: If you still want to use sort instead of heap, I'd recommend using `slices.SortFunc()` per [its document](https://pkg.go.dev/sort#Slice) mentioned.

controller/rpc/scanner.go
```diff
@@ -3,193 +3,229 @@ package rpc
...
+	for {
+		// If no pending requests, wait for new ones from channel (blocking)
+		if len(pendingRequests) == 0 {
+			pendingRequests = append(pendingRequests, <-mgr.requestChan)
+		}
+
+		pendingRequests = append(pendingRequests, mgr.collectPendingRequests()...)
+
+		// Sort by pullTime to get earliest request first
+		sort.Slice(pendingRequests, func(i, j int) bool {
```
holyspectral: If you still want to use sort instead of heap, I'd recommend using `slices.SortFunc()` per [its document](https://pkg.go.dev/sort#Slice) mentioned.

holyspectral (commented):
It looks much better now and I'm glad to see many review comments are now addressed.  I will take a look again once all the comments are addressed.

controller/kv/helper.go
pohanhuangtw: I think they’re aligned. The scan credits have already been sorted, so if this picks the wrong one, it will just pick the next.

controller/kv/helper.go
pohanhuangtw: Good point, I’ll make the max retry count configurable so it can be tuned for slower clusters.

controller/rpc/scanner.go
pohanhuangtw: you are right, the current approach may not be the best.
I would love to hear a better way to ensure the older task guarantee finished before context cancel

controller/cache/scan.go
pohanhuangtw: Yes, they are. I’m afraid there might be some edge cases, so I merged them here to keep the behavior consistent.

controller/controller.go
```diff
@@ -626,10 +623,20 @@ func main() {
 		}
 	}
 
+	maxClusterOperationRetries := 10
+	if os.Getenv("MAX_CLUSTER_OPERATION_RETRIES") != "" {
+		if retries, err := strconv.Atoi(os.Getenv("MAX_CLUSTER_OPERATION_RETRIES")); err == nil {
```
holyspectral: I'd log error if it can't be parsed into an integer.

controller/controller.go
```diff
@@ -626,10 +623,20 @@ func main() {
 		}
 	}
 
+	maxClusterOperationRetries := 10
+	if os.Getenv("MAX_CLUSTER_OPERATION_RETRIES") != "" {
```
holyspectral: Could we put the logic in `newClusterHelper()` instead so you don't have to pass it layer by layer from `main()`?

controller/controller.go
```diff
@@ -626,10 +623,20 @@ func main() {
 		}
 	}
 
+	maxClusterOperationRetries := 10
```
holyspectral: nit:
```suggestion
	maxScanCreditRetries := 10
```

controller/kv/helper.go
```diff
@@ -1582,6 +1627,227 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < m.maxClusterOperationRetries; i++ {
+		if err = cluster.PutRev(key, value, 0); err == cluster.ErrPutCAS {
+			continue
+		}
+		break
+	}
```
holyspectral: We probably don't need to call `json.Marshal()` since it's always empty?

```suggestion
	var err error

	for i := 0; i < m.maxClusterOperationRetries; i++ {
		if err = cluster.PutRev(key, []byte("{}"), 0); err == cluster.ErrPutCAS {
			continue
		}
		break
	}
```

controller/kv/helper.go
```diff
@@ -1582,6 +1627,227 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+	scanners := make(map[string]int)
+
+	var value []byte
+	var err error
+	value, err = json.Marshal(scanners)
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < m.maxClusterOperationRetries; i++ {
```
holyspectral: Maybe [RetryOnCASError()](https://github.com/neuvector/neuvector/blob/0089e547d4c4de4ecefc5f604a8f857d22d35596/controller/kv/certmanager.go#L91) would help?  Feel free to refactor it as you like.

controller/kv/helper.go
```diff
@@ -1582,6 +1627,227 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+			if err := json.Unmarshal(value, &scanners); err != nil {
+				return err
+			}
+		} else if !acquire {
+			// Nothing has been recorded for this controller yet, so releasing is a no-op.
+			return nil
+		}
+
+		if acquire {
+			// Acquire credit: increment credit count for given scanner, up to the max allowed
```
holyspectral: We should return error when the credit count is out of bound instead.

controller/kv/helper.go
```diff
@@ -1582,6 +1627,227 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+		if err = cluster.PutRev(key, value, 0); err == cluster.ErrPutCAS {
+			continue
+		}
+		break
+	}
+	return nil
+}
+
+// TrackCreditAcquisition records or removes a controller's ownership of scanner credits.
+func (m clusterHelper) TrackCreditAcquisition(controllerID, scannerID string, maxConcurrentScansPerScanner int, acquire bool) error {
```
holyspectral: nit: I'd recommend defining an enum to replace `acquire bool` because `acquire=false` doesn't mean `release=true`.

controller/kv/helper.go
```diff
@@ -1582,6 +1627,227 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+		}
+
+		value, err = json.Marshal(scanners)
+		if err != nil {
+			return err
+		}
+
+		if err := cluster.PutRev(key, value, rev); err == nil {
+			return nil
+		}
```
holyspectral: We should log unexpected errors, e.g., non ErrPutCAS.

holyspectral (commented):
Thanks for the change!  I left my comments based on https://github.com/pohanhuangtw/neuvector/compare/f83cd188e5a57453f7135b4434d527853c190617...0089e547d4c4de4ecefc5f604a8f857d22d35596 

I know everyone has different preference, but I would recommend not rebasing to the main branch or squashing commits until you get an approval, especially for a medium size or large size PR.  This will help reviewer a lot.

controller/rpc/scanner.go
holyspectral: This is still not resolved.

controller/kv/helper.go
holyspectral: Do you have plan to address this?

controller/kv/helper.go
holyspectral: I mean, if I read the code correctly, all errors are retried here instead of just CAS errors.  Maybe you need to update the comment accordingly?

controller/kv/helper.go
pohanhuangtw: I think I miss it when rebasing, will fix it in next commits

controller/controller.go
pohanhuangtw: I think this is a great idea, thanks for reminding!

controller/kv/helper.go
pohanhuangtw: you are right about this!

controller/kv/helper.go
```diff
@@ -331,6 +364,17 @@ func newClusterHelper(id, version string, persist bool, keyRotationDuration time
...
 	clusHelperImpl.persist = persist
 	clusHelperImpl.keyRotationDuration = keyRotationDuration
+
+	maxScanCreditRetries := 10
+	if os.Getenv("MAX_CLUSTER_OPERATION_RETRIES") != "" {
+		retries, err := strconv.Atoi(os.Getenv("MAX_CLUSTER_OPERATION_RETRIES"))
+		if err != nil {
+			log.WithFields(log.Fields{"error": err}).Error("Failed to parse MAX_CLUSTER_OPERATION_RETRIES as int")
+		}
+		maxScanCreditRetries = retries
```
holyspectral:
```suggestion
		if err != nil {
			log.WithFields(log.Fields{"error": err}).Error("Failed to parse MAX_CLUSTER_OPERATION_RETRIES as int")
		} else {
			maxScanCreditRetries = retries
		}
```

holyspectral (changes requested):
Thanks.  Just a minor issue.  

I will take another overall look once this is fixed.

controller/kv/helper.go
```diff
@@ -1582,6 +1643,244 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+	key := share.CLUSScannerKey(scannerId)
+	for retry := 0; retry < m.maxScanCreditRetries; retry++ {
+		// Get current scanner state with revision
+		scanner, rev, err := m.GetScannerRev(scannerId)
+		if err != nil {
+			// Check if scanner was deleted (object not found)
+			if errors.Is(err, common.ErrObjectNotFound) {
+				// Scanner has been deleted - credit is implicitly released with the scanner
+				log.WithFields(log.Fields{"scanner": scannerId}).Debug("Scanner not found during credit release, likely deleted")
+				return nil
```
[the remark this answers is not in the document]
pohanhuangtw: This is intentional: if the scanner is missing, it was deleted and the credit is implicitly released. We return nil to keep the release operation idempotent.

controller/kv/helper.go
```diff
@@ -1582,6 +1643,244 @@ func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
...
+	}
+
+	// Return the ownership to the scanner
+	for scannerID, credit := range scanners {
+		err = m.ReleaseScanCredit(scannerID, credit)
+		if err != nil {
+			return err
+		}
+	}
+
```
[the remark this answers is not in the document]
pohanhuangtw: Thanks — this is intentional.

`ReleaseScanCredit` is treated as idempotent: if the scanner is already gone, there’s nothing to release on that scanner side, and the new controller will reconcile the correct state.

controller/rpc/scanner.go
```diff
@@ -3,193 +3,230 @@ package rpc
...
+		case <-req.shutdownChan:
+			req.resultChan <- acquireScannerResult{err: context.Canceled}
+			pendingRequests = pendingRequests[1:]
+			continue
+		default:
+		}
+		scanner, err := mgr.clusterHelper.PickLeastLoadedScanner()
+
+		if err != nil {
+			// Retry - add backoff and keep it in queue for next iteration
```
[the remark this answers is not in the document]
pohanhuangtw: Will add reset in the acquireScanner function.

controller/rpc/scanner.go
```diff
@@ -3,193 +3,230 @@ package rpc
...
+			req.attempts++
+			time.Sleep(req.backoff.NextBackOff())
+			continue
+		}
+		pendingRequests = pendingRequests[1:]
+		req.resultChan <- acquireScannerResult{scannerID: scanner.ID, err: nil}
 	}
 }
```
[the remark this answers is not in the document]
pohanhuangtw: This goroutine is designed to run for the entire lifetime of the controller process. When the controller shuts down, the process terminates.

pohanhuangtw:
Merge it for resolving all of comments from copilot

PR #2137, opened 2025-10-30
NVSHAS-10116: Controller throwing errors '...Watch failure - error...'

williamlin-suse:


controller/resource/kubernetes_resource.go
```diff
@@ -1251,24 +1262,142 @@ func xlateValidatingWebhookConfiguration(obj metav1.Object) (string, interface{}
...
+exit_watcher:
+	for {
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				log.WithFields(log.Fields{"error": err}).Error("Channel is closed")
+				return
+			}
+			// We only care about write events, not events like CHMOD.
+			if event.Has(fsnotify.Remove) {
```
holyspectral, line 1331: I understand that fsnotify has some limitation like [this one](https://github.com/fsnotify/fsnotify?tab=readme-ov-file#linux).  Could you add some comments about why we check `fsnofity.Remove` here?

controller/resource/kubernetes_resource.go
```diff
@@ -1251,24 +1262,142 @@ func xlateValidatingWebhookConfiguration(obj metav1.Object) (string, interface{}
...
+}
+
+func (d *kubernetes) monitorK8sTokenFile() {
+	const filePath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+	targetPath := evalSymlinks(filePath)
+	info, err := os.Stat(targetPath)
+	if err != nil {
+		log.WithFields(log.Fields{"error": err}).Error("Failed to get file info")
+		return
+	}
```
holyspectral: My understanding is that `os.Stat()` will follow symlink.  Do we really need to evaluate the symlink separately?

williamlin-suse: In my testing os.Stat() doesn't follow the link. 
It simply returns the info of the symbolic itself(/var/run/secrets/kubernetes.io/serviceaccount/token) which never changes after it's created.

controller/resource/kubernetes_resource.go
williamlin-suse: comment added

controller/resource/kubernetes_resource.go
holyspectral: I do believe that `os.Stat()` will follow the symbolic link.   That’s how the underlying Linux stat and lstat syscalls work.  

I think the problem is probably that fsnotify can’t get the ideal path, i.e., the symbolic link path?

williamlin-suse: In my testing on SE's env, os.Stat() does not follow the link.
Let me ask SE to start the env again for confirmation

williamlin-suse: You are correct. 
`os.Stat()` does follow the symbolic link.
Don't know why I mistakenly observed that `os.Stat()` does not follow the symbolic link in that AKS env.

holyspectral (approved):
LGTM.  Thanks!

PR #2151, opened 2025-11-21
feat: refactor the rego generate process, expose GenerateRegoCode

pohanhuangtw:
### Summary
- Current function will call `AddPolicy` to connect the opa server. However, there is no opa server live in converter project, we need to extract the rego generate part out to GenerateRegoCode to avoid connect to opa server.


kyledong-suse (commented):
From the code logic perspective, it LGTM.
It may still need another review from @williamlin-suse and @holyspectral for the product level logic.

williamlin-suse (approved):
LGTM

PR #2159, opened 2025-12-01
NVSHAS-10194: [Controller] Use 'user_created' as default cfg_type val…

williamlin-suse:
…ue if it's empty in the request payload of creating DLP/WAF sensors

### DIFF
--- a/controller/rest/dlp_rule.go
+++ b/controller/rest/dlp_rule.go
@@ -552,12 +552,16 @@ func handlerDlpSensorCreate(w http.ResponseWriter, r *http.Request, ps httproute
 	body, _ := io.ReadAll(r.Body)
 	var rconf api.RESTDlpSensorConfigData
 	err := json.Unmarshal(body, &rconf)
-	if err == nil && rconf.Config != nil && rconf.Config.CfgType == "" {
+	if err != nil || rconf.Config == nil {
+		log.WithFields(log.Fields{"error": err, "rconf": rconf}).Error("Request error")
+		restRespError(w, http.StatusBadRequest, api.RESTErrInvalidRequest)
+		return
+	}
+	if rconf.Config.CfgType == "" {
 		rconf.Config.CfgType = api.CfgTypeUserCreated
 	}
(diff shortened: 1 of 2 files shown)


controller/rest/dlp_rule.go
kyledong-suse: I think this part can be restructured like the following to make it cleaner to understand:

```suggestion
	if err != nil || rconf.Config == nil {
		log.WithFields(log.Fields{"error": err}).Error("Request error")
		restRespError(w, http.StatusBadRequest, api.RESTErrInvalidRequest)
		return
	}
	
	if rconf.Config.CfgType == "" {
		rconf.Config.CfgType = api.CfgTypeUserCreated
	}
	
	if rconf.Config.CfgType != api.CfgTypeUserCreated &&
	   rconf.Config.CfgType != api.CfgTypeFederal {
		log.WithFields(log.Fields{"rconf": rconf}).Error("Request error")
		restRespError(w, http.StatusBadRequest, api.RESTErrInvalidRequest)
		return
	}
```

controller/rest/waf_rule.go
```diff
@@ -433,6 +433,9 @@ func handlerWafSensorCreate(w http.ResponseWriter, r *http.Request, ps httproute
 	body, _ := io.ReadAll(r.Body)
 	var rconf api.RESTWafSensorConfigData
 	err := json.Unmarshal(body, &rconf)
+	if err == nil && rconf.Config != nil && rconf.Config.CfgType == "" {
+		rconf.Config.CfgType = api.CfgTypeUserCreated
+	}
```
kyledong-suse: Same comment as the other one.

kyledong-suse (changes requested):
Generally looks good to me, would you please avoid redundant checks? Thanks.

controller/rest/dlp_rule.go
williamlin-suse: changed code as advised

controller/rest/waf_rule.go
williamlin-suse: changed code as advised

PR #2162, opened 2025-12-06
NVSHAS-10198: cannot add/edit user-created process profile rule in CRD group

williamlin-suse:
In NV 5.4.7, user-created process profile/file monitor rule in CRD group can still be added/edited thru REST API.
This PR is to have the same behavior as in NV 5.4.7

### DIFF
--- a/controller/rest/file_monitor.go
+++ b/controller/rest/file_monitor.go
@@ -197,8 +197,9 @@ func handlerFileMonitorConfig(w http.ResponseWriter, r *http.Request, ps httprou
 	profConf, profRev := clusHelper.GetFileMonitorProfile(group)
 	ruleConf, ruleRev := clusHelper.GetFileAccessRule(group)
 
-	if profConf != nil && profConf.CfgType == share.GroundCfg {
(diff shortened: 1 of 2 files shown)


controller/rest/file_monitor.go
```diff
@@ -197,11 +197,6 @@ func handlerFileMonitorConfig(w http.ResponseWriter, r *http.Request, ps httprou
 	profConf, profRev := clusHelper.GetFileMonitorProfile(group)
 	ruleConf, ruleRev := clusHelper.GetFileAccessRule(group)
 
-	if profConf != nil && profConf.CfgType == share.GroundCfg {
-		restRespError(w, http.StatusBadRequest, api.RESTErrOpNotAllowed)
-		return
-	}
-
```
kyledong-suse, line 204: Do we still want to check `if profConf == nil` and return error?
If we don't check it here, `profConf` will be used afterwards, and if it is nil, it may result in panic

williamlin-suse: 5.4.7 doesn't check `profConf == nil` here.
However, I added the code to check whether `profConf` or `ruleConf` is nil here. Both are expected to be non-nil.

PR #2166, opened 2025-12-11
chore: expose PrintMainFunction for converter

pohanhuangtw:
### Summary
- expose this for converter projects


kyledong-suse (approved):
LGTM

PR #2179, opened 2026-01-06
NVSHAS-10202: add triggerBenchRescan when the report is not exist in the consul

pohanhuangtw:
### Summary
- Sometimes, when all controller pods are down, the benchmark report is lost. Add `triggerBenchRescan` to re-run the benchmark when the report is missing and the rescan is not triggered by the enforcer.

### How to verify
- delete then all controller pods, then check the asset > node > compliance, make sure you can still read the report.


controller/rest/bench.go
```diff
@@ -742,6 +742,9 @@ func getCISStatusFromCluster(bench share.BenchType, id string) (int, string) {
 
 func getCISReportFromCluster(bench share.BenchType, id string, cpf *complianceProfileFilter, acc *access.AccessControl) (*api.RESTBenchReport, int, string) {
 	rpt, code, errMsg := _getCISReportFromCluster(bench, id, true, cpf)
+	if code == api.RESTErrFailReadCluster {
```
williamlin-suse, line 745: In func _getCISReportFromCluster(), there is comment `// not all bench type exist, for example custom check` :
```
key := share.CLUSBenchReportKey(id, bench)
	var r share.CLUSBenchReport
	if value, err := cluster.Get(key); err != nil || len(value) == 0 {
		// not all bench type exist, for example custom check, so use INFO level debug
		log.WithFields(log.Fields{"error": err, "key": key}).Info("Benchmark report not found")
		return nil, api.RESTErrFailReadCluster, "Failed to read benchmark report"
	} else {
        .......
    }
    ......
```
but triggerBenchRescan() is always called when api.RESTErrFailReadCluster is returned.
What if it's expected that the bench type doesn't exist?

pohanhuangtw: Hi @williamlin-suse ,
Thanks for your advice, so I add RESTErrBenchmarkNotFound in the latest commit to avoid such confusion, also add the missing bench that can be manually re-run.

pohanhuangtw:
Hi @williamlin-suse,
I think we should trigger a rescan even when the key is missing. If the controller’s KV store goes down, the report/key may not be regenerated automatically in time. So we should check whether the result exists in the KV store; if it doesn’t, we should regenerate it.

controller/rest/bench.go
williamlin-suse: RESTErrBenchmarkNotFound is added in which commit?

controller/rest/bench.go
```diff
@@ -768,6 +772,42 @@ func getKubeCISReportFromCluster(id string, cpf *complianceProfileFilter, acc *a
...
 }
 
+func triggerBenchRescan(bench share.BenchType, id string, acc *access.AccessControl) {
+	var invokeBench func(string) error
+	var err error
+
+	switch bench {
+	case share.BenchKubeMaster, share.BenchKubeWorker:
+		invokeBench = rpc.RunKubernetesBench
+	case share.BenchDockerHost, share.BenchContainer, share.BenchContainerSecret, share.BenchContainerSetID:
```
williamlin-suse, line 782: What about share.BenchCustomHost & share.BenchCustomContainer ?
What happens if custom check is configured & controller deployment is restarted for share.BenchCustomContainer and share.BenchCustomHost ?

pohanhuangtw: I add the support grpc calling for both of them in latest commits

controller/rest/bench.go
williamlin-suse: Where is RESTErrBenchmarkNotFound defined & used ?

pohanhuangtw: After investigating, we should re-run the benchmark whenever reading the report fails. The report key can disappear when the controller restarts / KV is unavailable, so it’s not always a clean “not found”. Therefore `RESTErrFailReadCluster` fits better than adding `RESTErrBenchmarkNotFound`.

PR #2184, opened 2026-01-15
NVSHAS-10221: [Controller] Export local groups(and other policies) th…

williamlin-suse:
…ru redirected UI from primary cluster always fails

### DIFF
--- a/controller/rest/federation.go
+++ b/controller/rest/federation.go
@@ -3512,6 +3512,10 @@ func handlerFedClusterForward(w http.ResponseWriter, r *http.Request, ps httprou
 	acc := accCaller
 	id := ps.ByName("id")
 	request := ps.ByName("request")
+	reqPath := request
+	if u, err := url.ParseRequestURI(request); err == nil {
+		reqPath = u.Path
+	}
 	forbidden := false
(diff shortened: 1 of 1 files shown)


controller/rest/federation.go
```diff
@@ -3512,6 +3512,10 @@ func handlerFedClusterForward(w http.ResponseWriter, r *http.Request, ps httprou
 	acc := accCaller
 	id := ps.ByName("id")
 	request := ps.ByName("request")
+	reqPath := request
+	if u, err := url.ParseRequestURI(request); err == nil {
+		reqPath = u.Path
```
pohanhuangtw, line 3517: why not just update request here?

pohanhuangtw: like 
```go
if u, err := url.ParseRequestURI(request); err == nil {
		request = u.Path
}
```

And keep the rest of the it as it was.
What do you think?

williamlin-suse: The following code still references `request` for other purpose.
Overwriting `request` with `reqPath` could break other functions

pohanhuangtw: Get it.

PR #2189, opened 2026-01-22
fix: fix the concurrent map read/write panic for macPortPairMap

kyledong-suse:


agent/engine.go
```diff
@@ -762,10 +764,20 @@ func getContainerIDByName(name string) string {
...
 	gInfoRLock()
-	defer gInfoRUnlock()
-	if pair, ok := gInfo.macPortPairMap[mac.String()]; ok {
-		for _, addr := range pair.Addrs {
-			if ip.String() == addr.IPNet.IP.String() {
+	pair, ok := gInfo.macPortPairMap[mac.String()]
+	var addrs []share.CLUSIPAddr
+	if ok && pair != nil {
+		// Copy the Addrs slice to avoid holding the lock during iteration
+		// and to prevent race conditions if the map entry is modified
```
williamlin-suse, line 771: With gInfoRLock()/gInfoRUnlock(), why could the map entry be modified to cause race conditions?
I wonder is the change in getIPAddrScope() necessary?

kyledong-suse: I think the read lock protects the map, not the `*pipe.InterceptPair` stores in it. If another goroutine modifies `pair.Addrs` while we iterate, that will be a race condition.

williamlin-suse: But here nothing is updated. It's simply iteration How could it cause panic?

PR #2191, opened 2026-01-27
NVSHAS-10131: Improve the Vulnerabilities Report

williamlin-suse:


controller/rest/host.go
```diff
@@ -17,24 +18,17 @@
...
-	query := restParseQuery(r)
-
+func getHosts(query *restQuery, acc *access.AccessControl) []*api.RESTHost {
 	var hosts []*api.RESTHost
-	var resp api.RESTHostsData
-	resp.Hosts = make([]*api.RESTHost, 0)
+	var retHosts []*api.RESTHost
 
+	if query.limit > 0 {
+		retHosts = make([]*api.RESTHost, 0, query.limit)
```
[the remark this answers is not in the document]
williamlin-suse: fixed in new commit

controller/rest/workload.go
```diff
@@ -307,10 +257,16 @@
 		sort.Slice(wls, func(i, j int) bool { return wls[i].Name < wls[j].Name })
 	}
 
+	var retWorkloads []*api.RESTWorkload
+	if query.limit > 0 {
+		retWorkloads = make([]*api.RESTWorkload, 0, query.limit)
```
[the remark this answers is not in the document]
williamlin-suse: fixed in new commit

controller/rest/workload.go
```diff
@@ -322,25 +273,82 @@ func handlerWorkloadListBase(apiVer string, w http.ResponseWriter, r *http.Reque
...
+	if query.start > 0 && count <= query.start {
+		restRespSuccess(w, r, resp, acc, login, nil, "Get container list")
+		return
+	}
+
+	// get POST body
+	idlist := utils.NewSet()
+	if r.Method == http.MethodPost {
+		var idListData api.RESTAssetIDList
+		body, _ := io.ReadAll(r.Body)
```
olblak: Is there any reason to not handle err here?

williamlin-suse: > Is there any reason to not handle err here?

It's just we don't usually handle io.ReadAll() error in NV controller's REST API handlers

controller/rest/workload.go
```diff
@@ -322,25 +273,82 @@ func handlerWorkloadListBase(apiVer string, w http.ResponseWriter, r *http.Reque
...
+			restRespError(w, http.StatusBadRequest, api.RESTErrInvalidRequest)
+			return
+		}
+		idlist = utils.NewSetFromStringSlice(idListData.IDs)
 	}
 
-	log.WithFields(log.Fields{"entries": len(respV1.Workloads)}).Debug("Response")
+	respV1.Workloads = getWorkloads(query, idlist, acc)
 
 	if apiVer == "v2" {
```
pohanhuang, line 345: Is it possible that we separate the v1, v2 logic with two function? The current implementation looks convolute for me.

williamlin-suse: Sure. Revert the changes in `controller/rest/workload.go` and move `getWorkloads()` to `controller/rest/scanner.go`

controller/api/apis.go
```diff
@@ -2533,6 +2534,27 @@ type RESTScanReportData struct {
...
+	HostName string `json:"host_name"`
+	RESTScanReport
+}
+
+type RESTHostScanReportData struct {
+	HostsScanData []*RESTHostScanData `json:"hosts_scan_data"`
+}
+
+type RESTWorkloadScanData struct {
+	//Workload     *RESTWorkloadBriefV2 `json:"workload"`
```
pohanhuang: please remove this line, if not using it.

controller/rest/scanner.go
```diff
@@ -268,6 +268,113 @@ func handlerScanWorkloadReport(w http.ResponseWriter, r *http.Request, ps httpro
...
+		}
+
+		// Sort
+		restNewSorter(data, query.sorts).Sort()
+
+		// Copy the result
+		wls = make([]*api.RESTWorkload, len(cached))
+		for i, d := range data {
+			wls[i] = d.(*api.RESTWorkload)
+		}
```
pohanhuang:
```suggestion
wls := make([]*api.RESTWorkload, 0, len(data))
for _, d := range data {
    wl, ok := d.(*api.RESTWorkload)
    if !ok {
        // handle unexpected type?
        continue // or return error
    }
    wls = append(wls, wl)
}
```

pohanhuang (changes requested):
Overall LGTM, left some minor comment

pohanhuang (approved):
lgtm

controller/rest/scanner.go
```diff
@@ -477,6 +587,45 @@ func handlerScanHostReport(w http.ResponseWriter, r *http.Request, ps httprouter
 	restRespSuccess(w, r, resp, acc, login, nil, "Get host scan report")
 }
 
+func handlerHostsScanReport(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
holyspectral, line 685: It looks like the query parameter will work on the host list, instead of the scan report.  This is a little not intuitive but I think we can just document it.

controller/rest/scanner.go
```diff
@@ -268,6 +268,116 @@ func handlerScanWorkloadReport(w http.ResponseWriter, r *http.Request, ps httpro
...
+		}
+		retWorkloads = wls[query.start:end]
+	}
+
+	log.WithFields(log.Fields{"entries": len(retWorkloads)}).Debug("Response")
+
+	return retWorkloads
+}
+
+func handlerWorkloadsScanReport(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
holyspectral: Similarly, it looks like the query parameter will work on the workload list, instead of the scan report.  This is a little not intuitive but I think we can just document it.

holyspectral (commented):
I think it looks good to me.  It's not very intuitive as to how query parameters will work on the two new APIs, `/v1/scan/hosts/scan_report` and `/v1/scan/workloads/scan_report`, in terms of RESTful, but I don't think it's a deal breaker.  Maybe we just need to mention it in the API docs.

For example, the `filters` query parameter on `/v1/scan/hosts/scan_report` API will work on hosts, instead of scan report.  Again, maybe we just need to document it.

holyspectral:
It looks like the last commit merged a lot of upstream changes, probably unintentionally.  Given that I don't have the permission to push to william's fork, to save everyone's time and keep history at the same place, I'll create a new PR based on this one.

PR #2197, opened 2026-02-05
fix: add task count to avoid the autoscale down the scanner replica number

pohanhuang:
### Problem
- Autoscale task count is incomplete — it does not account for tasks tracked by ScannerAcquisitionMgr, causing premature scale-down and RPC failures.
- Ensure the ownership with no CAS

### Solution
Add rpc.ScannerAcquisitionMgr.TaskCount() to the total task count.


williamlin-suse (approved):
LGTM

PR #2208, opened 2026-02-17
NVSHAS-10131: Improve the Vulnerabilities Report

holyspectral:
This PR is based on #2191 with unintentional changes removed.  

Here is the change that I can see from V2:
  1. Now the two APIs are POST APIs and use their own filtering logic instead of the host or workload ones.
  2. The API now takes filters from request data, rather than query parameter.
  3. API doc is updated accordingly.

== Edit on Feb. 20 ==

There are some changes in the implementation and API:

- Pages-related settings are removed because it can cause duplicate items easily in a dynamic environment, especially in workload scan.
- The cursor-related fields, `last_stop_*` and `stop_*` are now consolidated into ther `cursor` field.  The query and response share the same naming. 
- `cve_package` is added into the cursor so we can deal with multiple same CVE findings on different packages.
- The protocol for `/hosts/scan_report` and `/workloads/scan_report` is the same now, e.g., `hosts_scan_data` is renamed to `scan_data`.
- `assets_left` might not be very useful.  It's removed for now, but we can add it back per UI team's feedback.

### DIFF
--- a/controller/api/apis.yaml
+++ b/controller/api/apis.yaml
@@ -9848,7 +9848,6 @@ definitions:
     required:
       - workload_name
       - workload_domain
-      - workload_host_name
       - workload_image
       - workload_image_id
       - host_name
--- a/controller/rest/scanner.go
+++ b/controller/rest/scanner.go
@@ -343,7 +343,7 @@ func filterAssets(filters []api.RESTAssetsScanReportFilter, cursor *api.RESTScan
 
 	switch v := assets[0].(type) {
 	case *api.RESTWorkload:
-		filterNames := utils.NewSetFromStringSlice([]string{"domain", "host_name"})
+		filterNames := utils.NewSetFromStringSlice([]string{"domain", "host_name", "name", "image"})
 		restFilters, err := parseFilters(filterNames, filters)
 		if err != nil {
 			return nil, err


controller/api/apis.yaml
```diff
@@ -2552,6 +2552,28 @@ paths:
...
+          name: body
+          description: Filter data
+          required: true
+          schema:
+            $ref: '#/definitions/RESTAssetsScanReportQuery'
+      responses:
+        '200':
+          description: Success
+          schema:
+            $ref: '#/definitions/RESTHostsScanReportData'
```
holyspectral: I'd keep these in the internal_apis.yaml since these new APIs are not RESTful.

controller/api/apis.yaml
```diff
@@ -3042,6 +3064,28 @@ paths:
       responses:
         '200':
           description: Success
+  /v1/scan/workloads/scan_report:
```
holyspectral: Same here.

controller/rest/scanner.go
```diff
@@ -268,6 +270,209 @@ func handlerScanWorkloadReport(w http.ResponseWriter, r *http.Request, ps httpro
...
+		if skipWorkload(rconf.LastStopAtAsset, wl.Domain, wl.HostName, wl.Name) {
+			continue
+		}
+		vuls, _, _ := cacher.GetVulnerabilityReport(wl.ID, showTag)
+		if len(vuls) == 0 {
+			continue
+		}
+		sort.Slice(vuls, func(i, j int) bool {
+			return vuls[i].Name < vuls[j].Name
+		})
```
holyspectral: minor: slices.SortFunc() is more efficient than sort.Slice.

controller/rest/scanner.go
```diff
@@ -477,6 +682,129 @@ func handlerScanHostReport(w http.ResponseWriter, r *http.Request, ps httprouter
...
+	cached := cacher.GetAllHosts(acc)
+	if len(cached) <= queryStart {
+		restRespSuccess(w, r, resp, acc, login, nil, "Get hosts scan report")
+		return
+	}
+
+	// Sort by host name
+	sort.Slice(cached, func(i, j int) bool {
+		return cached[i].Name < cached[j].Name
+	})
```
holyspectral: minor: slices.SortFunc() is more efficient than sort.Slice.

controller/rest/scanner.go
```diff
@@ -477,6 +682,129 @@ func handlerScanHostReport(w http.ResponseWriter, r *http.Request, ps httprouter
...
+		if rconf.LastStopAtAsset.HostName != "" && host.Name < rconf.LastStopAtAsset.HostName {
+			continue
+		}
+		vuls, _, _ := cacher.GetVulnerabilityReport(host.ID, showTag)
+		if len(vuls) == 0 {
+			continue
+		}
+		sort.Slice(vuls, func(i, j int) bool {
+			return vuls[i].Name < vuls[j].Name
+		})
```
holyspectral: minor: slices.SortFunc() is more efficient than sort.Slice.

controller/rest/scanner.go
```diff
@@ -268,6 +270,209 @@ func handlerScanWorkloadReport(w http.ResponseWriter, r *http.Request, ps httpro
...
+	// Sort by namespace -> host_name -> workload name
+	sort.Slice(cached, func(i, j int) bool {
+		if cached[i].Domain == cached[j].Domain {
+			if cached[i].HostName == cached[j].HostName {
+				return cached[i].Name < cached[j].Name
+			}
+			return cached[i].HostName < cached[j].HostName
+		}
+		return cached[i].Domain < cached[j].Domain
+	})
```
holyspectral: minor: slices.SortFunc() is more efficient than sort.Slice.

controller/rest/scanner.go
```diff
@@ -268,6 +270,209 @@ func handlerScanWorkloadReport(w http.ResponseWriter, r *http.Request, ps httpro
...
+	}
+
+	var resp api.RESTWorkloadsScanReportData
+	resp.WorkloadsScanData = make([]*api.RESTWorkloadScanData, 0)
+	showTag := ""
+	if rconf.ShowAccepted {
+		showTag = api.QueryValueShowAccepted
+	}
+
+	queryStart := rconf.Page * rconf.AssetsPerPage
```
holyspectral: Based on the code, it looks like it will get from cache each time when these two new APIs are called.  This could cause duplicate items during pagination and I'm not sure if this is the best way to implement it. 

E.g., say we have this workload, and each page contains two items, so we should get these 3 pages in the happy case.
- A0, A2
- B0, B2
- C0, C2

However, if the workload A1 joins when we generate the report, we will see duplicate items:
- A0, A2
- (workload A1 joins)
- A2, B0
- B2, C0
- C2

holyspectral: We've had `LastStopAtAsset` and `LastStopAtCVE` as the cursor of cursor-based pagination.  Do we still need the `Page`?   The page-based pagination would be prone to race condition.

controller/rest/scanner.go
```diff
@@ -477,6 +682,129 @@ func handlerScanHostReport(w http.ResponseWriter, r *http.Request, ps httprouter
...
+	}
+
+	var resp api.RESTHostsScanReportData
+	resp.HostsScanData = make([]*api.RESTHostScanData, 0)
+	showTag := ""
+	if rconf.ShowAccepted {
+		showTag = api.QueryValueShowAccepted
+	}
+
+	queryStart := rconf.Page * rconf.AssetsPerPage
```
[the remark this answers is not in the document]
holyspectral: Page will probably be removed.

controller/rest/scanner.go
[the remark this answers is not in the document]
holyspectral: Page will probably be removed.

controller/rest/scanner.go
```diff
@@ -477,6 +682,129 @@ func handlerScanHostReport(w http.ResponseWriter, r *http.Request, ps httprouter
...
+			continue
+		}
+		sort.Slice(vuls, func(i, j int) bool {
+			return vuls[i].Name < vuls[j].Name
+		})
+		for _, cve := range vuls {
+			if rconf.VulScoreFilter != nil {
+				if skipCVE(*rconf.VulScoreFilter, rconf.LastStopAtCVE, cve) {
+					continue
+				}
```
holyspectral: I think the `rconf.LastStopAtCVE` usage here is weird.  Say we have both `LastStopAtAsset` and `LastStopAtCVE` set, then the `LastStopAtCVE` will still filter out CVEs reported in the second Asset, which is not ideal.

controller/api/apis.yaml
holyspectral: On a second thought, I'll keep the API doc in the same place for now.

controller/rest/scanner.go
```diff
@@ -268,6 +278,250 @@ func handlerScanWorkloadReport(w http.ResponseWriter, r *http.Request, ps httpro
...
+
+			ret.ScanData = append(ret.ScanData, &scanData)
+			if len(ret.ScanData) >= maxCveRecords {
+				maxReached = true
+				break outer
+			}
+		}
+	}
+
+	ret.AssetsLeft = len(cachedAssets) - i - 1
```
kyledong-suse: Here the `ret.AssetsLeft` can be -1 if cachedAssets is empty and the for loop never runs.
It's better to add a check, i.e. `if ret.AssetsLeft < 0 {
    ret.AssetsLeft = 0
}`

controller/rest/scanner.go
```diff
@@ -268,6 +278,250 @@ func handlerScanWorkloadReport(w http.ResponseWriter, r *http.Request, ps httpro
...
+	}
+
+	// 1. Get assets, apply filters and sort them.
+	cachedAssets := getAssetFunc()
+
+	var err error
+	cachedAssets, err = filterAssets(rconf.Filters, &rconf.Cursor, cachedAssets)
+	if err != nil {
+		log.WithFields(log.Fields{"error": err}).Warn("Failed to filter assets")
+		// fallthrough
```
kyledong-suse: The `fallthrough` comment is a bit misleading, can we change it to something like `continue with unfiltered assets`.

controller/rest/scanner.go
```diff
@@ -268,6 +278,250 @@ func handlerScanWorkloadReport(w http.ResponseWriter, r *http.Request, ps httpro
...
+	var itemKey api.RESTScanReportCursor
+	maxReached := false
+outer:
+	for i, asset = range cachedAssets {
+		vuls, _, _ := cacheInterface.GetVulnerabilityReport(asset.GetID(), showTag)
+
+		vuls, err = filterAndSortCVE(rconf.VulScoreFilter, vuls)
+		if err != nil {
+			log.WithFields(log.Fields{"error": err}).Warn("Failed to filter vulnerabilities")
+			// fallthrough
```
kyledong-suse: same here

kyledong-suse (commented):
LGTM, just few minor comments.

controller/rest/scanner.go
holyspectral: Thanks!  yeah that's an issue.  As mentioned in the PR description, I don't know if frontend would utilize the field.  I will either remove the field completely or address the issue you found after getting the feedback from frontend team.  Thanks!

controller/rest/scanner.go
holyspectral: yeah what I meant is fallthrough in the context of golang's switch case, but agreed it's probably not the best word choice.  I will just add more context to the warning text.

controller/rest/scanner.go
holyspectral: On a second thought, I will probably just remove the field in this version and add back if needed.

kyledong-suse (approved):
LGTM

PR #2224, opened 2026-03-10
NVSHAS-10244: Neuvector Rancher SSO Authentication error when rancher…

### LINKED ISSUE #2225: Neuvector Rancher SSO Authentication error when rancher helm release name is not "rancher"
#1936

**Environment**
Platform: Rancher

**Describe the bug**
NeuVector Rancher SSO authentication fails when the Rancher helm release name is not "rancher".

**To Reproduce**
1. Install Rancher using helm and make sure the release name is not "rancher"
2. Install NeuVector App inside Rancher
3. Try to navigate to the NeuVector app using the link from inside Rancher UI
4. Received the authentication error

williamlin-suse:
fix #2225 

… helm release name is not 'rancher'

### DIFF
--- a/controller/resource/kubernetes_resource.go
+++ b/controller/resource/kubernetes_resource.go
@@ -2264,6 +2264,10 @@ func initNvRscMapForSSO() {
 }
 
 func IsRancherFlavor() bool {
+	if len(nvRscMapSSO) > 0 {
+		return true
+	}
+
 	if name := os.Getenv("RANCHER_CLUSTER_NAME"); name != "" {
 		initNvRscMapForSSO()
 		return true
@@ -2273,9 +2277,6 @@ func IsRancherFlavor() bool {
 	if _, err := global.ORCH.GetResource(RscTypeNamespace, "", nsName); err != nil {
 		log.WithFields(log.Fields{"namespace": nsName, "err": err}).Info("resource no found")
 		return false
-	}
-	if len(nvRscMapSSO) > 0 {
-		return true
 	}
 	svcnames := utils.NewSetFromStringSlice([]string{"cattle-cluster-agent", "rancher", "rancher-prime"})
 	if svcName, ok := getRancherSvcName(nsName); ok {


controller/resource/kubernetes_resource.go
```diff
@@ -2159,13 +2159,33 @@ func SetK8sVersion(k8sVer string) {
...
+							}
+						}
+					}
+				}
+			}
+		}
+	}
+	return ""
+}
+
```
pohanhuang, line 2186:
```suggestion
func getRancherSvcName(nsName string) (string, bool) {
    objs, err := global.ORCH.ListResource(RscTypeDeployment, nsName)
    if err != nil {
        return "", false
    }

    for _, obj := range objs {
        deploy, ok := obj.(*appsv1.Deployment)
        if !ok {
            continue
        }
        if val, found := findEnvInRancherContainer(deploy); found {
            return val, true
        }
    }
    return "", false
}

func findEnvInRancherContainer(deploy *appsv1.Deployment) (string, bool) {
    for _, c := range deploy.Spec.Template.Spec.Containers {
        if c.Name != "rancher" {
            continue
        }
        for _, env := range c.Env {
            if env.Name == "CATTLE_PEER_SERVICE" {
                return env.Value, true
            }
        }
    }
    return "", false
}
```

Would be more readable, and go Idiomatic

It would be great if we add some unit-test to cover it.

controller/resource/kubernetes_resource.go
```diff
@@ -2198,7 +2218,10 @@ func IsRancherFlavor() bool {
 				share.PERMS_SECURITY_EVENTS_ID:  "securityevents",
 				share.PERM_FED_ID:               "federation",
 			}
-			for _, svcname := range svcnames {
+			if svcName := getRancherSvcName(nsName); svcName != "" {
+				svcnames.Add(svcName)
+			}
```
pohanhuang:
```suggestion
if svcName, ok := getRancherSvcName(nsName); ok {
    svcnames.Add(svcName)
}
```

pohanhuang:
Thanks @holyspectral pointing out, if we want to find the service name, why not just filter the service with container name and other condition?

williamlin-suse:
For `Say if the container name is not rancher anymore?`, we can never predict what will be changed in future Rancher versions including their APIs that NV relies on for SSO support.
It's like how can we be sure the k8s token file path in the pod won't change in the future k8s versions?

Users(QA) can easily create another issue about Rancher flavor not displayed on NV UI -> Assets -> Platforms page when NV is deployed on Rancher-with-custom-name env by manifest(yaml) without Rancher SSO enabled (i.e. they don't have RANCHER_SSO/RANCHER_EP or proposed RANCHER_CLUSTER_NAME env variables in their NV controller pods)
Then what should be do for that?

I think we should not assume NV is always deployed by helm when NV is deployed on Rancher env.

controller/resource/kubernetes_resource.go
```diff
@@ -2159,93 +2159,132 @@ func SetK8sVersion(k8sVer string) {
...
+	}
+
+	nsName := "cattle-system"
+	if _, err := global.ORCH.GetResource(RscTypeNamespace, "", nsName); err != nil {
+		log.WithFields(log.Fields{"namespace": nsName, "err": err}).Info("resource no found")
+		return false
+	}
+	if len(nvRscMapSSO) > 0 {
+		return true
+	}
```
holyspectral:
```suggestion
	if len(nvRscMapSSO) > 0 {
		return true
	}
	
	if name := os.Getenv("RANCHER_CLUSTER_NAME"); name != "" {
		initNvRscMapForSSO()
		return true
	}

	nsName := "cattle-system"
	if _, err := global.ORCH.GetResource(RscTypeNamespace, "", nsName); err != nil {
		log.WithFields(log.Fields{"namespace": nsName, "err": err}).Info("resource no found")
		return false
	}

```

PR #2230, opened 2026-03-11
NVSHAS-10281: suppress log in the response rule is not working

### LINKED ISSUE #2231: suppress log in the response rule is not working
**Environment**
Platform: all
Kubernetes/Platform Version(s): all 

**Describe the bug**
Response rules to suppress logs for activity logs do not work

**To Reproduce**
1. Deploy NV
2. Configure syslog in Settings -> Configuration
3. Create event response rule with group=containers, name=Container.Start, action=suppress log
4. Deploy iperf

**Expected behavior**
There should be no Container.Start log for the iperf received by the configured syslog

**Actual behavior**
The configured syslog still receives Container.Start log for the iperf

williamlin-suse:
fix #2231

### DIFF
--- a/controller/cache/response.go
+++ b/controller/cache/response.go
@@ -314,11 +314,11 @@ func lookup(desc *eventDesc) []actionDesc {
 					continue
 				}
 			} else if rule.Event != desc.event {
-				// The type of activity logs(Container.Start/Container.Stop/Container.Remove/Container.Secured/Container.Unsecured) is
-				// "activity" while the type of event logs is "event".
-				// However, response rules for Event apply to both event logs & activity logs
-				// So for activity logs(desc.event being "activity"), response rules for Event(rule.Event being "event") could still match.
-				if desc.event != share.EventActivity || rule.Event != share.EventEvent {
+				// "event" response rules are designed to match "event" log & "activity" logs
+				// (e.g. Container.Start/Container.Stop/Container.Remove/Container.Secured/Container.Unsecured),
+				// So for activity logs, do not skip them simply because their "event" value is "activity" but not "event"
+				evalActivityLogWithEventRule := (desc.event == share.EventActivity && rule.Event == share.EventEvent)
+				if !evalActivityLogWithEventRule {
 					continue
 				}
 			}
--- a/controller/cache/response_test.go
+++ b/controller/cache/response_test.go
@@ -3,151 +3,168 @@ package cache
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
 	"github.com/neuvector/neuvector/controller/api"
 	"github.com/neuvector/neuvector/share"
 	"github.com/neuvector/neuvector/share/utils"
 )
 
-func TestResponseRuleLookup(t *testing.T) {
-	preTest()
-
-	g1 := share.CLUSGroup{Name: api.AllContainerGroup, CfgType: share.UserCreated}
+func setupTestGroupCache(testWorkloadID string) {
+	g := share.CLUSGroup{Name: api.AllContainerGroup, CfgType: share.UserCreated}
 	groupCacheMap[api.AllContainerGroup] = &groupCache{
-		group:               &g1,
+		group:               &g,
 		usedByPolicy:        utils.NewSet(),
 		usedByResponseRules: utils.NewSet(),
-		members:             utils.NewSet("bc40937a5ab126c1188a2e01f2a52833adfdc4b04f1d22134a4ca5967a64631d"),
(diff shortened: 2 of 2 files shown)


controller/cache/response.go
```diff
@@ -313,8 +313,14 @@ func lookup(desc *eventDesc) []actionDesc {
...
-			} else if rule.Event != desc.event && !(desc.event == share.EventActivity && rule.Event == share.EventEvent) {
-				continue
+			} else if rule.Event != desc.event {
+				// The type of activity logs(Container.Start/Container.Stop/Container.Remove/Container.Secured/Container.Unsecured) is
+				// "activity" while the type of event logs is "event".
+				// However, response rules for Event apply to both event logs & activity logs
+				// So for activity logs(desc.event being "activity"), response rules for Event(rule.Event being "event") could still match.
+				if desc.event != share.EventActivity || rule.Event != share.EventEvent {
+					continue
+				}
```
pohanhuang:
```suggestion
// "event" response rules are designed to also match "activity" logs
// (e.g. Container.Start/Stop/Remove/Secured/Unsecured),
// so don't skip when an activity log meets an event rule.
activityMatchesEventRule := desc.event == share.EventActivity && rule.Event == share.EventEvent
if !activityMatchesEventRule {
    continue
}
```
I believe the above re-write is more straight forward and readable.

controller/cache/response_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"testing"
+
+	"github.com/neuvector/neuvector/controller/api"
+	"github.com/neuvector/neuvector/share"
+	"github.com/neuvector/neuvector/share/utils"
+)
+
+func TestResponseRuleLookup(t *testing.T) {
```
pohanhuang: This test works, but I think it is a bit hard to read because the setup is much larger than the actual assertion.

The main behavior being tested seems to be:
1. prepare material (group/local cache, event, activity)
2. test for many edge cases

May be easier to maintain if we split these into subtests and extract a few helpers, for example:
- helper to build an event rule
- helper to build a level rule
- helper to build an activity eventDesc
- helper to build/reset the response policy cache

That way the test can focus more on the matching behavior instead of the cache/log boilerplate.

controller/cache/response_test.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	desc := eventDesc{id: log.WorkloadID, event: share.EventActivity, name: log.Name, level: log.Level, arg: log}
+
+	//--- activity log matches 1 response rule(for Event)
+	matched := lookup(&desc)
+	if len(matched) != 1 || matched[0].id != 8 || len(matched[0].actions) != 1 || matched[0].actions[0] != share.EventActionSuppressLog {
+		t.Errorf("Unexpected response rule match result [1]")
+		t.Logf("  Expect: 1 rule matched\n")
+		t.Logf("  Actual: rule(s) matched: %v\n", matched)
+	}
```
pohanhuang: For example,
```go
func setupTestGroupCache() {
	g := share.CLUSGroup{Name: api.AllContainerGroup, CfgType: share.UserCreated}
	groupCacheMap[api.AllContainerGroup] = &groupCache{
		group:               &g,
		usedByPolicy:        utils.NewSet(),
		usedByResponseRules: utils.NewSet(),
		members:             utils.NewSet(testWorkloadID),
	}
}

func TestResponseRuleLookup(t *testing.T) {
	preTest()
	defer postTest()
	defer cleanupTestState()
 
	setupTestGroupCache()
 
	t.Run("activity log matches one event rule by name", func(t *testing.T) {
		// Rule 10: matches Controller.Start (won't match)
		// Rule 9:  matches Container.Start but is disabled (won't match)
		// Rule 8:  matches Container.Start and is enabled (should match)
		disabledRule := newEventRule(9, api.EventNameContainerStart)
		disabledRule.Disable = true
 
		localResPolicyCache = buildResPolicyCache([]*share.CLUSResponseRule{
			newEventRule(10, api.EventNameControllerStart),
			disabledRule,
			newEventRule(8, api.EventNameContainerStart),
		})
 
		desc := newActivityDesc("Container.Start", "Info")
		matched := lookup(&desc)

        require.Len(t, matched, 1, "expected exactly one matched rule")
        assert.Equal(t, uint32(8), matched[0].id, "unexpected matched rule ID")
        assert.Len(t, matched[0].actions, 1, "expected exactly one action")
        assert.Equal(t, share.EventActionSuppressLog, matched[0].actions[0], "unexpected action")
	})
...
}

```

controller/cache/response.go
williamlin-suse: There is no such logic about "skip when an activity log meets an event rule."
`activityMatchesEventRule` is misleading

PR #2243, opened 2026-03-19
feat: support os scan status on both repo and registry scan

pohanhuang:
### Summary
Support the os_scan_status for user to know if their os package is scanned or not.
- `unknown` represent os has not been scanned, require user to re-scan to know
- `support` represent os has been scanned
- `unsupport` represent os was not been scanned, our scanner does not support such os in our scan db

### Issue address
https://github.com/neuvector/neuvector/issues/2247

### Scanner PR
https://github.com/neuvector/scanner/pull/335

### CI Scan Results
```bash
while true; do
  http_code=$(curl -sk -o /tmp/nv-redis-scan.json -w '%{http_code}' \
    -H 'X-Auth-Apikey: l:29PLSvAgiFv97Q4y9FIoI#QaG+2Ga&a8YBk3&jSegTBkDIUsAJbbmeVS8IC7wabz' \
    -H 'Content-Type: application/json' \
    -X POST \
    'https://10.115.49.227:32764/v1/scan/repository' \
    -d '{
      "request": {
        "registry": "https://index.docker.io",
        "username": "",
        "password": "",
        "repository": "library/redis",
        "tag": "latest",
        "scan_layers": true
      }
    }')
  [ "$http_code" = "304" ] || break
  sleep 2
done

jq -r '.report.os_scan_status' /tmp/nv-redis-scan.json

will get => supported

```
#### Upgrade from older version will get `unknown`
upgrade-from-old-version

#### Install from scratch will get `support` or `unsupport`
new

### DIFF
--- a/share/scan/scan_utils.go
+++ b/share/scan/scan_utils.go
@@ -80,17 +80,6 @@ var scanErrString = []string{
 	share.ScanErrorCode_ScanErrSignatureScanError:    "signature scan error",
 }
 
-func OSScanStatusToString(status share.OSScanStatus) string {
-	switch status {
-	case share.OSScanStatus_OSScanStatusSupported:
-		return "supported"
-	case share.OSScanStatus_OSScanStatusUnsupported:
-		return "unsupported"
-	default:
-		return "unknown"
-	}
-}
-
 type CacheRecord struct {
 	Layer   string    `json:"layerID,omitempty"`
 	Size    uint64    `json:"size,omitempty"`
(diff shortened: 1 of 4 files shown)


share/scan/scan_utils.go
```diff
@@ -80,6 +80,17 @@ var scanErrString = []string{
 	share.ScanErrorCode_ScanErrSignatureScanError:    "signature scan error",
 }
 
+func OSScanStatusToString(status share.OSScanStatus) string {
```
holyspectral: nit: Not sure if you can use [String() here](https://github.com/pohanhuang/neuvector/blob/2c04492f136d67b14b2860776f85a78d5801fffa/share/scan.pb.go#L166) by renaming those enums in proto file.  If so, you don't have to write a converter like this.

pohanhuang: You got it, will use the String() instead, thanks.

PR #2257, opened 2026-03-25
issue-2235: feat: add a column containing image digest in generated reports for containers

williamlin-suse:
[2235](https://github.com/neuvector/neuvector/pull/2257)

## Description

A container image tag is mutable information,  
In order to clearly identify compromised images we also need to know the image digest.

**Additional context**
Add a new field `workload_image_digest` in the response of `POST(/v1/scan/workloads/scan_report)` for front-end to add a new column for workload's image digest.

<!DOCTYPE html>

<html>
<body>

image_name | tags | digest | workload_name | namespace | host_name | CVE name | link | severity | score | score_v3 | package_name | package_version | fixed_version | description | feed_rating | file_name | vectors | vectors_v3 | in_base_image | published_timestamp | last_modified_timestamp
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
quay.io/calico/apiserver | v3.31.3 | insert image digest here | k8s_calico-apiserver_calico-apiserver-b9ddbd7b6-vt7qj_calico-system_d925ced4-19ef-4336-934e-6d2f3a8d4ace_0 | calico-system | K8-Master | CVE-2024-7598 | https://github.com/kubernetes/kubernetes/issues/126587 | Low | 1 | 3.1 | go:k8s.io/kubernetes | 1.33.6 |   | Network restriction bypass via race condition during namespace termination | Low | code/apiserver |   | CVSS:3.1/AV:A/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N |   | Jan 20 1970 20:01:30 | Jan 20 1970 20:01:45
quay.io/calico/apiserver | v3.31.3 | insert image digest here | k8s_calico-apiserver_calico-apiserver-b9ddbd7b6-vt7qj_calico-system_d925ced4-19ef-4336-934e-6d2f3a8d4ace_0 | calico-system | K8-Master | CVE-2025-61726 | http://people.ubuntu.com/~ubuntu-security/cve/CVE-2025-61726 | High | 7 | 7.5 | go:stdlib | 1.24.11 | >=1.24.12;>=1.25.6 | The net/url package does not set a limit on the number of query parameters in a query.While the maximum size of query parameters in URLs is generally limited by the maximum request header size, the net/http.Request.ParseForm method can parse large URL-encoded forms. Parsing a large form containing many unique query parameters can cause excessive memory consumption. | High | code/apiserver |   | CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H |   | Jan 21 1970 03:33:51 | Jan 21 1970 03:46:43

</body>

</html>

### DIFF
--- a/share/container/types.go
+++ b/share/container/types.go
@@ -143,7 +143,7 @@ type ContainerMetaExtra struct {
 	ContainerMeta
 	ImageID          string
 	ImageDigest      string
-	ImageDigestExtra []string // on docer runtime env, there is no guarantee we we only get one digest for an image. so store the 2nd~ digests here
+	ImageDigestExtra []string // on docker runtime env, there is no guarantee we we only get one digest for an image. so store the 2nd~ digests here
 	ImgCreateAt      time.Time
 	Author           string
 	Privileged       bool


share/container/docker.go
```diff
@@ -671,20 +677,24 @@ func isSha256String(input string) bool {
...
 		}
 
 		image, err := d.client.ImageInspect(context.Background(), imageID)
 		if err == nil && len(image.RepoTags) > 0 {
 			for _, repo := range image.RepoTags {
 				repoTag = repo // report the last one
 			}
+			for _, digest := range image.RepoDigests {
+				imageDigest = digest // report the last one
+			}
```
holyspectral: I'm slightly concerned about how we get the image digest from docker daemon here. 

1. If the same image is downloaded from different repos by container runtime, the repo digest won't always be synced with the workload.  Say if one workload uses open source version and another workload uses rancher mirrored version, we will see them not synced.
2. This is also implemented based on an assumption that the image digest on different repository will be the same, but this is never guaranteed.

Given these issues could only happen in docker, if docker support is a priority, we should consider supporting multiple repo digest in this field and not strip the repo part.

share/container/docker.go
```diff
@@ -225,11 +226,15 @@ func (d *dockerDriver) GetContainer(id string) (*ContainerMetaExtra, error) {
 		return nil, err
 	}
 
+	repoTag, imageDigest := d.getImageRepoTag(info.Image, info.Config.Image)
+	if idx := strings.Index(imageDigest, "@sha256:"); idx > 0 {
+		imageDigest = imageDigest[idx+len("@sha256:"):]
+	}
```
holyspectral: You can use [strings.Cut()](https://pkg.go.dev/strings#Cut)

share/container/docker.go
williamlin-suse: Your concern is solid. 

Actually when it's docker, the response payload of POST(/v1/scan/workloads/scan_report) is like:
```
{
    "cursor": { ... },
    "scan_data": [
        {
            "host_name": "wlin-k8sworker42",
            "workload_name": "k8s_iperfserver_iperfserver-758cfc9b49-zpsgq_demo_c699f28e-f833-4adf-92f9-9bdc39b8211b_0",
            "workload_domain": "demo",
            "workload_image": "sha256:e8b5ab3eed38ffa33a953544b8921f17f04adc5d91302e2c061f285843e87a62",
            "workload_image_id": "e8b5ab3eed38ffa33a953544b8921f17f04adc5d91302e2c061f285843e87a62",
            "workload_image_digest": "",
            "name": "CVE-2016-10009",
            "score": 7.5,
            "severity": "High",
            ........
        },
        ......
    ]
}
```
where the `workload_image`/`workload_image_id`/`workload_image_digest` are from `object/workload/{node_id}/{workload_id}`
`workload_image_id` is actually the image digest sha for `docker pull`

In this scenario we don't actually have the image repo(name&tag) for the container:
image

and current logic causes the downloaded workloads scan report to have incorrect content like:
```
image_name, tags, workload_name, .......
sha256, e8b5ab3eed38ffa33a953544b8921f17f04adc5d91302e2c061f285843e87a62, k8s_iperfserver_iperfserver-758cfc9b49-zpsgq_demo_c699f28e-f833-4adf-92f9-9bdc39b8211b_0, ......
```

williamlin-suse: Sorry I need to use `git push -f` to push the new commit after the PR for switching from `docker/docker` to `moby/moby` was merged 10 days ago (I cannot make it work to sync the new commits in `main` to my branch for this PR)

I changed the code so that the repo part is stripped from the repo digest only when there is one RepoDigest on docker env.
If there are multiple image digests found and we cannot tell which one is for the container, all of them are collected without any change. 
It means workloads' scan report may get a container entry with multiple digests

share/container/types.go
```diff
@@ -141,21 +141,22 @@ type ContainerMeta struct {
...
-	StartedAt   time.Time
-	FinishedAt  time.Time
-	MemoryLimit int64
-	CPUs        string
-	ProxyMesh   bool
-	Sidecar     bool
-	RunAsRoot   bool
+	ImageID          string
+	ImageDigest      string
+	ImageDigestExtra []string // on docer runtime env, there is no guarantee we we only get one digest for an image. so store the 2nd~ digests here
```
holyspectral:
```suggestion
	ImageDigestExtra []string // on docker runtime env, there is no guarantee we we only get one digest for an image. so store the 2nd~ digests here
```

share/container/docker.go
```diff
@@ -677,33 +697,44 @@ func isSha256String(input string) bool {
...
 
 		image, err := d.client.ImageInspect(context.Background(), imageID)
-		if err == nil && len(image.RepoTags) > 0 {
-			for _, repo := range image.RepoTags {
-				repoTag = repo // report the last one
+		if err == nil {
+			if len(image.RepoTags) > 0 {
+				for _, repo := range image.RepoTags {
+					repoTag = repo // report the last one
+				}
```
holyspectral, line 717: I see this is because of backward compatibility.  I'm fine with keeping it as it is.

Issue #2260, opened 2026-03-27
Add Critical CVE Severity to support CVSS v3 scores 9.0-10.0


williamlin-suse:
Case: [NVSHAS-8242](https://jira.suse.com/browse/NVSHAS-8242)

Add CRITICAL labeling throughout the back-end to include CRITICAL in addition to HIGH and MEDIUM. 
Use CVSS v3 levels to apply correct criticality per CVSS v3 specs. 
Always use CVSS v3 values by default.

williamlin-suse:
Changes for criteria key in admission control rules & response rules:

- For admission control rules:
  -- new criteria keys:
      1. cveCriticalCount # count of critical CVEs
      2. cveHighCountNoCritical  # count of high CVEs only
      3. cveCriticalWithFixCount   # count of `critical with fix` CVEs
      4. cveHighWithFixCountNoCritical  # count of `high with fix` CVEs only

  -- existing related criteria keys (for backward compatibility)
      1. cveHighCount  # count of critical & high CVEs
      2. cveHighWithFixCount  # count of `critical with fix` & `high with fix` CVEs

UI will display these criteria keys based on their meanings
      
- For response rules:
  -- new criteria keys:
      1. cve-critical # for count of critical CVEs
      2. cve-high-only  # for count of high CVEs only
      3. cve-critical-with-fix   # for count of `critical with fix` CVEs
      4. cve-high-only-with-fix  # for count of `high with fix` CVEs only

  -- existing related criteria keys (for backward compatibility)
      1. cve-high  # for count of critical & high CVEs
      2. cve-high-with-fix  # for count of `critical with fix` & `high with fix` CVEs

venkateshjayagopal:
@williamlin-suse Do we have an open ticket  for documentation update?.
https://open-docs.neuvector.com/policy/admission

williamlin-suse:
> [@williamlin-suse](https://github.com/williamlin-suse) Do we have an open ticket for documentation update?. https://open-docs.neuvector.com/policy/admission

no

PR #2274, opened 2026-04-13
NVSHAS-8242: Add Critical CVE Severity to support CVSS v3 scores 9.0-…

williamlin-suse:
https://github.com/neuvector/neuvector/issues/2260

- For admission control rules:
  -- new criteria keys:
      1. cveCriticalCount # count of critical CVEs
      2. cveHighCountNoCritical  # count of high CVEs only
      3. cveCriticalWithFixCount   # count of `critical with fix` CVEs
      4. cveHighWithFixCountNoCritical  # count of `high with fix` CVEs only

  -- existing related criteria keys (for backward compatibility)
      1. cveHighCount  # count of critical & high CVEs
      2. cveHighWithFixCount  # count of `critical with fix` & `high with fix` CVEs

UI will display these criteria keys based on their meanings
      
- For response rules:
  -- new criteria keys:
      1. cve-critical # for count of critical CVEs
      2. cve-high-only  # for count of high CVEs only
      3. cve-critical-with-fix   # for count of `critical with fix` CVEs
      4. cve-high-only-with-fix  # for count of `high with fix` CVEs only

  -- existing related criteria keys (for backward compatibility)
      1. cve-high  # for count of critical & high CVEs
      2. cve-high-with-fix  # for count of `critical with fix` & `high with fix` CVEs


controller/cache/admission.go
```diff
@@ -2021,33 +2021,6 @@ func (m CacheMethod) WaitUntilApiPathReady() bool {
 	return false
 }
 
-func (m CacheMethod) IsImageScanned(c *nvsysadmission.AdmContainerInfo) (bool, int, int) {
```
pohanhuang, line 2024: When does NeuVector need to call this function? From UI side?

controller/cache/response.go
```diff
@@ -221,41 +255,27 @@ func matchConditions(desc *eventDesc, conds []share.CLUSEventCondition) bool {
...
+		case share.EventCondTypeCVEHighOnly:
+			count, err := strconv.Atoi(d.CondValue)
+			if err != nil || desc.cve_high < count {
 				return false
 			}
-			if len(ss) >= 1 && len(desc.cve_high_fixed_info) < cveCountInRule {
-				// it's configured like: ( high_vul_with_fix:X ) meaning "# of high vul(with fix) >= X"
+		case share.EventCondTypeCVEHigh:
+			count, err := strconv.Atoi(d.CondValue)
+			if err != nil || (desc.cve_high+desc.cve_critical) < count {
```
pohanhuang, line 269: Why do we count critical in high cve as well? I know we should handle backward compatible, do we open new criteria for user? 

@holyspectral what do you think? 
NeuVector used to mix the cve count with High and Critcal, since we separate them out in this PR, do you think we should create new rules for user to maintain backward compatible?

controller/rest/scanner.go
```diff
@@ -605,7 +605,8 @@ func handlerScanImageSummary(w http.ResponseWriter, r *http.Request, ps httprout
 		w := wl.ScanSummary
 
 		old, ok := imageMap[wl.ImageID]
-		if !ok || old.HighVuls < w.HighVuls || (old.HighVuls == w.HighVuls && old.MedVuls < w.MedVuls) {
+		if !ok || old.CriticalVuls < w.CriticalVuls || (old.CriticalVuls == w.CriticalVuls && old.HighVuls < w.HighVuls) ||
+			(old.CriticalVuls == w.CriticalVuls && old.HighVuls == w.HighVuls && old.MedVuls < w.MedVuls) {
```
pohanhuang: maybe add this would be more readable
```go
func isMoreSevere(incoming, existing api.RESTScanImageSummary) bool {
    if incoming.CriticalVuls != existing.CriticalVuls {
        return incoming.CriticalVuls > existing.CriticalVuls
    }
    if incoming.HighVuls != existing.HighVuls {
        return incoming.HighVuls > existing.HighVuls
    }
    return incoming.MedVuls > existing.MedVuls
}
```

controller/cache/response.go
```diff
@@ -195,6 +195,40 @@ func responseRuleConfigUpdate(nType cluster.ClusterNotifyType, key string, value
 	}
 }
 
+func matchCVEWithFixConditions(condValue string, cve_fixed_info []scanUtils.FixedVulInfo) bool {
+	match := true
+	ss := strings.Split(condValue, "/")
+	cveCountInRule, err := strconv.Atoi(ss[0]) // high vul with fix count configured in response rule
```
pohanhuang: Since this function is open for high / critical cve
```suggestion
	cveCountInRule, err := strconv.Atoi(ss[0]) // vul with fix count configured in response rule
```

controller/cache/response.go
```diff
@@ -195,6 +195,40 @@ func responseRuleConfigUpdate(nType cluster.ClusterNotifyType, key string, value
...
+			dur := time.Since(time.Unix(info.PubTS, 0))
+			if dur.Hours() >= hoursReported { // found high cve that is reported before <daysReported> days ago
+				reportedBeforeNDays += 1
+			}
+		}
+		if reportedBeforeNDays == 0 {
+			return false
+		}
+	}
+	return match
```
pohanhuang: Current implementation does not count this, but simply return true, which is a bug
```suggestion
return reportedBeforeNDays > cveCountInRule
```

controller/cache/response.go
```diff
@@ -195,6 +195,40 @@ func responseRuleConfigUpdate(nType cluster.ClusterNotifyType, key string, value
 	}
 }
 
+func matchCVEWithFixConditions(condValue string, cve_fixed_info []scanUtils.FixedVulInfo) bool {
+	match := true
```
pohanhuang: remove this match, because the function does not use this here.
```suggestion
```

controller/cache/response.go
```diff
@@ -195,6 +195,40 @@ func responseRuleConfigUpdate(nType cluster.ClusterNotifyType, key string, value
 	}
 }
 
+func matchCVEWithFixConditions(condValue string, cve_fixed_info []scanUtils.FixedVulInfo) bool {
```
pohanhuang: rename with cveFixedInfo to match go idiomatic
```suggestion
func matchCVEWithFixConditions(condValue string, cveFixedInfo []scanUtils.FixedVulInfo) bool {
```

controller/cache/response.go
```diff
@@ -195,6 +195,40 @@ func responseRuleConfigUpdate(nType cluster.ClusterNotifyType, key string, value
...
+	}
+	if len(ss) == 2 {
+		// it's configured like: ( high_vul_with_fix:X/Y ) meaning "# of (high vul that are reported Y days ago AND have fix) >= X"
+		daysReported, err := strconv.Atoi(ss[1])
+		if err != nil {
+			return false
+		}
+		hoursReported := float64(24 * daysReported) // "reported before N hours" that is configured in response rule
+		reportedBeforeNDays := 0
+		// calculate how many high cve(with fix) that are reported before <daysReported> days
```
pohanhuang:
```suggestion
		// calculate how many cve(with fix) that are reported before <daysReported> days
```

controller/cache/response.go
```diff
@@ -195,6 +195,40 @@ func responseRuleConfigUpdate(nType cluster.ClusterNotifyType, key string, value
...
+func matchCVEWithFixConditions(condValue string, cve_fixed_info []scanUtils.FixedVulInfo) bool {
+	match := true
+	ss := strings.Split(condValue, "/")
+	cveCountInRule, err := strconv.Atoi(ss[0]) // high vul with fix count configured in response rule
+	// get settings from response rule
+	if err != nil {
+		return false
+	}
+	if len(ss) >= 1 && len(cve_fixed_info) < cveCountInRule {
+		// it's configured like: ( high_vul_with_fix:X ) meaning "# of high vul(with fix) >= X"
```
pohanhuang:
```suggestion
		// it's configured like: ( high_vul_with_fix:X ) meaning "# of vul(with fix) >= X"
```

pohanhuang (changes requested):
Thanks for jumping into this issue.
I recalled there is many place is using isCriticalMatch, do you know what this is for? should we fix this in this PR?

williamlin-suse:
> Thanks for jumping into this issue. I recalled there is many place is using isCriticalMatch, do you know what this is for? should we fix this in this PR?

isCriticalMatch actually means the matched rule gets allow/deny action(not including deny/monitor), not related to CVE severity.
Will rename isCriticalMatch to isNonMonitorMatched to avoid confusion

controller/cache/admission.go
williamlin-suse: func (m CacheMethod) IsImageScanned() is obsolete and not called at all.
So remove it in this PR

controller/cache/response.go
williamlin-suse: Please see the top PR comment.
`share.EventCondTypeCVEHigh(cve-high)` is for backward compatibility
`share.EventCondTypeCVECritical(cve-critical)` & `share.EventCondTypeCVEHighOnly(cve-high-only)` are new criteria (lines 258 ~ 267)

controller/cache/response.go
williamlin-suse: All the return statements befoer the last return statement in matchCVEWithFixConditions() are for not-matched result.
So when it reaches the last statement, it should return `true` (i.e. the value of `match`), not `reportedBeforeNDays > cveCountInRule`

For example: 
When the response rule is like `cve-critical-with-fix:3` (meaning rule matched when `count of critical cve with fix >= 3` ) and there 3 critical CVEs with fix, `reportedBeforeNDays > cveCountInRule` returns false but it should return true

controller/cache/response.go
williamlin-suse: sure

controller/cache/response.go
williamlin-suse: ok

controller/cache/response.go
williamlin-suse: ok

controller/rest/scanner.go
williamlin-suse: code changed as advised

controller/cache/response.go
williamlin-suse: The comment is corrected with real example

controller/cache/response.go
williamlin-suse: ok

controller/cache/response.go
```diff
@@ -195,6 +195,39 @@ func responseRuleConfigUpdate(nType cluster.ClusterNotifyType, key string, value
...
+			dur := time.Since(time.Unix(info.PubTS, 0))
+			if dur.Hours() >= hoursReported { // found cve that is reported before <daysReported> days ago
+				reportedBeforeNDays += 1
+			}
+		}
+		if reportedBeforeNDays == 0 {
+			return false
+		}
+	}
+	return true
```
pohanhuang, line 228: I am a bit confused. As long as any single CVE was reported more than Y days ago, the condition passes — regardless of X. 

For example, with `cve-critical-with-fix:3/7`, if only 1 CVE is older than 7 days, it still returns true instead of false

Does this behavior expected?

pohanhuang (changes requested):
Found a logic issue

controller/cache/response.go
williamlin-suse: Good catch! This is indeed a big.
Fixed in https://github.com/neuvector/neuvector/pull/2274/changes/e65d4edf7e391b1c31033e31469428ecd66262da

pohanhuang (approved):
LGTM!

PR #2280, opened 2026-04-23
ci: automatically update golang version

holyspectral:
## Description

This PR creates an updatecli job to automatically adopt new golang version for us. 

### Checklist

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -17,7 +17,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5.6.0
         with:
-          go-version: stable
+          go-version-file: "go.mod"
       - name: Install dependencies
         run: |
           sudo apt update && sudo apt install -y libpcre3-dev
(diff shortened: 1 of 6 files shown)


kyledong-suse (approved):

pohanhuang (approved):

pohanhuang (changes requested):
nit: This PR only ensure the go version bump.
Should we run `go mod tidy` or ` go mod vendor` in after action to keep the go libs up-to-date?

Like the example here:
https://www.updatecli.io/docs/plugins/resource/gomod/

holyspectral:
@pohanhuang Thanks for the comment!  I just tested with [this PR](https://github.com/holyspectral/neuvector/pull/15).  It seems like the `go mod tidy` and `go mod vendor` lead to no change.  I'd say the current code is good unless we're upgrading those dependencies. 

The only situation I can think of from updating golang version is that some dependencies don't support the newer golang version yet, so we may have to upgrade those dependencies.  Then we would end up with some changes in `vendor/` folder.  However, I think this is something that we have to intervene manually anyway.

pohanhuang (approved):

PR #2282, opened 2026-04-25
dpi/kafka: support modern Kafka API key/version range (KIP-482)

boraozturk:
## Summary

The Kafka DPI parser was written against Kafka 0.10 (api_key `0..16`, api_version `0..4`). Anything newer is currently rejected by the request-header sanity check, or mis-parsed as a legacy v0/v1 body. As a result modern clients (Java 2.4+, librdkafka, sarama, kafka-python 2.x, ...) that emit higher api_versions or KIP-482 flexible-version requests do not get identified as Kafka, and sessions surface as `Any/HTTP` in the Manager UI.

This PR teaches the parser about the modern range without trying to deep-parse every per-API flexible-version body.

### Changes

- Raise the front-gate caps to `KAFKA_API_KEY_MAX = 90` and `KAFKA_API_VERSION_MAX = 32` in both `kafka_parser` and `kafka_midstream`. Sized to keep headroom for future `ApiKeys` without re-rejecting modern `ApiVersions`/`CreateTopics`/`SaslAuthenticate`/...

- Introduce a `kafka_max_parsed_version[]` table + `kafka_can_deep_parse()` so the existing deep handlers (`check_produce_request`, `check_fetch_request`, ...) only run on `(api_key, version)` tuples whose v0/v1 body layout they actually understand. Higher versions and every flexible-version request fall through with `checked_request=false`; correlation-ID matching in `check_corr_id_valid()` then validates the session — exactly like the existing default branch already does for unknown `api_keys`.

- Extend `kafka_midstream` the same way. Legacy versions still attempt the deep parse for stronger evidence, but flexible/new requests are accepted on header sanity alone via a new helper `kafka_request_header_plausible()` that validates `size + api_key + api_version + a printable/nullable client_id`. The `client_id` field has the same int16-prefixed wire format in v1 and v2 request headers, so this works for KIP-482 traffic without needing a varint/compact-string body parser.

### Why this approach

The parser's job here is session identification, not deep policy enforcement on Kafka semantics — `kafka_data_t` only carries `valid_request`, `checked_request`, and correlation IDs (no extracted topics flow back into NV). So writing flexible-version body parsers for every API would be invasive churn with no detection benefit. Letting unsupported `(api_key, version)` tuples fall through to correlation-ID matching is consistent with how the file already handles unknown api_keys today.

### Tested

- Built into a `linux/amd64` enforcer image and deployed on a Rancher/k8s cluster running Kafka 3.x clients (Java + .NET). Sessions on broker port 9092 are now identified as `Kafka` in the Manager UI for both new connections and mid-stream attaches; before this PR the same traffic surfaced as `Any/HTTP`.
- Legacy clients (Kafka 0.10, api_version ≤ 4) keep the original deep-parse code path unchanged.

## Test plan

- [x] CI: `./unitest.sh` passes
- [x] CI: `golangci-lint` passes (no Go changes)
- [x] Manual: legacy Kafka client (api_version ≤ 4) still detected via deep parse
- [x] Manual: Kafka 3.x Java client (Produce v9+, Fetch v12+) detected via correlation-ID match
- [x] Manual: mid-stream attach for a long-lived modern Kafka session detected via header sanity

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dp/dpi/parsers/dpi_kafka.c
+++ b/dp/dpi/parsers/dpi_kafka.c
@@ -3,6 +3,11 @@
 
 #include "dpi/dpi_module.h"
 
+/* Kafka request API keys. Only the legacy block (0..16) is deep-parsed
+ * against a known v0/v1 body layout. Newer keys and any "flexible version"
+ * (KIP-482) of the legacy keys are accepted but not deep-parsed; session
+ * identification then falls back to correlation-ID matching in
+ * check_corr_id_valid(). See kafka_can_deep_parse() below. */
 #define    PRODUCE                  0
 #define    FETCH                    1
 #define    LIST_OFFSETS             2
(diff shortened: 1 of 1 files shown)


boraozturk:
cc @williamlin-suse @holyspectral — would appreciate your review when you get a chance. cc @gfsuse @kyledong-suse for DPI parser context.

Issue #2286, opened 2026-04-29
Include CVE DB Version in NeuVector Vulnerability Report


williamlin-suse:
Case: [NVSHAS-10298](https://jira.suse.com/browse/NVSHAS-10298)

1. In vulnerabilities report, show image digest of each scanned image for each CVE
2. In asset report, show CVE DB version & CVE DB creation time & image digest for each scanned image

williamlin-suse:
A new field `include_no_vul_assets` (in the same object as `last_modified_timestamp`) is supported for the request body of `POST(/v1/assetvul)` to include those images-without-vulnerability in the assets view report.

Example of the response payload of 
1. `GET(/v1/vulasset)`: 
( new fields: `.vulnerabilities[].images[].digest` )
```
{
    "vulnerabilities": [
        {
            "name": "CVE-2026-40164",
            ..........
            "score_v3": 7.5,
            "vectors_v3": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H",
            "published_timestamp": 1776125767,
            "last_modified_timestamp": 1776439573,
            "feed_rating": "medium",
            "images": [
                {
                    "id": "da2d5bf83da3cb05769f0bdd4e506f7255d10f02134bce8a45c76058195d30eb",
                    "display_name": "nvlab/iperf:ubuntu",
                    "policy_mode": "",
                    "digest": "sha256:cde84a24e39752874f06bed45cb482eab24f6db4bf12e3f0f2e13a7fbc7afb2e"
                }
            ]
        },
```

2. `POST(/v1/assetvul)`:  
( new fields: `.images[].digest`, `.images[].cvedb_version`, `.images[].cvedb_create_time` & `.no_vul_images` )
```
{
    "workloads": [....],
    "nodes": [....],
    "platforms": [....],
    "images": [
        {
            "id": "91c024c03eecdd577620cc4fd8a15c74a33b7188c9dc6a16209b145c7e2b230c",
            "name": "nvlab/iperf:C08210",
            "digest": "sha256:5c5ed201be3ab5e8ffb4a948af0c18cefc675311611ad538ce66cde7f0f8ade5",
            "critical": 63,
            "high": 40,
            "medium": 38,
            "low": 1,
            "vulnerabilities": [
                ........,
            ],
            "cvedb_version": "4.151",
            "cvedb_create_time": "2026-04-30T21:01:57Z"
        },
        {....},
    ],
[...]

PR #2295, opened 2026-04-30
NVSHAS-10298: Include CVE DB Version in NeuVector Vulnerability Report

williamlin-suse:
https://github.com/neuvector/neuvector/issues/2286

### DIFF
--- a/controller/rest/vulasset.go
+++ b/controller/rest/vulasset.go
@@ -580,6 +580,7 @@ func combineQueryFilter(r *http.Request) (*db.VulQueryFilter, error) {
 	}
 	queryFilter.Filters = vulQF.Filters
 	queryFilter.Filters.LastModifiedTime = qf.Filters.LastModifiedTime
+	queryFilter.Filters.IncludeNoVulAssets = qf.Filters.IncludeNoVulAssets
 
 	return queryFilter, nil
 }
--- a/controller/rest/vulasset.go
+++ b/controller/rest/vulasset.go
@@ -530,9 +530,14 @@ func getAssetViewSession(w http.ResponseWriter, r *http.Request) {
 	elapsed := time.Since(start)
 	queryStat.PerfStats = append(queryStat.PerfStats, fmt.Sprintf("1/2, get vul from db, took=%v", elapsed))
 
+	var noVulImageIDs []string
+	if queryFilter.Filters.IncludeNoVulAssets {
+		noVulImageIDs = allowed[db.AssetImage].Difference(utils.NewSetFromStringSlice(assetsMap[db.AssetImage])).ToStringSlice()
+	}
+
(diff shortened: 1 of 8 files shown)


db/assetvuls.go
```diff
@@ -365,6 +372,41 @@ func getImageAssetView(vulMap map[string]*DbVulAsset, assets []string, queryFilt
...
+func getNoVulImageAssetView(assets []string, queryFilter *VulQueryFilter) ([]*api.RESTNoVulImageAsset, error) {
+	records := make([]*api.RESTNoVulImageAsset, 0)
+
+	if len(assets) == 0 {
+		return records, nil
+	}
+
+	columns := []interface{}{"assetid", "name", "I_digest", "cvedb_version", "cvedb_createtime"}
+	dialect := goqu.Dialect("sqlite3")
+	statement, args, _ := dialect.From(Table_assetvuls).Select(columns...).Where(buildWhereClauseForImage(assets, queryFilter.Filters)).Prepared(true).ToSQL()
```
holyspectral: Please handle the error from `ToSQL()`.

db/assetvuls.go
```diff
@@ -365,6 +372,41 @@ func getImageAssetView(vulMap map[string]*DbVulAsset, assets []string, queryFilt
...
+
+	rows, err := dbHandle.Query(statement, args...)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	poolSize := queryFilter.ThreadCount
+	pool := pond.New(poolSize, 0, pond.MinWorkers(poolSize))
```
holyspectral: Perhaps you don't need the pool here since all actions are done inline?

holyspectral (commented):
This PR seems to do two things:
1. Include CVE db version.
2. Allow API caller to return images scanned but without vulnerabilities. 

I'd separate them into two commits for easier review next time. 

Other than that, just two small issues.

controller/rest/vulasset.go
kyledong-suse, line 584: Should we also add an assignment for `IncludeNoVulAssets` here? Otherwide, line 534 will not be able to get the correct value.

db/assetvuls.go
williamlin-suse: Changed code to handle all calls of `ToSQL()`

db/assetvuls.go
williamlin-suse: Sure. Removed the pool usage in Handle the error from `getNoVulImageAssetView()`

controller/rest/vulasset.go
williamlin-suse: Confirmed with front-end  that the new option is in the request body, not URL query parameter.
So move the logic from `db/GetVulnerabilityQuery()` to `controller/rest/combineQueryFilter()` as advised

kyledong-suse (approved):
LGTM, thanks!

PR #2383, opened 2026-05-19
feat: move readiness probe from exec to httpget

### LINKED ISSUE #2227: UBI controller images don't work in newer OpenShift environment
**Environment**
Kubernetes/Platform Version(s): OpenShift

**Describe the bug**

When NV UBI images are used in newer OpenShift platform, e.g., 4.19, controller's own readiness probe would be flagged by enforcer as not allowed commands

holyspectral:
## Description

This PR implements an alternative readiness probe to replace the existing readiness probe.  

This is to address the issue where [NV Protect can't handle readiness probe correctly](https://github.com/neuvector/neuvector/blob/9ae6ca2349ce2e117a39707689655a4875731be5/agent/probe/process.go#L3072) when the base image is ubi9 image.

In the ubi9 image, `/usr/bin/cat` is actually a shell script pointing to `/usr/bin/coreutils --coreutils-prog-shebang=cat`.  That's why the executable is not allowed by the NV allow list.  

However, to allow `/usr/bin/coreutils` correctly using an allow list, there are a few challenges:

1. If we check only `proc.path == "/usr/bin/coreutils"`, we have to make sure that only `cat` is allowed.  However, coreutils can be used for `cp`, `mv` and other commands, which opens a relatively big attack surface. 
2. Additionally, we have to verify what file `cat` can access.  The current implementation relies on `proc.cmdline` to check what file the `/usr/bin/cat` accesses.  This is actually not reliable as `proc.cmdline` is prone to race condition.  I can reproduce the false positives from the readiness probe even with BCI images as the base image. 
3. Besides, even if we figure out how to allow the `/usr/bin/cat` on UBI9, we still have to define two allow list, one for normal NV release and one for UBI images.   

With these, it makes more sense to just switch to a `httpGet` in readiness probe. This comes with these benefits:

- We address the false positives from `/usr/bin/cat`.
- Consistent behavior in OpenShift scenario and k8s secenario.
- Potentially less attack surface in the future since `/usr/bin/cat` could be changed to not allowed.  Having said that, the allow list is kept for backward compatibility reason for now, so readiness probes of old controllers won't be affected during controller rolling update.

The link of helm chart PR: https://github.com/neuvector/neuvector-helm/pull/575

Fix #2227 

## Test

Install neuvector and verify if controller becomes ready. 

### Checklist

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -546,16 +546,17 @@ func main() {
 	var internalCertControllerCancel context.CancelFunc
 	var ctx context.Context
 
-	if os.Getenv("AUTO_INTERNAL_CERT") != "" {
-
-		log.Info("start initializing k8s internal secret controller and wait for internal secret creation if it's not created")
-
+	// Healthz server handles internal cert migration (enabled by default) and readiness checks (always enabled in k8s).
+	if platform == share.PlatformKubernetes {
 		go func() {
 			if err := healthz.StartHealthzServer(); err != nil {
 				log.WithError(err).Warn("failed to start healthz server")
 			}
 		}()
+	}
 
+	if os.Getenv("AUTO_INTERNAL_CERT") != "" {
+		log.Info("start initializing k8s internal secret controller and wait for internal secret creation if it's not created")
 		ctx, internalCertControllerCancel = context.WithCancel(context.Background())
 		defer internalCertControllerCancel()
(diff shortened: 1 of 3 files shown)


williamlin-suse (approved):
LGTM

PR #2418, opened 2026-06-04
enhance: reject load or upload the empty or corrupt cvedb

pohanhuang:
## Description
- Skip empty CVEDB updates from scanner registration
- Reject KV reloads that fail or reconstruct to an empty DB so the controller does not replace a valid in-memory DB with empty/corrupt data.

## Issue
https://github.com/neuvector/neuvector/issues/2423

### DIFF
--- a/controller/cache/scan.go
+++ b/controller/cache/scan.go
@@ -1235,12 +1235,22 @@ func ScannerUpdateHandler(nType cluster.ClusterNotifyType, key string, value []b
 					CVEDB:           make(map[string]*share.ScanVulnerability),
 				}
 
-				// Reassemble
-				dbs := clusHelper.GetScannerDB(newStore)
+				// Reassemble. Keep the current in-memory DB if the new store cannot be read
+				// or contains no usable entries.
+				dbs, err := clusHelper.GetScannerDB(newStore)
+				if err != nil {
+					log.WithFields(log.Fields{"error": err, "store": newStore, "version": s.CVEDBVersion}).Error("Failed to load scanner DB")
+					return
+				}
 				for _, db := range dbs {
 					for _, cve := range db.CVEDB {
 						newDB.CVEDB[cve.Name] = cve
 					}
+				}
+
+				if len(newDB.CVEDB) == 0 {
+					log.WithFields(log.Fields{"store": newStore, "version": s.CVEDBVersion}).Error("Dropping empty scanner DB update")
+					return
 				}
 
 				log.WithFields(log.Fields{"cvedb": newDB.CVEDBVersion, "entries": len(newDB.CVEDB)}).Info()
--- a/controller/grpc.go
+++ b/controller/grpc.go
@@ -178,7 +178,11 @@ func (ss *ScanService) HealthCheck(ctx context.Context, data *share.ScannerRegis
 	}
 
 	clusHelper := kv.GetClusterHelper()
-	s := clusHelper.GetScanner(scannerID, access.NewReaderAccessControl())
+	s, err := clusHelper.GetScanner(scannerID, access.NewReaderAccessControl())
+	if err != nil && !errors.Is(err, cluster.ErrKeyNotFound) {
+		log.WithFields(log.Fields{"error": err, "scanner": scannerID}).Warn("Failed to get scanner during health check")
+		return nil, fmt.Errorf("failed to get scanner %s: %w", scannerID, err)
+	}
 	if s != nil {
 		visible = true
 	}
@@ -220,8 +224,18 @@ func (ss *ScanService) scannerRegister(data *share.ScannerRegisterData) error {
 	}
 	defer clusHelper.ReleaseLock(lock)
 
+	hasCveDB := len(data.CVEDB) > 0
+	if !hasCveDB {
+		log.WithFields(log.Fields{"scanner": data.ID, "version": data.CVEDBVersion}).Warn("Skip empty scanner DB update")
+		return errors.New("scanner cvedb is empty")
+	}
+
 	// Check if the database is newer.
-	s := clusHelper.GetScanner(share.CLUSScannerDBVersionID, access.NewReaderAccessControl())
+	s, err := clusHelper.GetScanner(share.CLUSScannerDBVersionID, access.NewReaderAccessControl())
+	if err != nil && !errors.Is(err, cluster.ErrKeyNotFound) {
+		log.WithFields(log.Fields{"error": err, "version": data.CVEDBVersion}).Warn("Failed to get scanner DB version record")
+		return fmt.Errorf("failed to get scanner DB version record: %w", err)
+	}
 	if s == nil {
 		writeDB = true
 	} else {
--- a/controller/kv/helper.go
+++ b/controller/kv/helper.go
@@ -134,14 +134,14 @@ type ClusterHelper interface {
 	DeleteProcessProfileTxn(txn *cluster.ClusterTransact, group string) error
 	GetAllProcessProfileSubKeys(scope string) utils.Set
 
-	GetScanner(id string, acc *access.AccessControl) *share.CLUSScanner
+	GetScanner(id string, acc *access.AccessControl) (*share.CLUSScanner, error)
 	GetAllScanner(acc *access.AccessControl) []*share.CLUSScanner
 	PutScannerTxn(txn *cluster.ClusterTransact, s *share.CLUSScanner) error
 	DeleteScanner(id string) error
 	GetScannerStats(id string) (*share.CLUSScannerStats, error)
 	CreateScannerStats(id string) error
 	PutScannerStats(id string, objType share.ScanObjectType, result *share.ScanResult) error
-	GetScannerDB(store string) []*share.CLUSScannerDB
+	GetScannerDB(store string) ([]*share.CLUSScannerDB, error)
 
 	// InitCreditOwners initializes the scan credit owners for the cluster.
 	InitCreditOwners() error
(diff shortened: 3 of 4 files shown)


controller/grpc.go
```diff
@@ -222,14 +222,20 @@ func (ss *ScanService) scannerRegister(data *share.ScannerRegisterData) error {
 
 	// Check if the database is newer.
 	s := clusHelper.GetScanner(share.CLUSScannerDBVersionID, access.NewReaderAccessControl())
```
holyspectral: Maybe you should also make this function call return the error?  Then you don't have to check `s == nil` unless you want to be super cautious/defensive.

controller/cache/scan.go
```diff
@@ -1235,14 +1235,24 @@ func ScannerUpdateHandler(nType cluster.ClusterNotifyType, key string, value []b
...
 
-				// Reassemble
-				dbs := clusHelper.GetScannerDB(newStore)
+				// Reassemble. Keep the current in-memory DB if the new store cannot be read
+				// or contains no usable entries.
+				dbs, err := clusHelper.GetScannerDB(newStore)
+				if err != nil {
+					log.WithFields(log.Fields{"error": err, "store": newStore, "version": s.CVEDBVersion}).Error("Failed to load scanner DB")
+					return
+				}
```
holyspectral, line 1244: The error here is actually not handled as well.  We probably should have a retry mechanism [here](https://github.com/neuvector/neuvector/blob/d4ceaa5fcdcca7a25a04baf55a2e0c6e4748c8ab/share/cluster/consul.go#L1335), similar to the workqueue that controller-runtime has.  Otherwise if it fails to handle an event from consul, the event will just disappear and never get handled again.

holyspectral (commented):
Thanks for looking into this.  Overall it looks good to me, but I think we should expand the scope a little so it would catch similar issues in the existing code base.

pohanhuang:
@holyspectral I just add error handling for clusHelper.GetScanner.
Let's handle retry mechanism in another PR.

Please help review when you are available, thanks

controller/grpc.go
```diff
@@ -221,15 +224,24 @@ func (ss *ScanService) scannerRegister(data *share.ScannerRegisterData) error {
 	defer clusHelper.ReleaseLock(lock)
 
 	// Check if the database is newer.
-	s := clusHelper.GetScanner(share.CLUSScannerDBVersionID, access.NewReaderAccessControl())
+	s, err := clusHelper.GetScanner(share.CLUSScannerDBVersionID, access.NewReaderAccessControl())
+	if err != nil {
+		return fmt.Errorf("failed to get scanner DB version record: %w", err)
+	}
```
holyspectral, line 238: You probably should also handle `ErrKeyNotFound`?  Otherwise in the fresh installation it will never succeed.

controller/grpc.go
```diff
@@ -221,15 +224,24 @@ func (ss *ScanService) scannerRegister(data *share.ScannerRegisterData) error {
...
+		}
+	} else if hasCveDB {
 		ver, _ := utils.NewVersion(s.CVEDBVersion)
 		if newVer.Compare(ver) > 0 || len(data.CVEDB) > s.CVEDBEntries {
 			writeDB = true
 		}
 	}
+	if !hasCveDB {
+		log.WithFields(log.Fields{"scanner": data.ID, "version": data.CVEDBVersion}).Warn("Skip empty scanner DB update")
+	}
```
holyspectral: I'd expect it will be like this: (pseudo code)

```
	if len(data.CVEDB) == 0 {
	    // This will prevent the scanner from registering.  It will keep retrying but this is the best we can do as the scanner is invalid.
		log.WithFields(log.Fields{"scanner": data.ID, "version": data.CVEDBVersion}).Warn("Skip empty scanner DB update")
		return errors.New("cvedb is empty")
	}
	s, err := clusHelper.GetScanner(share.CLUSScannerDBVersionID, access.NewReaderAccessControl())
	if err != nil {
        if !errors.Is(err, cluster.ErrKeyNotFound) {
            // This will force scanner to retry
            return fmt.Errorf("failed to get scanner DB version record: %w", err)
        }
    }
    // the below is basically the same logic. 
    if s == nil {
            // cvedb is empty.  We should update cvedb.
            writeDB = true
    }
    ver, _ := utils.NewVersion(s.CVEDBVersion)
	if newVer.Compare(ver) > 0 || len(data.CVEDB) > s.CVEDBEntries {
	    writeDB = true
	}
```

controller/kv/helper.go
```diff
@@ -1578,20 +1578,25 @@ func (m clusterHelper) PutScannerStats(id string, objType share.ScanObjectType,
...
 		if !acc.Authorize(&s, nil) {
-			return nil
+			return nil, common.ErrObjectAccessDenied
 		}
 
-		return &s
+		return &s, nil
 	}
-	return nil
+	return nil, common.ErrObjectNotFound
```
holyspectral: TBH I don't know if it's a good idea to use these errors defined in common package.  If it is, we should also make sure the error types returned here are consistent.  E.g., in line 1585, when the data is not available in consul, the error will be `cluster.ErrKeyNotFound`, but in line 1599 we have another similar error `common.ErrObjectNotFound` from `common` package.

holyspectral (commented):
Thanks!  A few follow-up comments for the latest changes.

controller/kv/helper.go
```diff
@@ -1617,28 +1622,33 @@ func (m clusterHelper) DeleteScanner(id string) error {
 	return cluster.Delete(key)
 }
 
-func (m clusterHelper) GetScannerDB(store string) []*share.CLUSScannerDB {
+func (m clusterHelper) GetScannerDB(store string) ([]*share.CLUSScannerDB, error) {
```
holyspectral, line 1626: This is an open question and the code here I think looks good to me as well - How could we ensure that when we extract the cvedb here, the items are as expected?  For example, say cvedb contains 100 gzip files stored in consul.  The logic here extracts them all, but it only contains 99 files. It seems like there is no mechanism to detect that.  I believe this could be useful when the retry mechanism mentioned in [here](https://github.com/neuvector/neuvector/pull/2418/changes#r3357080180) is absent.

controller/kv/helper.go
```diff
@@ -1578,28 +1578,33 @@ func (m clusterHelper) PutScannerStats(id string, objType share.ScanObjectType,
...
+	return nil, cluster.ErrKeyNotFound
 }
 
 // GetScannerRev gets scanner with revision for internal operations (no auth check)
 func (m clusterHelper) GetScannerRev(id string) (*share.CLUSScanner, uint64, error) {
 	key := share.CLUSScannerKey(id)
 	value, rev, err := m.get(key)
 	if err != nil || value == nil {
-		return nil, 0, common.ErrObjectNotFound
+		return nil, 0, cluster.ErrKeyNotFound
```
holyspectral, line 1608: I'd add a comment why this value is chosen.

controller/grpc.go
```diff
@@ -178,7 +178,10 @@ func (ss *ScanService) HealthCheck(ctx context.Context, data *share.ScannerRegis
 	}
 
 	clusHelper := kv.GetClusterHelper()
-	s := clusHelper.GetScanner(scannerID, access.NewReaderAccessControl())
+	s, err := clusHelper.GetScanner(scannerID, access.NewReaderAccessControl())
+	if err != nil && !errors.Is(err, cluster.ErrKeyNotFound) {
+		return nil, fmt.Errorf("failed to get scanner %s: %w", scannerID, err)
+	}
```
holyspectral, line 185: The error here is returned to scanner.  I'd also log a warning here so we have something to reference on controller side.

controller/grpc.go
```diff
@@ -220,8 +223,17 @@ func (ss *ScanService) scannerRegister(data *share.ScannerRegisterData) error {
...
+	if !hasCveDB {
+		log.WithFields(log.Fields{"scanner": data.ID, "version": data.CVEDBVersion}).Warn("Skip empty scanner DB update")
+		return errors.New("scanner cvedb is empty")
+	}
+
 	// Check if the database is newer.
-	s := clusHelper.GetScanner(share.CLUSScannerDBVersionID, access.NewReaderAccessControl())
+	s, err := clusHelper.GetScanner(share.CLUSScannerDBVersionID, access.NewReaderAccessControl())
+	if err != nil && !errors.Is(err, cluster.ErrKeyNotFound) {
+		return fmt.Errorf("failed to get scanner DB version record: %w", err)
```
holyspectral, line 237: Similar to [this comment](https://github.com/neuvector/neuvector/pull/2418/changes#r3363722758), let's keep a warning here.

holyspectral (commented):
Thanks!  It looks much better now.  Just some minor comments about a separate logs inside controller.  Let's also talk more about the cvedb verification.

controller/kv/helper.go
holyspectral: If we have something to detect that, we can ask users/agents to monitor this specific log, so it will significantly reduce our support cost.  WDYT?

pohanhuang: Hi @holyspectral  thanks for bring this topic out! I suggest we create a separate issue to track this.
Here's my proposed approach:

1. Publish CVEDB metadata to GHCR as the source of truth. (contains the meta like CVEID only)
2. When updating the CVEDB, pull this metadata and validate against it before applying the update.
What do you think? We can finish the current refactoring first, then tackle this as a follow-up.

WDYT?

pohanhuang: https://github.com/neuvector/neuvector/issues/2424 issue create to track the open question

holyspectral (commented):
LGTM.  Let's squash it together then we can merge.

pohanhuang:
@holyspectral I just squash it, please help review and merge when you are available.
Thanks

holyspectral (approved):

PR #2455, opened 2026-06-10
refactor: add reexam intf if container's intf is empty at first time

kyledong-suse:
## Description

This is an improvement that in recent OpenShift cluster, customer's build pod cannot detect interfaces. 
There is a race condition, when the Build pod starts, NV intercept runs early which doesn't find any network interfaces(CNI not ready yet, or not fully ready). Then CNI/OVN finishes: veth + IP assigned to the pod early before NV starts to monitor interface changes. So during the build runs, there is no further netlink changes, which `intfMonitorLoop` is never triggered, and interfaces stay empty until pod exits.
This PR schedule a reexam after 4s if the first time interface is empty.

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/agent/engine.go
+++ b/agent/engine.go
@@ -1852,6 +1852,9 @@ func examNetworkInterface(c *containerData) bool {
 		if intfAdded {
 			programBridge(c)
 			programDP(c, false, nil)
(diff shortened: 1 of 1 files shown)


holyspectral (approved):
I think the finding makes sense.  Thanks!

williamlin-suse (approved):
LGTM

kyledong-suse:
Thanks for reviewing it. I'll convert it to draft to let QA to run a fully test before merge it.

PR #2459, opened 2026-06-10
code-scanning-23-24-27.2: Size computation for allocation may overflow

williamlin-suse:
## Description

Fix https://github.com/neuvector/neuvector/security/code-scanning/23 , https://github.com/neuvector/neuvector/security/code-scanning/24 , https://github.com/neuvector/neuvector/security/code-scanning/27

### DIFF
--- a/controller/rest/federation.go
+++ b/controller/rest/federation.go
@@ -922,7 +922,11 @@ func getJointClusterToken(rc *share.CLUSFedJointClusterInfo, clusterID string, u
 		var data []byte
 		var statusCode int
 		var proxyUsed bool
-		body, _ := json.Marshal(reqTo)
+		body, err := json.Marshal(reqTo)
+		if err != nil {
+			log.WithFields(log.Fields{"error": err}).Error("failed to marshal request payload")
+			return "", err
+		}
 		// call joint cluster for generating a regular auth token
 		if _, statusCode, data, proxyUsed, err = sendReqToJointCluster(rc.RestInfo, clusterID, "", http.MethodPost, "v1/fed_auth",
 			"", jsonContentType, _tagAuthJointCluster, body, false, _notForward, false, true, acc); err == nil {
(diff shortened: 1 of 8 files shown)


share/utils/utils.go
```diff
@@ -922,11 +922,16 @@ func GetSupportedTLSCipherSuites() []uint16 {
 }
 
 func Encrypt(encryptionKey, text []byte) ([]byte, error) {
+	textLength := len(text)
+	if textLength >= (4294967296 - aes.BlockSize) {
+		return nil, fmt.Errorf("size too big: %v", textLength)
+	}
+	textLength += aes.BlockSize // max value: 4G
```
holyspectral: Not a defeat but this is really a big number.  lol

controller/rest/federation.go
```diff
@@ -3127,7 +3183,10 @@ func pollFedScanData(cachedRegConfigRev *uint64, cachedScanResultHash map[string
 	}
 
 	reqTo.ID = jointCluster.ID
-	reqTo.JointTicket = jwtGenFedTicket(jointCluster.Secret, jwtFedJointTicketLife)
+	reqTo.JointTicket, err = jwtGenFedTicket(jointCluster.Secret, jwtFedJointTicketLife)
+	if err != nil {
+		return 0, updated, deleted, delRegs, true
+	}
```
holyspectral, line 3192: It seems like the error is not printed or returned to caller.  It's better to keep a log here so the error won't be ignored silently.

controller/rest/policy.go
```diff
@@ -2257,6 +2257,10 @@ LOOP_ALL_IDS:
 						break LOOP_ALL_IDS
 					} else {
 						reservedLen := len(mon.Filters) + len(mon.FiltersCRD)
+						// KVValueSizeMax is 512 * 1024. So it's reasonable to assume there is no more than 512 * 1024 mon.Filters+mon.FiltersCRD entries
+						if reservedLen < 0 || reservedLen > cluster.KVValueSizeMax {
+							reservedLen = cluster.KVValueSizeMax
+						}
```
holyspectral: Maybe we could write it like this to avoid the overflow handling, as the logic to handle overflow is always not straightforward.  

```suggestion
						reservedLen := min(len(mon.Filters), cluster.KVValueSizeMax) + min(len(mon.FiltersCRD), cluster.KVValueSizeMax)
						// KVValueSizeMax is 512 * 1024. So it's reasonable to assume there is no more than 512 * 1024 mon.Filters+mon.FiltersCRD entries
						if filterSize +  filterCRDSize > cluster.KVValueSizeMax {
							reservedLen = cluster.KVValueSizeMax
						}
```

holyspectral (commented):
LGTM.  Just two minor findings.

holyspectral (approved):
LGTM.  Thanks!

PR #2460, opened 2026-06-11
test(e2e): authentication tests

holyspectral:
## Description

A simple test case to login controller. 

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/auth_test.go
+++ b/test/e2e/auth_test.go
@@ -1,0 +1,94 @@
+package e2e_test
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+	"sigs.k8s.io/e2e-framework/pkg/types"
+)
+
+type nvAuthRequest struct {
+	Password nvPasswordAuth `json:"password"`
+}
+
+type nvPasswordAuth struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
+}
+
+type nvAuthResponse struct {
+	Token struct {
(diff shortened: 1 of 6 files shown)


test/e2e/auth_test.go
```diff
@@ -0,0 +1,95 @@
...
+		loginTimeout       = 2 * time.Minute
+	)
+	deadline := time.Now().Add(loginTimeout)
+	var lastErr error
+	for time.Now().Before(deadline) {
+		lastErr = tryLogin(client, url, body)
+		if lastErr == nil {
+			return ctx
+		}
+		t.Logf("login attempt failed (%v), retrying in %s", lastErr, loginRetryInterval)
```
pohanhuang, line 67: Is this expected to treat every error the same?

test/e2e/client_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+		nodes := &corev1.NodeList{}
+		if err = client.Resources().List(ctx, nodes); err != nil {
+			return ctx, fmt.Errorf("list nodes: %w", err)
+		}
+		if len(nodes.Items) == 0 {
+			return ctx, fmt.Errorf("no nodes found in cluster")
+		}
+		nodeIP := ""
+		for _, addr := range nodes.Items[0].Status.Addresses {
```
pohanhuang, line 57: Does it assume we always single node situation?
Otherwise in multi-node, it should use the leader node?

holyspectral: Actually how the NodePort Service works is that, the port will redirect the traffic to controller pods, so using the first node is fine here and should continue working if there are multiple nodes.  Of course, we can try to have a ingress controller to handle this part better, but I'd rather do that later when we have more coverage in PR checks.

test/e2e/auth_test.go
holyspectral: Yes for now.  We can do more iterations for this part when it starts failing.

pohanhuang (approved):

PR #2475, opened 2026-06-12
fix(lint): resolve errcheck in dp, pipe, access, db, container, global, scan, and utils packages

holyspectral:
## Description

Fix 50 unchecked error returns across 20 files: propagate errors where possible, log at appropriate levels (Warn default, Debug for high-frequency or advisory paths), and use t.Fatalf/require.NoError in test code. Notable: replace net.ParseMAC with byte-literal HardwareAddr in tc.go to eliminate the error path entirely for format-computed MACs.

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests


kyledong-suse (approved):
LGTM, thanks.

PR #2477, opened 2026-06-12
issue-2469: Remove the alert, Neuvector is running {version1} and can be upgraded to {version2}, from UI

### LINKED ISSUE #2469: [Task] Remove "Neuvector is running {version1} and can be upgraded to {version2}" from UI
**Summary**

1. Different release cycles
rancher, rancher prime and future appco have different release cycles.  In appco scenario, the release will even be fully controlled by appco team.  This can cause issues easily when users get a prompt saying that an upgrade is available, but it's not available in the specific channel.  I think we usually get questions about rancher release right after community release exactly for this reason. 
Supporting multiple release channel might be doable, but it is lots of maintenance effort. 

2. Rancher/rancher prime already have similar features.
Keeping two different mechanism leads to conflict information.

3. Operation gap
Remembering to update the file is an operation gap.  I'd be more than happy to get rid of it whenever it's possible. 

4. Org silo
The file is maintained by EIO team and it's like a black box to us.  It's hard to have a mechanism to monitor it and maintain it.
Like you pointed out, it failed for half a year and no one actually noticed it.  I also noticed this today that it's generating 5.4.7 instead of 5.5.2 we have in file. 

**Describe the task**
NeuVector UI will stop popping up the message "Neuvector is running {version1} and can be upgraded to {version2}" to prompt users to upgrade.

williamlin-suse:
## Description

Fix #2469

### DIFF
--- a/controller/api/internal_apis.go
+++ b/controller/api/internal_apis.go
@@ -139,26 +139,6 @@ type RESTProfilingData struct {
 	Profiling *RESTProfiling `json:"profiling"`
 }
 
-type RESTK8sNvRbacStatus struct {
-	ClusterRoleErrors        []string                   `json:"clusterrole_errors,omitempty"`        // obsolete
-	ClusterRoleBindingErrors []string                   `json:"clusterrolebinding_errors,omitempty"` // obsolete
-	RoleErrors               []string                   `json:"role_errors,omitempty"`               // obsolete
-	RoleBindingErrors        []string                   `json:"rolebinding_errors,omitempty"`        // obsolete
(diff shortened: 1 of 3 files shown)


kyledong-suse (commented):
Generally LGTM. Just a question about those fields, i.e., MinUpgradeVersion, MaxUpgradeVersion. They are not used anymore, do we still want to keep them?

williamlin-suse:
Front-end prefers to keep the logic about checking the value of `telemetryStatus.max_upgrade_version.tag` in the response.
If `max_upgrade_version` is removed from `telemetryStatus`, it will cause exception in UI.
So I still keep them in our code ....

kyledong-suse (approved):

PR #2486, opened 2026-06-15
code-scanning-22-25: Size computation for allocation may overflow

williamlin-suse:
## Description

Fix https://github.com/neuvector/neuvector/security/code-scanning/22 , https://github.com/neuvector/neuvector/security/code-scanning/25

### DIFF
--- a/controller/cache/csp_billing_adapter.go
+++ b/controller/cache/csp_billing_adapter.go
@@ -48,7 +48,11 @@ func (m CacheMethod) GetNvUsage(fedRole string) api.RESTNvUsage {
 		unreachable := 0
 		cspUsages := make(map[string]int)
 		fedCacheMutexRLock()
-		memberUsages := make([]*api.RESTClusterCspUsage, 0, len(fedJoinedClusterStatusCache)+1)
+		capacity := len(fedJoinedClusterStatusCache)
+		if capacity < common.MaxFedManagedClusters {
+			capacity++
+		}
+		memberUsages := make([]*api.RESTClusterCspUsage, 0, capacity)
 		for _, cached := range fedJoinedClusterStatusCache {
 			if connectedStates.Contains(cached.Status) {
 				_, memberCspType := common.GetMappedCspType(nil, &cached.CspType)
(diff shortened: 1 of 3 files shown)


controller/cache/csp_billing_adapter.go
```diff
@@ -48,7 +49,11 @@ func (m CacheMethod) GetNvUsage(fedRole string) api.RESTNvUsage {
 		unreachable := 0
 		cspUsages := make(map[string]int)
 		fedCacheMutexRLock()
-		memberUsages := make([]*api.RESTClusterCspUsage, 0, len(fedJoinedClusterStatusCache)+1)
+		capacity := len(fedJoinedClusterStatusCache)
+		if capacity < math.MaxInt {
+			capacity++
+		}
+		memberUsages := make([]*api.RESTClusterCspUsage, 0, capacity)
```
holyspectral, line 55: I'd recommend defining a big but also reasonable limit and sharing it everywhere.  The make statement here obviously can't allocate 9,223,372,036,854,775,807 - 1 items.

holyspectral (commented):
Thanks for working on this!  Just a comment about defining the max capacity.

controller/rest/server.go
```diff
@@ -1526,7 +1527,11 @@ func handlerServerRoleGroupsConfig(w http.ResponseWriter, r *http.Request, ps ht
 }
 
 func configOneGroupRolesMapping(groupRoleMapping *share.GroupRoleMapping, clusGroupMappedRoles []*share.GroupRoleMapping, acc *access.AccessControl) ([]*share.GroupRoleMapping, error) {
-	newSettings := make([]*share.GroupRoleMapping, 0, len(clusGroupMappedRoles)+1)
+	capacity := len(clusGroupMappedRoles)
+	if capacity < math.MaxInt {
+		capacity++
+	}
+	newSettings := make([]*share.GroupRoleMapping, 0, capacity)
```
holyspectral, line 1533:
```suggestion
	capacity := len(clusGroupMappedRoles)
	if capacity < common.MaxFedManagedClusters {
		capacity++
	}
	newSettings := make([]*share.GroupRoleMapping, 0, capacity)
```

I also like to keep it a one-liner, but it's not a must-have, e.g., (pseudo code)

```
    capacity := min(len(clusGroupMappedRoles) + 1, common.MaxFedManagedClusters)
```

holyspectral (commented):
LGTM, but I'm also open to changing it back to what you previously have, if my version would cause false positive.  I'd also recommend making sure two implementations are consistent, so maybe we can reuse the same patterns elsewhere.

holyspectral (approved):
LGTM.  Thanks!

PR #2489, opened 2026-06-15
feat: optimize scanner registration flow

### LINKED ISSUE #2485: scanner cvedb upgrade takes huge memory utilization
The current cvedb update takes a lot of memory.  After investigation, this can be categorized into two parts:

1. cvedb sent by scanner.  When scanner starts, especially when triggered by neuvector updater cronjob, it sends the whole cvedb to controller via gRPC API to see if the one stored in controller is up-to-date. This part contains:
    - gRPC buffer. 
    - internal buffer to hold the cvedb before they're written into consul as gzip format.
2. When the cvedb stored in consul gets updated, it will trigger controller pod to reload:
    - The memory allocated for new cvedb.
    - A temporary storage for cvedb before it gets parsed.

In part#1, depending on which controller pod gets the request, a controller pod can receive requests from multiple scanners.  This could lead to a high memory pressure and could cause OOM if the resource of the node or controller pod is constrained.

The part#2 also requires 2~3 copies of cvedb in the memory.  

This issue is to review the memory utilization and improve the overall utilization during scanner cvedb update.

holyspectral:
## Description

Today the scanner registration flow could cause huge memory pressure to controller pods.  To reduce the impact, this PR contains two parts:

1. Implement a new gRPC API, ScannerRegisterV3, to allow scanners to not send cvedb to controller during registration.
2. Instead of keeping the whole uncompressed CVEDB in memory, most of the data is moved to sqlite file after reading from consul.  The data will only be read into memory cache when it's being requested.  

Fixes #2485 

### Checklist

- [ ] includes documentation
- [x] adds unit tests
- [x] adds or updates e2e tests


controller/grpc.go
pohanhuang, line 229: Can we call  the `cvedb := preprocessDB(data)` here?
Ensuring the memory only exist in this scope to shorten the time where memory can be hold.

pohanhuang, line 650: Mentioned above.

db/models.go
```diff
@@ -191,16 +192,18 @@ const (
 	Table_assetvuls  = "assetvuls"
 	Table_querystats = "querystats"
 	Table_bench      = "bench"
+	Table_cvedb      = "cvedb"
+	Table_cvedbMeta  = "cvedb_meta"
```
pohanhuang: This is not needed, no meta table here.

pohanhuang (changes requested):
Most LGTM, thanks for the changes 👍 Left some comments.

In https://github.com/neuvector/neuvector/blob/main/share/utils/utils.go#L141-L150
I think we can further improve

```go
func GzipBytes(buf []byte) []byte {
	zb, err := GzipBytesWithLevel(buf, gzip.BestCompression)
	if err != nil {
		log.WithFields(log.Fields{"err": err, "len": len(buf)}).Error("failed to gzip bytes")
		return nil
	}
	return zb
}

func GzipBytesWithLevel(buf []byte, level int) ([]byte, error) {
	var b bytes.Buffer

	w, err := gzip.NewWriterLevel(&b, level)
	if err != nil {
		return nil, err
	}

	if _, err := w.Write(buf); err != nil {
		_ = w.Close()
		return nil, err
	}

	if err := w.Close(); err != nil {
		return nil, err
	}

	return b.Bytes(), nil
}
```

controller/grpc.go
pohanhuang, line 614: Why does this need to re-read from the cluster?Why does this need to re-read from the cluster? Can't it just pass the scanner in `scannerRegister` that needs to be registered?

holyspectral: I have no strong opinion, but if we want to do this, I think we should remove the `ScannerRegister()` API, so we don't have to care about the backward compatibility.  WDYT?

holyspectral:
@pohanhuang thanks.  Please take a look at my last commit. 

> GzipBytes

Of course we can do that, but I'm not quite sure if this is a great idea because usually with a higher compression rate it will require more CPU resource to extract.  Maybe we should keep it simple.  WDYT?

controller/grpc.go
holyspectral: In v2, the original API was reverted back to its original status.

controller/grpc.go
```diff
@@ -36,34 +38,186 @@ import (
...
+		return 0, err
+	}
+	// "name":<value>, — 2 quotes around name + 1 colon + value + 1 comma
+	return len(name) + 4 + len(b), nil
+}
+
+// exceedsSlotSize reports whether adding delta bytes to the current slot would exceed the threshold.
+// An empty slot is never considered over the limit to prevent infinite flush loops on oversized entries.
+func (w *incrementalDBWriter) exceedsSlotSize(delta int) bool {
+	return len(w.db.CVEDB) > 0 && w.currentSize+delta > cvedbSlotSizeMax
```
pohanhuang, line 132: The Consul slot size limit is 512 KB. cvedbSlotSizeMax is set to 400 KB (pre-compression), which is ~80% of the limit — leaving headroom for gzip overhead variance.

However, if a single oversized entry causes the compressed slot to exceed 512 KB, it will fail permanently with no way to recover without a code change.
cvedbSlotSizeMax is already a var, so one improvement would be to allow it to be overridden via an environment variable (similar to CVEDB_PAGE_SIZE), so operators can tune it without redeployment if they hit this edge case.

controller/grpc.go
```diff
@@ -150,27 +381,160 @@ func (ss *ScanService) ScannerRegisterStream(stream share.ControllerScanService_
...
+	log.WithFields(log.Fields{"scanner": req.ID, "version": req.CVEDBVersion, "entries": writer.Total}).Info("CVEDB received, registering scanner")
+
+	data := &share.ScannerRegisterData{
+		CVEDBVersion:    req.CVEDBVersion,
+		CVEDBCreateTime: req.CVEDBCreateTime,
+		RPCServer:       req.RPCServer,
+		RPCServerPort:   req.RPCServerPort,
+		ID:              req.ID,
+	}
+	if err := ss.finalizeDBCommit(data, newStore, writer.Total); err != nil {
```
pohanhuang, line 537: Could you confirm that `CLUSLockScannerKey` is never acquired while already holding `CLUSLockScannerDBUploadKey` in any other code path? 

The current call chain always acquires them in this order (ScannerDBUploadKey → ScannerKey), so as long as the reverse order never occurs elsewhere, there's no deadlock risk.

pohanhuang (changes requested):
Minor issue, other LGTM
We should add env_var for cvedbSlotSizeMax incase there is an edge case where the some CVE item is too big.

controller/grpc.go
```diff
@@ -36,34 +38,186 @@ import (
...
+	w.Total++
+	return nil
+}
+
+func (w *incrementalDBWriter) flush() error {
+	value, err := json.Marshal(w.db)
+	if err != nil {
+		return fmt.Errorf("failed to marshal db slot %d: %w", w.batch, err)
+	}
+	zb := utils.GzipBytes(value)
```
pohanhuang, line 160: nit: may be enhanced with BestCompression  https://pkg.go.dev/compress/gzip#pkg-constants

controller/grpc.go
holyspectral: I'd say this is an interesting edge case, but I'd respectfully disagree that this is a concern.  In order for this issue to be triggered, you'd need one one single cve data to exceed 400kb and this could already cause issues in the V2 API.  Even if this happens, given we control the cvedb, there are ways for us to mitigate from source.  

Of course having the variable configurable is a good practice, but in this case I'd rather not provide this, so we don't have to support scenario like `cvedbSlotSizeMax=500` or `cvedbSlotSizeMax=1000`.  Let's just use cvedb to control this.

controller/grpc.go
holyspectral: yeah I have some doubt about the config in [this comment](https://github.com/neuvector/neuvector/pull/2489#issuecomment-4782278695).  Maybe you can create a GitHub issue and we can follow up there?  It seems like a small fix that can be done separetely.

controller/grpc.go
holyspectral: The CLUSLockScannerDBUploadKey is acquired via `acquireCVEDBUploadLock` and this happens in the start of `ScannerRegisterStream()` and `ScannerRegisterV3()`.  On the other hand, `CLUSLockScannerKey` is acquired in `scannerRegister()` and `finalizeDBCommit()`, which is the last part of `ScannerRegisterV3()`

In other words, the order is always `CLUSLockScannerDBUploadKey` first and `CLUSLockScannerKey` later and it won't deadlock.

In addition, the lock has TTL so it will automatically resolve itself if it times out.

pohanhuang (approved):
LGTM

PR #2495, opened 2026-06-17
code-scanning-89: Uncontrolled data used in path expression

williamlin-suse:
## Description

Fix https://github.com/neuvector/neuvector/security/code-scanning/89

### DIFF
--- a/db/assetvuls.go
+++ b/db/assetvuls.go
@@ -27,6 +27,10 @@ import (
 type BuildWhereClauseFunc func(allowedID []string, queryFilter *api.VulQueryFilterViewModel) exp.ExpressionList
 type BuildWhereClauseAllFunc func(queryFilter *api.VulQueryFilterViewModel) exp.ExpressionList
 
+const (
+	queryTokenIdLen = 6 // do not change the length
+)
+
 func GetAssetVulIDByAssetID(assetID string) (*DbAssetVul, error) {
 	dialect := goqu.Dialect("sqlite3")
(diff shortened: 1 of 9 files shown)


db/assetvuls.go
```diff
@@ -992,6 +996,29 @@ func batchProcessAssetView(pool *pond.WorkerPool, mu *sync.Mutex, cvePackages ma
...
+	}
+	for i := 0; i < len(queryToken); i++ {
+		c := queryToken[i]
+		if (c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F') {
+			continue
+		}
+		return err
+	}
+	return nil
+}
```
holyspectral, line 1020: nit: In this case where you want to use a pre-defined error. it's better to give it a name.  

```suggestion
func vaildateQueryToken(queryToken string) error {
	invalidToken := errors.New("invalid query token")
	if len(queryToken) != queryTokenIdLen*2 {
		return invalidToken
	}
	for i := 0; i < len(queryToken); i++ {
		c := queryToken[i]
		if (c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F') {
			continue
		}
		return invalidToken
	}
	return nil
}
```

holyspectral (approved):
LGTM.  Just a nit.  

Also could you help to remove [this function](https://github.com/neuvector/neuvector/blob/fadaefd6588e4e97693e2c1d53c3c3bf8d3bbd91/share/utils/utils.go#L1396) when you're on it?  It's not used anymore and it will be flagged by other linters anyway.

PR #2519, opened 2026-06-24
fix: silent error handling in consul kv slot reads in network functions

### LINKED ISSUE #2512: Silent error handling in consul kv slot reads in network functions
Currently, there are multiple places on agent that fail silently. These may result in race condition or unexpected behaviour for network policy rules recalculation. Those silent error handling may miss some allow rules that should be applied to workload. If the original allow traffic doesn't have an allow rule to match it, it'll get denied by default which user will see unexpected implicit deny security events or potential network connection issue.

kyledong-suse:
## Description

Fix #2512 

1. add error handling for getting rules
2. add a retry for getting policy config

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/agent/policy/network.go
+++ b/agent/policy/network.go
@@ -6,6 +6,7 @@ import "C"
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"reflect"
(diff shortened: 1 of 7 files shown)


agent/policy/network.go
```diff
@@ -1503,24 +1502,34 @@ func (e *Engine) PushFqdnInfoToDP() error {
...
+				errs = append(errs, fmt.Errorf("key %s: failed to unzip", key))
+				continue
 			}
+			var fqdnip share.CLUSFqdnIp
+			if dbgError := json.Unmarshal(uzb, &fqdnip); dbgError != nil {
+				log.WithFields(log.Fields{"dbgError": dbgError, "key": key}).Error("Failed to decode FQDN data")
+				errs = append(errs, fmt.Errorf("key %s: %w", key, dbgError))
+				continue
+			}
+			dp.DPCtrlSetFqdnIp(&fqdnip)
```
holyspectral, line 1526: Is it okay to set partial fqdn ip to dp?  Or it has to go as a whole?

kyledong-suse: Partial FQDN push is fine as we don't want to block DP to be started. I can add a per-key Get retry here like what I did for `getPolicyConfig`, but the different is for FQDN -> IP map we don't fail with error.

holyspectral: yeah I think it would be great if we could make the behavior consistent.

holyspectral (approved):

PR #2550, opened 2026-06-29
fix: reduce log noises from dlp/dp policy

holyspectral:
## Description

Skip the warning if the previous dlp/dp policy doesn't exist. 

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests


kyledong-suse (commented):
LGTM, thanks! I just have one comment.
The other places like in [cluster.go](https://github.com/neuvector/neuvector/blob/main/agent/cluster.go#L824), and in helper.go [GetAllSigstoreRootsOfTrust](https://github.com/neuvector/neuvector/blob/main/controller/kv/helper.go#L4094) and [GetAllSigstoreVerifiersForRoot](https://github.com/neuvector/neuvector/blob/main/controller/kv/helper.go#L4232), they use `!errors.Is(err, cluster.ErrEmptyStore)`.
In [endpoint.go](https://github.com/neuvector/neuvector/blob/main/controller/kv/endpoint.go#L611) it uses `err == cluster.ErrEmptyStore`
I suggest we make them consistent.

kyledong-suse (approved):

PR #2555, opened 2026-06-29
code-scanning-95-96-97-98: Database query built from user-controlled sources

williamlin-suse:
## Description

Fix https://github.com/neuvector/neuvector/security/code-scanning/95, 
https://github.com/neuvector/neuvector/security/code-scanning/96, 
https://github.com/neuvector/neuvector/security/code-scanning/97, https://github.com/neuvector/neuvector/security/code-scanning/98

### DIFF
--- a/controller/api/apis.go
+++ b/controller/api/apis.go
@@ -4184,6 +4184,7 @@ type UserAccessControl struct {
 type QuerySessionRequest struct {
 	Type         int // 0=vul, 1=asset
 	QueryToken   string
+	LoginID      string
 	CreationTime int64
 	UserAccess   *UserAccessControl
 	Filters      *VulQueryFilterViewModel
--- a/controller/api/apis.go
+++ b/controller/api/apis.go
@@ -4184,7 +4184,6 @@ type UserAccessControl struct {
 type QuerySessionRequest struct {
 	Type         int // 0=vul, 1=asset
 	QueryID      string
-	LoginID      string
 	CreationTime int64
 	UserAccess   *UserAccessControl
 	Filters      *VulQueryFilterViewModel
--- a/controller/api/apis.go
+++ b/controller/api/apis.go
@@ -70,7 +70,7 @@ const RESTErrPromoteFail int = 49
 const RESTErrPlatformAuthDisabled int = 50
 const RESTErrRancherUnauthorized int = 51
 const RESTErrRemoteExportFail int = 52
-const RESTErrInvalidQueryToken int = 53
+const RESTErrInvalidQueryID int = 53
 const RESTErrPollJobNotFoundError int = 54
 const RESTErrServerError int = 55
 
@@ -4183,7 +4183,7 @@ type UserAccessControl struct {
 
 type QuerySessionRequest struct {
 	Type         int // 0=vul, 1=asset
-	QueryToken   string
+	QueryID      string
 	LoginID      string
 	CreationTime int64
 	UserAccess   *UserAccessControl
@@ -4289,7 +4289,7 @@ type RESTImageAssetViewV2 struct {
 type RESTVulQueryStats struct {
 	TotalRecordCount        int                     `json:"total_records"`
 	TotalMatchedRecordCount int                     `json:"total_matched_records"`
-	QueryToken              string                  `json:"query_token"`
+	QueryID                 string                  `json:"query_id"`
 	PerfStats               []string                `json:"debug_perf_stats"`
 	Summary                 *VulAssetSessionSummary `json:"summary"`
 	Status                  string                  `json:"status"`
@@ -4298,7 +4298,7 @@ type RESTVulQueryStats struct {
 // for asset pagination
 type RESTAssetQueryStats struct {
 	TotalRecordCount int                  `json:"total_records"`
-	QueryToken       string               `json:"query_token"`
+	QueryID          string               `json:"query_id"`
 	PerfStats        []string             `json:"debug_perf_stats"`
 	Summary          *AssetSessionSummary `json:"summary"`
 }
--- a/controller/rest/auth.go
+++ b/controller/rest/auth.go
@@ -684,7 +684,7 @@ func restReq2User(r *http.Request) (*loginSession, int, string) {
 				}
 
 				s := &loginSession{
-					id:          "apikey_" + apikeyAccount.Name,
+					id:          "apikey_" + hex.EncodeToString([]byte(apikeyAccount.Name)),
 					fullname:    apikeyAccount.Name,
 					remote:      r.RemoteAddr,
 					domainRoles: roles,
--- a/controller/rest/auth.go
+++ b/controller/rest/auth.go
@@ -684,7 +684,7 @@ func restReq2User(r *http.Request) (*loginSession, int, string) {
 				}
 
 				s := &loginSession{
-					id:          "apikey_" + hex.EncodeToString([]byte(apikeyAccount.Name)),
+					id:          "api_" + hex.EncodeToString([]byte(apikeyAccount.Name)),
 					fullname:    apikeyAccount.Name,
 					remote:      r.RemoteAddr,
 					domainRoles: roles,
--- a/controller/rest/auth.go
+++ b/controller/rest/auth.go
@@ -684,7 +684,7 @@ func restReq2User(r *http.Request) (*loginSession, int, string) {
 				}
 
 				s := &loginSession{
-					id:          "api_" + hex.EncodeToString([]byte(apikeyAccount.Name)),
+					id:          "apikey_" + hex.EncodeToString([]byte(apikeyAccount.Name)),
 					fullname:    apikeyAccount.Name,
 					remote:      r.RemoteAddr,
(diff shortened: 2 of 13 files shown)


williamlin-suse:
`vaildateQueryToken(queryToken)` called in `formatSessionTempTableName(queryToken)` makes sure the tableName(composed from user-input value) is valid

db/querystats.go
```diff
@@ -237,26 +236,19 @@ func deleteSessionTempTableInMemDb(queryToken string) error {
...
-				continue
-			}
-			break
+	tableName, err := formatSessionTempTableName(queryToken)
+	if err != nil {
+		return err
+	}
+	for i := 0; i < 10; i++ {
+		sql := fmt.Sprintf("DROP TABLE IF EXISTS '%s';", tableName)
+		_, err := db.Exec(sql)
```
holyspectral: hmm it looks like the issue is still being reported.  Maybe you want to take a second look?

williamlin-suse: vaildateQueryToken(queryToken) called in formatSessionTempTableName(queryToken) indeed makes sure the tableName is valid (i.e. SQL injection is no longer possible)
The same formatSessionTempTableName(queryToken) call fixes the alerts in other functions.
By design the temp tables are leveraged for speeding up the vulnerability query. 
Because of this, the SQL command is always built from user-controlled sources(tableName composed from queryToken).
I add the login id in the queryToken for composing the temp table name after the caller passes authentication. In this way a session(jwtToken) cannot be used to query/delete the temp tables  created for other sessions(jwtToken).

db/querystats.go
```diff
@@ -217,46 +221,40 @@ func DeleteQuerySessionByToken(queryToken string) error {
...
 
-func deleteSessionTempTableInMemDb(queryToken string) error {
+func deleteSessionTempTableInMemDb(queryToken, loginID string) error {
+	tableName, err := formatSessionTempTableName(queryToken, loginID)
+	if err != nil {
+		return err
+	}
+
 	memdbMutex.Lock()
 	defer memdbMutex.Unlock()
```
williamlin-suse, line 242: Line 245 is empty.
This alert is probably caused by prior commit but the code scanning tool/github cannot refresh the latest scan results

db/querystats.go
```diff
@@ -237,26 +236,19 @@
...
-				continue
-			}
-			break
+	tableName, err := formatSessionTempTableName(queryToken)
+	if err != nil {
+		return err
+	}
+	for i := 0; i < 10; i++ {
+		sql := fmt.Sprintf("DROP TABLE IF EXISTS '%s';", tableName)
+		_, err := db.Exec(sql)
```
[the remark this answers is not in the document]
williamlin-suse: Same as https://github.com/neuvector/neuvector/pull/2555/changes#r3502366348

db/models.go
```diff
@@ -354,8 +356,12 @@
 	return schema
 }
 
-func formatSessionTempTableName(queryToken string) string {
-	return fmt.Sprintf("tmp_session_%s", queryToken)
+func formatSessionTempTableName(queryID, loginID string) (string, error) {
+	if err := vaildateQueryID(queryID, loginID); err != nil {
+		return "", err
+	}
+	b := sha512.Sum512([]byte(queryID))
```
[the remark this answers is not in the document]
williamlin-suse: CodeQL is uncapable to tell that loginID is not sensitive data.
Using variable names to decide whether it's an alert is a bad rule in CodeQL

controller/rest/auth.go
```diff
@@ -684,7 +684,7 @@ func restReq2User(r *http.Request) (*loginSession, int, string) {
 				}
 
 				s := &loginSession{
-					id:          "apikey_" + apikeyAccount.Name,
+					id:          "api_" + hex.EncodeToString([]byte(apikeyAccount.Name)),
```
holyspectral: Could you double check if this is intentional as it changes the prefix from `apikey_` to `api_`?

db/assetvuls.go
```diff
@@ -1271,7 +1277,7 @@ func DupAssetSessionTableToFile(sessionToken string) error {
 
 func GetImageAssetSession(queryFilter *AssetQueryFilter) ([]*api.RESTImageAssetViewV2, int, error) {
 	if queryFilter != nil {
-		if err := vaildateQueryToken(queryFilter.QueryToken); err != nil {
+		if err := vaildateQueryID(queryFilter.QueryID); err != nil {
```
holyspectral: I can see `vaildateQueryID()` is called multiple times when `formatSessionTempTableName()` is used, but I think it's okay to fail early.

db/querystats.go
```diff
@@ -217,46 +216,40 @@ func DeleteQuerySessionByToken(queryToken string) error {
...
-		for i := 0; i < 10; i++ {
-			sql := fmt.Sprintf("DROP TABLE IF EXISTS '%s';", formatSessionTempTableName(queryToken))
-			_, err := db.Exec(sql)
-			if err != nil {
-				time.Sleep(100 * time.Millisecond)
-				continue
-			}
-			break
+	for i := 0; i < 10; i++ {
+		sql := fmt.Sprintf("DROP TABLE IF EXISTS '%s';", tableName)
```
holyspectral: I'd add a comment here to say why this is safe.

holyspectral (commented):
Thanks.  This is much clearer now.  Just a few minor comments.

controller/api/apis.go
```diff
@@ -4288,7 +4288,7 @@ type RESTImageAssetViewV2 struct {
 type RESTVulQueryStats struct {
 	TotalRecordCount        int                     `json:"total_records"`
 	TotalMatchedRecordCount int                     `json:"total_matched_records"`
-	QueryToken              string                  `json:"query_token"`
+	QueryID                 string                  `json:"query_id"`
```
holyspectral, line 4291: It looks like manager has these definition, e.g., https://github.com/neuvector/manager/blob/07446c5d0fd04a9a4bdf04816dfaa95bbd296628/admin/webapp/websrc/app/routes/vulnerabilities/vulnerabilities.service.ts#L67 and https://github.com/neuvector/manager/blob/07446c5d0fd04a9a4bdf04816dfaa95bbd296628/admin/webapp/websrc/app/common/services/registries.service.ts#L217.  I'd recommend checking with UI team to make sure these won't have adverse effect.

controller/rest/auth.go
williamlin-suse: oops. missed that. fixed in the 5th commit

controller/api/apis.go
williamlin-suse: I already inform Steven about this. After this PR is merged, UI needs to make corresponding change for handling the response of the 2 internal APIs:
```
POST(/v1/vulasset)
POST(/v1/assetvul)
```

db/querystats.go
```diff
@@ -217,46 +216,42 @@ func DeleteQuerySessionByToken(queryToken string) error {
...
-		for i := 0; i < 10; i++ {
-			sql := fmt.Sprintf("DROP TABLE IF EXISTS '%s';", formatSessionTempTableName(queryToken))
-			_, err := db.Exec(sql)
-			if err != nil {
-				time.Sleep(100 * time.Millisecond)
-				continue
-			}
-			break
+	// tableName is in the format "tmp_session_%s" where the "%s" part is a hex-encoded string(consists of the 16 lowercase hexadecimal characters: 0123456789abcdef)
+	// that SQL injection is impossible thru it
```
holyspectral, line 248: It's great to have this defined.  Can we have a cleanup code to ensure it?

db/querystats.go
```diff
@@ -217,46 +216,42 @@ func DeleteQuerySessionByToken(queryToken string) error {
...
+	sql := fmt.Sprintf("DROP TABLE IF EXISTS '%s';", tableName)
+	for i := 0; i < 10; i++ {
+		_, err := db.Exec(sql)
+		if err != nil {
+			time.Sleep(100 * time.Millisecond)
+			continue
 		}
+		break
 	}
```
holyspectral, line 257: Can we log an error here if it continued to fail for 10 times?

holyspectral (commented):
Thanks!  I think this looks great already.  Just two comments that could potentially make it more solid.

kyledong-suse (approved):
Based on last comments from @holyspectral, the latest commits LGTM.

db/querystats.go
williamlin-suse: In https://github.com/neuvector/neuvector/pull/2555/changes#diff-1c4da1c59df5f58a7b5191aaf635d5eef60c28acaff380f92a20a8c6c94c54b0R259 the error is returned to caller for logging

db/querystats.go
williamlin-suse: done by https://github.com/neuvector/neuvector/pull/2555/changes#diff-1c4da1c59df5f58a7b5191aaf635d5eef60c28acaff380f92a20a8c6c94c54b0R236

db/querystats.go
williamlin-suse: comment is added

PR #2557, opened 2026-06-30
fix: address govet issues

holyspectral:
## Description

Fix #XXX

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests


share/cluster/consul.go
```diff
@@ -1087,7 +1087,7 @@ func addNodeWatcher(watcher NodeWatcher) bool {
 	v := reflect.ValueOf(nodeWatchers)
 	f := reflect.ValueOf(watcher)
 	for i := 0; i < v.Len(); i++ {
-		if v.Index(i).Interface() == f {
+		if v.Index(i).Pointer() == f.Pointer() {
```
holyspectral, line 1090: This was actually a bug because `reflect.Value` would never equal to an empty interface.

williamlin-suse (approved):
LGTM

PR #2559, opened 2026-06-30
fix: reduce noisy logs

holyspectral:
## Description

This PR addresses two issues:

1. Previously, clusterHelper could return `cluster.ErrEmptyStore`, `cluster.ErrKeyNotFound` and `common.ErrObjectNotFound`.  This PR makes clusterHelper only returns `common.ErrObjectNotFound` by translating the error internally.
2. In the previous changes, more errors are handled, but not all logic ignores `common.ErrObjectNotFound` and other errors.  In this PR, those issues are addressed. 

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests


williamlin-suse (approved):
LGTM

Issue #1260, opened 2024-03-21
not running on arm64 raspberry pi5


didlawowo:
**Environment**
K3S / mixed node x86 / arm

**Describe the bug**
pod seems running but in fact got error on arm node (think the probe is not correct)

│ 2024-03-21T05:54:04.461|ERRO|AGT|probe.(*Probe).getNewConnections: Failed to read sockets - error=Error in netlink message                                      │

**To Reproduce**
helm investall vector latest chart

**Expected behavior**
running pod

didlawowo:
any news ?

DSP-doublehappy111:
Hi, Has this problem been solved

brokenlander:
Ran into this on an arm64 node and I think I can explain it.

It is not arm64 as such, it is the page size. `getNewConnections()` reads INET_DIAG dumps into a 5120 byte buffer, and the kernel sizes a dump datagram at `NLMSG_GOODSIZE`:

```c
#if PAGE_SIZE < 8192UL
#define NLMSG_GOODSIZE  SKB_WITH_OVERHEAD(PAGE_SIZE)
#else
#define NLMSG_GOODSIZE  SKB_WITH_OVERHEAD(8192UL)
#endif
```

`netlink_dump()` only shrinks that to the reader's buffer when the reader's buffer is larger. With 4K pages it comes out around 3776, under 5120, so everything fits and you never see this. With pages of 8K or more it is around 7872 and every datagram is truncated.

Ours is 64K, which puts it well over the line. That would also explain why this reads as an arm64 bug when most arm64 machines are fine: 4K-page arm64 never crosses the threshold, only 16K and 64K builds do. Worth running `getconf PAGESIZE` on the arm node to check, since I cannot tell from here what your kernel is built with.

Measured on our host with the same 5120 buffer, using `MSG_TRUNC` to see the real size:

```
PAGE_SIZE = 65536
buffer 5120:   real datagrams 7784-7868, 13 of 15 truncated, 578 sockets read
buffer 32768:  0 truncated, 889 sockets
```

Once a dump is abandoned part way through, the kernel keeps `cb_running` set on that socket and refuses every later dump with EBUSY, which comes back as `NLMSG_ERROR` and prints as `Error in netlink message`. Nothing reopens the socket, so it stays that way for the life of the pod. You can check with `getconf PAGESIZE` on the node.

[...]

PR #2569, opened 2026-07-03
nvbench: k3s CIS check OOM-kills the enforcer on nodes with a large journal

brokenlander:
### Related Issue
https://github.com/neuvector/neuvector/issues/2570

### What happens

On our k3s clusters the enforcer pod sits in CrashLoopBackOff, getting OOM-killed and restarting every few minutes. It's node-specific: it hits the k3s control-plane node and not the quieter worker.

### What's actually using the memory

The Go agent looked fine, so I checked the kernel OOM log, and the process getting killed isn't the agent, it's a shell:

```
Memory cgroup out of memory: Killed process ... (sh) total-vm:4975296kB, anon-rss:3994112kB
```

~3.8 GB of anon memory in a `sh` process. At the same moment the `agent` was ~36 MB, and a heap profile of the agent showed only ~35 MB in use, which is why this is easy to miss: the memory isn't in the Go process at all.

Tracing the shell, it's `agent/nvbench/journal.tmpl` (run via `nstools` for the k3s CIS check). The line that blows up is:

```sh
JOURNAL_LOG=$(journalctl -D /var/log/journal -u k3s)
```

That command substitution buffers the entire k3s journal into a shell variable. On a control-plane node the journal is large (ours is ~3.8 GB), so `JOURNAL_LOG` grows to match and the cgroup OOMs.

### The fix

The only use of `JOURNAL_LOG` is `fetch_kube_cmd`, which greps it for the `Running <component>` lines, so there's no need to hold the whole journal:

```sh
JOURNAL_LOG=$(journalctl -D /var/log/journal -u k3s | grep 'Running')
```

I ran the original and the patched script head-to-head on the same node against the real ~3.8 GB journal:

- peak RSS: **~3.9 GB -> ~227 MB**
- final output: **byte-identical** (all components still detected: kube-apiserver, controller-manager, scheduler, kube-proxy, kubelet)

So it changes nothing but the memory, and the OOM is gone.

This is separate from #1260 (that's the arm64 sock_diag netlink noise); this one is the memory blow-up in the k3s CIS journal check.

### DIFF
--- a/agent/nvbench/journal.tmpl
+++ b/agent/nvbench/journal.tmpl
@@ -11,7 +11,11 @@ fetch_kube_cmd() {
     echo "$JOURNAL_LOG" |  grep "Running $1" | tail -n1   | sed "s/[\"']//g"
 }
 
-JOURNAL_LOG=$(journalctl -D /var/log/journal -u k3s)
+# Only the "Running <component>" lines are used below (see fetch_kube_cmd), so
+# filter them out here instead of buffering the entire k3s journal into this
+# variable. On a control-plane node with a large journal that slurp can reach
+# several GB of shell memory and OOM-kill the enforcer.
+JOURNAL_LOG=$(journalctl -D /var/log/journal -u k3s | grep 'Running')
 
 # Read logs for each component
 kube_apiserver_cmd=$(fetch_kube_cmd "$CIS_APISERVER_CMD")


pohanhuang (approved):
Hi Thanks for the contribution.
Great catch.

LGTM.

PR #2615, opened 2026-07-20
issue 2614: Revert  for Processes to NOT be the default

### LINKED ISSUE #2614: Revert "zero-drift" for Processes to NOT be the default
**Is your feature request related to a problem? Please describe.**
Background: Back when the zero drift feature was added to the product, it also had been set to be enabled by default in all new installations. There was a strong belief among some that doing so would remediate impacts of customers neglecting to ever promote groups out of Discover mode.

Impact: Despite good intentions, this decision has had a negative impact on almost every single user. Users expect the behavior, even to this day, to be as "normal." In other words, they are expecting NeuVector to learn processes while in discover mode; they expect discover to actually discover. Given that this default mode was never very broadly called out, many users are caught by surprise and it leads to a negative impact of the product. Additionally, the use case is applicable for a minority of users.

**Describe the solution you'd like**
Disable 'Zero-Drift' as the default. (i.e. use "Basic" profile baseline by default for new deployments)

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

williamlin-suse:
## Description

Fix [#2614](https://github.com/neuvector/neuvector/issues/2614)
Use `basic` by default for "new service profile baseline" in fresh NV deployment.

Need to update the
[Documentation](https://open-docs.neuvector.com/policy/processrules#policy---groups---process-profile-rules)
about this change.

## Test

This PR only affects fresh deployment.

```
curl -k -H "Content-Type: application/json" -H "X-Auth-Token: $TOKEN" "https://$ControllerIP:$ControllerPORT/v1/system/config" | jq -r '.config.new_service_profile_baseline'
```

It should get `basic` after this PR is merged.

### DIFF
--- a/controller/common/common.go
+++ b/controller/common/common.go
@@ -106,7 +106,7 @@ const RegistryFedRepoScanName string = "fed._repo_scan"
 var DefaultSystemConfig = share.CLUSSystemConfig{
 	NewServicePolicyMode:      share.PolicyModeLearn,
 	NewServiceProfileMode:     share.PolicyModeLearn,
-	NewServiceProfileBaseline: share.ProfileZeroDrift,
+	NewServiceProfileBaseline: share.ProfileBasic,
 	UnusedGroupAging:          share.UnusedGroupAgingDefault,
 	CLUSSyslogConfig: share.CLUSSyslogConfig{
 		SyslogIP:         nil,


kyledong-suse (approved):

venkateshjayagopal:
Tested with jenkins-nv-build-fleet-5874 build image. With Fresh deployment, the new service mode is set as "Basic".
Upgrading from previous versions, the new service mode will still be "Zero drift".

PR #2657, opened 2026-07-28
reduce unnecessary info/warn logs

williamlin-suse:
## Description

reduce unnecessary info/warn logs

PR #2662, opened 2026-07-28
reduce unnecessary info/warn logs (#2657)

williamlin-suse:
## Description

reduce unnecessary info/warn logs


kyledong-suse (approved):
LGTM

Issue #2673, opened 2026-08-04
dp: AF_PACKET ring fails to allocate on kernels with pages larger than 4K, enforcer captures nothing


brokenlander:
### What happens

We run NeuVector across a mixed fleet and one of our nodes is arm64 on a 64KB-page kernel (`6.8.0-1055-nvidia-64k`). On that node the enforcer comes up fine, connects to the controller, and reports healthy. The console shows workloads, pods and hosts, and 2 enforcers connected with 0 disconnected. Everything is green.

It captures no traffic at all. Not a low volume, none: `nv_conversation_bytes` has hundreds of series on each of our other clusters and zero on this one. We only found it by reading container logs, because nothing about the state of the enforcer says anything is wrong.

The only sign is this pair, repeating on every interface event:

```
dp_ring_v1: fail to mmap (size=0x400000).
dp_alloc_context: fail to open dp socket, iface=eth0
```

### The mmap is not the problem

4 MiB is a perfectly ordinary mapping, so this reads like memory pressure or a `MAP_LOCKED` limit, and that is where we spent our time first. It isn't either.

`setsockopt(PACKET_RX_RING)` is what fails, and `dp/ring.c` drops the return value:

```c
setsockopt(fd, SOL_PACKET, PACKET_RX_RING, req, sizeof(*req));
if (!tap) {
    setsockopt(fd, SOL_PACKET, PACKET_TX_RING, req, sizeof(*req));
}
```

Execution carries on to the `mmap`, which then fails for a second, derived reason: `packet_mmap()` computes `expected_size` from the rings that were allocated, no ring was allocated, so `expected_size` is 0 and it returns EINVAL. The error we see is two steps downstream of the real one, and `errno` is never printed, which is why it points nowhere useful.

### Why setsockopt refuses it

`packet_set_ring()` requires `tp_block_size` to be a multiple of `PAGE_SIZE`:

```c
if (unlikely(!PAGE_ALIGNED(req->tp_block_size)))
        goto out;                     /* -EINVAL */
```

and the block sizes are compile-time constants built around 4096:

```c
dp/ring.c:28  #define FRAME_SIZE_V1 (1024 * 2)
dp/ring.c:29  #define BLOCK_SIZE_V1 (FRAME_SIZE_V1 * 4)              /* 8192  */
dp/ring.c:32  #define BLOCK_SIZE_JUMBO_V1 (FRAME_SIZE_JUMBO_V1 * 2)  /* 32768 */
```

[...]

PR #2674, opened 2026-08-04
dp: size the AF_PACKET ring blocks from the runtime page size

brokenlander:
### Related Issue

https://github.com/neuvector/neuvector/issues/2673

### What happens

On an arm64 node with a 64KB-page kernel the enforcer runs, reports healthy and connected, and captures no traffic whatsoever. The only symptom is this, on every interface event:

```
dp_ring_v1: fail to mmap (size=0x400000).
dp_alloc_context: fail to open dp socket, iface=eth0
```

`tp_block_size` has to be a multiple of `PAGE_SIZE`. `BLOCK_SIZE_V1` is 8192 and `BLOCK_SIZE_JUMBO_V1` is 32768, both fine against 4096 and both rejected with EINVAL against 65536. The `setsockopt` return value is dropped, so the first thing anyone sees is the `mmap` failing further down with no errno, which sends you looking at memory limits instead.

### The fix

Round the block size up to the runtime page size, and divide the block count by the same factor. The total ring size stays exactly what it was, which matters because these are mapped `MAP_LOCKED` and there is one per monitored interface. Rounding the block size alone would have taken the standard V1 ring from 8K to 64K per block and multiplied the locked footprint by eight.

Everything involved is a power of two, so the factor divides cleanly, `block_size` stays a multiple of `frame_size`, and `ring->size` stays a power of two, which `dp_rx_v1()` and `dp_tx_v1()` need for their `& (ring->size - 1)` wraparound.

On a 4K page kernel every value comes out unchanged, so x86_64 behaviour is untouched.

Also checks the two `setsockopt` calls and prints what the kernel actually said, along with the geometry and the page size. Without that this class of failure is close to undiagnosable from the logs.

### Test

Built with the standard `-Werror` flags:

```
cd dp && make
```

Ring geometry before and after, on an arm64 host with a 64KB-page kernel. These numbers come from a standalone program asking the kernel for the same geometries `dp_ring_v1()` builds, before and after the scaling this patch adds, rather than from the patched binary in place:

```
PAGE_SIZE = 65536

as shipped:
  tap     bs=8192   fs=2048   bn=512   total=0x400000   | setsockopt=Invalid argument
  inline  bs=8192   fs=2048   bn=2048  total=0x1000000  | setsockopt=Invalid argument
  jumbo   bs=32768  fs=16384  bn=2048  total=0x4000000  | setsockopt=Invalid argument

patched:
  tap     bs=65536  fs=2048   bn=64    total=0x400000   | setsockopt=ok | mmap=ok
  inline  bs=65536  fs=2048   bn=256   total=0x1000000  | setsockopt=ok | mmap=ok
  jumbo   bs=65536  fs=16384  bn=1024  total=0x4000000  | setsockopt=ok | mmap=ok
```

Same totals, so no extra locked memory.

### Notes for reviewer

There is no test suite under `dp/`, and no CI runner has a page size other than 4K, so this cannot be covered by a unit test or reproduced in CI. The before and after above is from real hardware and is the best evidence I can offer. On a 4K host the new code path is a no-op by construction, so the risk to existing platforms is the `setsockopt` return check rather than the geometry.

One thing to note on the helper: if the block count were ever smaller than the scaling factor it clamps to one block, and that is the single case where the ring would grow rather than stay the same size. It cannot happen with the current constants, since the smallest count is 512 and the largest factor is 16 on a 64K page, but it is worth knowing if those numbers ever change.

### DIFF
--- a/dp/ring.c
+++ b/dp/ring.c
@@ -34,6 +34,39 @@ static uint8_t g_tso_packet[MAX_TSO_SIZE];
 #define FRAME_SIZE_V3 (1024 * 2)
 #define BLOCK_SIZE_V3 (1024 * 64)
 
+/*
+ * The kernel requires tp_block_size to be a multiple of PAGE_SIZE. The block
+ * sizes above are multiples of 4096, so they are fine on a 4K-page kernel but
+ * rejected with EINVAL on a 16K or 64K page kernel (arm64 builds with
+ * CONFIG_ARM64_16K_PAGES / CONFIG_ARM64_64K_PAGES).
+ *
+ * Round the block size up to the page size and divide the block count by the
+ * same factor, so the total ring size, and with it the MAP_LOCKED footprint,
+ * is unchanged. Every block size, frame size and block count here is a power
+ * of two, so the factor divides both exactly: block_size stays a multiple of
+ * frame_size, and ring->size stays a power of two, which dp_rx_v1() and
+ * dp_tx_v1() depend on for their "& (ring->size - 1)" wraparound.
+ *
+ * On a 4K-page kernel this is a no-op.
+ */
+static void dp_ring_fit_page_size(unsigned int *block_size, uint *blocks)
+{
+    long page_size = sysconf(_SC_PAGESIZE);
+
+    if (page_size <= 0 || *block_size == 0 || (*block_size % (unsigned int)page_size) == 0) {
+        return;
+    }
+
+    unsigned int aligned = (((*block_size) + (unsigned int)page_size - 1) /
+                            (unsigned int)page_size) * (unsigned int)page_size;
+    unsigned int factor = aligned / *block_size;
+
+    *block_size = aligned;
+    *blocks = (*blocks > factor) ? (*blocks / factor) : 1;
+
+    DEBUG_INIT("page_size=%ld block_size=%u blocks=%u\n", page_size, *block_size, *blocks);
+}
+
 static int dp_ring_bind(int fd, const char *iface)
 {
     struct sockaddr_ll ll;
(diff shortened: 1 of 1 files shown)


kyledong-suse (approved):
Please discard this review comment. I'll review the PR later.

kyledong-suse (changes requested):
Sorry that I missed your two PRs, I have reviewed https://github.com/neuvector/neuvector/pull/2675, which is good. I'm going to review this one later.

dp/ring.c
```diff
@@ -227,14 +264,23 @@ static int dp_ring_v1(int fd, const char *iface, dp_ring_t *ring, bool tap, bool
...
+        DEBUG_ERROR(DBG_CTRL, "fail to set rx ring: %s, block_size=%u frame_size=%u "
+                    "block_nr=%u frame_nr=%u page_size=%ld\n", strerror(errno),
+                    req->tp_block_size, req->tp_frame_size, req->tp_block_nr,
+                    req->tp_frame_nr, sysconf(_SC_PAGESIZE));
+        return -1;
+    }
     if (!tap) {
-        setsockopt(fd, SOL_PACKET, PACKET_TX_RING, req, sizeof(*req));
+        if (setsockopt(fd, SOL_PACKET, PACKET_TX_RING, req, sizeof(*req)) < 0) {
+            DEBUG_ERROR(DBG_CTRL, "fail to set tx ring: %s\n", strerror(errno));
```
kyledong-suse: Would you please mirror the same RX failure log for TX? I think it would be very useful for if TX ring setup fails with those information.

dp/ring.c
```diff
@@ -227,14 +264,23 @@ static int dp_ring_v1(int fd, const char *iface, dp_ring_t *ring, bool tap, bool
...
+            return -1;
+        }
     }
 
     ring->rx_map = mmap(NULL, ring->map_size, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_LOCKED, fd, 0);
     if (ring->rx_map == MAP_FAILED) {
-        DEBUG_ERROR(DBG_CTRL, "fail to mmap (size=0x%x).\n", ring->map_size);
+        DEBUG_ERROR(DBG_CTRL, "fail to mmap (size=0x%x): %s.\n", ring->map_size, strerror(errno));
         close(fd);
```
kyledong-suse: Would you please help to delete this `close(fd)` here? The caller `dp_open_socket` will close the `fd` if there is any failure.

kyledong-suse (commented):
Generally LGTM, just couple of minor comments.
When addressing review comments, please create a new commit instead of force-pushing or amending the original commit. That makes it much easier for me to review the incremental changes. Thanks!

dp/ring.c
brokenlander: Done, the TX failure now logs the same fields as the RX one. Rebased on main while I was at it.

dp/ring.c
brokenlander: Good catch, removed. dp_open_socket closes the fd on any failure so this was a double close.

kyledong-suse (approved):
LGTM, thanks!

PR #2679, opened 2026-08-05
NVSHAS-10304: match Istio proxymesh traffic against parent workload rules

### LINKED ISSUE #2680: Istio proxymesh traffic fails to match parent workload allow rules
When Istio sidecar injection is enabled, NeuVector can fail to match legitimate allow rules for proxymesh traffic monitored on the `lo` interface. Security events show implicit deny (often with source `Host:<node-IP>` via X-Forwarded-For), even when an allow rule exists on the parent workload (e.g. `nodes → nv.<workload>.<ns>`). Disabling sidecar injection stops the events; re-enabling them brings them back.

kyledong-suse:
## Description

The issue in customer's environment is with Istio sidecar, NV taps proxymesh traffic on lo. That EP has no useful policy, and dst is often 127.0.0.1, so lookups miss parent workload allow rules (e.g. nodes → cba-ui). istio proxymesh then reports implicit deny (sometimes Host:<node-IP>) even when the real connection should match an allow. Removing the sidecar stops the events.
This fix for proxymesh, switch to the parent workload policy handle early in reeval and wession reeval. It also improves when XFF is enabled, if dest is loopback, rewrite dst to parent pips. If XFF client IP equals packet SIP on proxymesh, still rematch (don’t skip) so dst rewrite can apply. 

FIX: #2680 

## Test

QA has run network policy and Istio+XFF tests, they all passed.

### Checklist

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/dp/apis.h
+++ b/dp/apis.h
@@ -222,6 +222,7 @@ typedef struct dpi_config_ {
     bool promisc;
     bool thrt_ssl_tls_1dot0;
     bool thrt_ssl_tls_1dot1;
+    bool match_proxymesh_parent_policy;
 
     io_mac_t dummy_mac;
     io_ep_t dummy_ep;
--- a/dp/dpi/dpi_policy.c
+++ b/dp/dpi/dpi_policy.c
@@ -640,6 +640,14 @@ int dpi_policy_lookup(dpi_packet_t *p, dpi_policy_hdl_t *hdl, uint32_t app,
                 dip = xff_replace_dst_ip;
             }
             memset(desc, 0, sizeof(dpi_policy_desc_t));
+        } else if (g_io_config && g_io_config->match_proxymesh_parent_policy &&
+                   cmp_mac_prefix(p->ep_mac, PROXYMESH_MAC_PREFIX)) {
+            if (xff_replace_dst_ip > 0) {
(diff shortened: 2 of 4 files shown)


holyspectral (commented):
I couldn't find significant defects in the change itself, but I have a few comments:

- This seems pretty environment-specific change, but we don't have any unit tests that cover these scenarios.  For a change like this, it's hard to know if it would break a specific scenario accidentally.
- This will also introduce a behavior change.  Before the change, the Istio proxymesh traffic will never match any rule, either allow or deny, but now they will.  If the traffic meets a DENY rule, after the upgrade, the traffic can start to be denied.

My suggestion is to implement a switch to control whether the workload policies will apply to Istio traffic.  If the change introduces any side effect or users prefer the previous behavior, the user can rollback to the original behavior.

kyledong-suse:
>My suggestion is to implement a switch to control whether the workload policies will apply to Istio traffic. If the change introduces any side effect or users prefer the previous behavior, the user can rollback to the original behavior.

That's a great suggestion. I have added the env var flag which can enable enforcer to match istio proxymesh with it's parent policy. By default, it's disabled and we keep the existing behavior.

williamlin-suse (approved):
LGTM

williamlin-suse (approved):

kyledong-suse:

PR #2699, opened 2026-08-12
limit read body size when necessary

williamlin-suse:
## Description

Limit reading request body size when necessary

### DIFF
--- a/controller/rest/admission.go
+++ b/controller/rest/admission.go
@@ -451,7 +451,7 @@ func handlerPatchAdmissionState(w http.ResponseWriter, r *http.Request, ps httpr
 		return
 	}
 
-	body, err := io.ReadAll(io.LimitReader(r.Body, MAX_REQUEST_BODY_SIZE))
+	body, err := io.ReadAll(getDefaultReader(r.Body))
 	if err != nil {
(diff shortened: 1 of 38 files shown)


pohanhuang (approved):

holyspectral (changes requested):
It looks good to me, but I'd have a centralized place for this instead, e.g.,

```
body, err := io.ReadAll(GetDefaultReader(r.Body))
```

```
func GetDefaultReader(r io.Reader) io.Reader {
    return io.LimitReader(r, MAX_REQUEST_BODY_SIZE)
}
```

I'd also adopt this on all REST calls, so we don't have to revisit these later.

holyspectral:
Thanks!  There seems to be a few left.  Could you also update these usage as well? 

image

williamlin-suse:
> Thanks! There seems to be a few left. Could you also update these usage as well?
> image

Those are left intentionally because their request payload could be bigger.
Those REST API handlers call io.ReadAll(r.Body) after authentication is done.
Are we sure we want to apply the 2MB size limit to all REST APIs which could break some functions?

holyspectral:
@williamlin-suse yeah that makes sense, but maybe you could apply a different maximum instead of making it unlimited?

PR #2705, opened 2026-08-12
fix: pod-service traffic is denied

### LINKED ISSUE #2704: Group contention resolution Restrict Mode yields unexpected results
When group contention resolution setting is configured as restrict, after the behavior is learned and switched to protect mode, the traffic would be blocked unless a few policies going to nv.ip.<service name> are added.

holyspectral:
## Description

When group contention mode is set as `restrict`, the pod-service traffic is blocked unintentionally until a special group is added.  This commit fixes the issue.

Fix #2704

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/dp/dpi/dpi_policy.c
+++ b/dp/dpi/dpi_policy.c
@@ -1061,7 +1061,11 @@ static int dpi_policy_lookup_by_key(dpi_policy_hdl_t *hdl, uint32_t sip, uint32_
         //egress traffic is to host ip
         if (is_internal && !(hdl->apply_dir & DP_POLICY_APPLY_EGRESS) && !ipv4_ent &&
             iptype != DP_IPTYPE_HOSTIP && iptype != DP_IPTYPE_UWLIP && !is_nbe && !check_sgm) {
-            // east-west egress traffic is always allowed
+            // internal (pod-to-pod) east-west egress traffic is always allowed.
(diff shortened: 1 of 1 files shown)


kyledong-suse (commented):
Thanks for the clear comparison in Jira about option 1 & 2. The changeset LGTM.

I think we'd better to also document under restrict mode + Protect on k8s, this bypass skips client-side matching for nv.ip.* / ClusterIP rules. Especially for explicit **deny** to nv.ip.* is no longer enforced. Not sure if there is actually a use case for this condition, but it's worth to add a comment.

holyspectral:
> I think we'd better to also document under restrict mode + Protect on k8s, this bypass skips client-side matching for nv.ip.* / ClusterIP rules. Especially for explicit deny to nv.ip.* is no longer enforced. Not sure if there is actually a use case for this condition, but it's worth to add a comment.

yeah given that we've already bypassed the cross namespace service traffic in Ln1081, I'd say the service traffic is meant to be bypassed, but I'll add more comment for sure.

kyledong-suse (approved):

holyspectral:

holyspectral:

holyspectral:
@Mergifyio backport release/v5.4

PR #2729, opened 2026-08-17
fix(dp): size the AF_PACKET ring blocks from the runtime page size

### LINKED ISSUE #2673: dp: AF_PACKET ring fails to allocate on kernels with pages larger than 4K, enforcer captures nothing

kyledong-suse:
## Description

Fix #2673 

Cherry pick the fix back to release/v5.6 branch. For more details about this fix, please refer to the original [PR](https://github.com/neuvector/neuvector/pull/2674) and issue. 

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/dp/ring.c
+++ b/dp/ring.c
@@ -34,6 +34,39 @@ static uint8_t g_tso_packet[MAX_TSO_SIZE];
 #define FRAME_SIZE_V3 (1024 * 2)
 #define BLOCK_SIZE_V3 (1024 * 64)
 
+/*
+ * The kernel requires tp_block_size to be a multiple of PAGE_SIZE. The block
+ * sizes above are multiples of 4096, so they are fine on a 4K-page kernel but
+ * rejected with EINVAL on a 16K or 64K page kernel (arm64 builds with
+ * CONFIG_ARM64_16K_PAGES / CONFIG_ARM64_64K_PAGES).
+ *
+ * Round the block size up to the page size and divide the block count by the
+ * same factor, so the total ring size, and with it the MAP_LOCKED footprint,
+ * is unchanged. Every block size, frame size and block count here is a power
+ * of two, so the factor divides both exactly: block_size stays a multiple of
+ * frame_size, and ring->size stays a power of two, which dp_rx_v1() and
+ * dp_tx_v1() depend on for their "& (ring->size - 1)" wraparound.
+ *
+ * On a 4K-page kernel this is a no-op.
+ */
+static void dp_ring_fit_page_size(unsigned int *block_size, uint *blocks)
+{
+    long page_size = sysconf(_SC_PAGESIZE);
+
+    if (page_size <= 0 || *block_size == 0 || (*block_size % (unsigned int)page_size) == 0) {
+        return;
+    }
+
+    unsigned int aligned = (((*block_size) + (unsigned int)page_size - 1) /
+                            (unsigned int)page_size) * (unsigned int)page_size;
+    unsigned int factor = aligned / *block_size;
+
+    *block_size = aligned;
+    *blocks = (*blocks > factor) ? (*blocks / factor) : 1;
+
+    DEBUG_INIT("page_size=%ld block_size=%u blocks=%u\n", page_size, *block_size, *blocks);
+}
+
 static int dp_ring_bind(int fd, const char *iface)
 {
     struct sockaddr_ll ll;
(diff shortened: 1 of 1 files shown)


williamlin-suse (approved):
LGTM

Issue #2784, opened 2026-09-04
false positive process profile violation events from bench scan


holyspectral:
With NV 5.4.6 enforcer's bench scan is possible to generate false positives due to an issue to determine if short-lived processes used in bench scan are enforcer's or not.  

Jira link: https://jira.suse.com/browse/NVSHAS-10273

PR #2785, opened 2026-09-04
fix: NVSHAS-10273 inherit parent sid when process absent

holyspectral:
## Description

Address #2784

This PR addresses a tricky situation.  NV relies on /proc/<pid>/ to know more information about a process. For short-lived processes, the window that NV can check its process is limited.

In case the enforcer can't find the process anymore, this fix makes short-lived processes to inherit parent sid, as the normal process does, so enforcer has sufficient information to make sure that the process comes from the process tree or not.

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests


kyledong-suse (approved):
LGTM, thanks!

PR #2805, opened 2026-09-08
fix: improve enforcer conn scaling

holyspectral:
## Description

This commit implements the low-hanging fruits for memory utilization from dpMsgConnection().

1. Make sure only one writeCluster running at the same time to prevent go routine leaks.
2. Ensure connsCache can be released through GC.
3. Drop non-critical connection events.

Fix #XXX

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/agent/timer.go
+++ b/agent/timer.go
@@ -11,6 +11,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -39,6 +40,7 @@ var threatMutex sync.Mutex
 var connectionMutex sync.Mutex
 var connsCache []*dp.ConnectionData
 var connsCacheMutex sync.Mutex
+var connsCacheDropped uint64 // low-priority connections dropped while connsCache was full
 var auditLogCache []*share.CLUSAuditLog
 var auditMutex sync.Mutex
 var fqdnIpCache []*share.CLUSFqdnIp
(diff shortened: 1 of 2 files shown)


agent/timer.go
```diff
@@ -385,6 +398,52 @@ func putAuditLogs() {
...
+const connsCacheMax int = connectionMapMax * 8
+const connsCacheHardMax int = connsCacheMax * 2
+
+// connIsHighPriority reports whether a connection must be reported even under
+// load (LEARN so it can be learned into policy; VIOLATE/DENY so violations are
+// not lost). Low-priority (OPEN/ALLOW/CHECK_*) connections are dropped first.
+func connIsHighPriority(conn *dp.Connection) bool {
+	return conn.PolicyAction > C.DP_POLICY_ACTION_CHECK_APP ||
+		conn.PolicyAction == C.DP_POLICY_ACTION_LEARN
+}
```
kyledong-suse, line 431: `connIsHighPriority` only looks at `PolicyAction`. Under load, `OPEN/ALLOW` connections are dropped before `updateConnection` runs.
DNS tunneling is detected later, and if they have `OPEN/ALLOW`, the connection had already dropped. Same reason for connections that already have `Severity / ThreatID` set but `PolicyAction` still `OPEN/ALLOW`.
I wonder if we should keep those connections as high priority even under load.

```
func connIsHighPriority(conn *dp.Connection) bool {
	if conn.PolicyAction > C.DP_POLICY_ACTION_CHECK_APP ||
		conn.PolicyAction == C.DP_POLICY_ACTION_LEARN {
		return true
	}
	if conn.ClientPort != 0 {
		return true
	}
	if conn.Severity > 0 || conn.ThreatID != 0 {
		return true
	}
	return false
}
```
WDYT?

kyledong-suse (commented):
Generally LGTM, thanks!
Just left one question about how we determine high priority connection.

kyledong-suse (approved):
LGTM, thanks!

Issue #1800, opened 2025-02-05
Quick processes can bypass rules


msnominet:
**Environment**
Platform: EKS
Kubernetes/Platform Version(s):  1.31, neuvector 5.4.2

**Describe the bug**
We have noticed that very quick running jobs, do not appear to register to any group, and do not block, even if there is a valid rule. By increasing the sleep time, to roughly five seconds, rules start to be applied. This means that an attacker can violate any rule, if they are quick enough.

**To Reproduce**
1.  Create namespace example
2.  create group that includes namespace example, and a rule that blocks all traffic
for reference here is the CR I use
```
apiVersion: v1
items:
- apiVersion: neuvector.com/v1
  kind: NvClusterSecurityRule
  metadata:
    name: badsite
  spec:
    dlp:
      settings: []
      status: true
    egress:
    - action: deny
      applications:
      - any 
      name: external-egress-bad-site-0
      ports: any
      priority: 0
      selector:
        comment: ""
        criteria: []
        name: external
        original_name: ""
    - action: allow
      applications:
      - any
      name: nv.coredns.kube-system-egress-badsite-1
      ports: tcp/53,udp/53
      priority: 0
      selector:
        comment: ""
        criteria:
        - key: domain
          op: =
          value: kube-system
        - key: service
          op: =
          value: coredns.kube-system
        name: nv.coredns.kube-system
        original_name: ""
    file: []
    ingress: []
    process: []
    target:
      policymode: N/A 
      selector:
        comment: ""
        criteria:
        - key: app
          op: =
          value: badsite
        - key: namespace
          op: =
          value: example
        grp_band_width: 0
        grp_sess_cur: 0
        grp_sess_rate: 0
        mon_metric: false
        name: badsite
        original_name: ""
    waf:
      settings:
      - action: deny
        name: bad-website
      status: true
kind: List
metadata: {}
---
apiVersion: v1
items:
- apiVersion: neuvector.com/v1
  kind: NvWafSecurityRule
  metadata:
    creationTimestamp: null
    name: bad-website
  spec:
    sensor:
      comment: detect bad website
      name: bad-website
      rules:
      - name: checker
        patterns:
        - context: url
          key: pattern
          op: regex
          value: .+badsite.+
kind: List
metadata: {}
```
3) Create job as below
```
---
apiVersion: batch/v1
kind: CronJob
metadata:
  name: badsite-cron
  namespace: example
  labels:
    app: badsite
spec:
  schedule: "* * * * *"
[...]

PR #2862, opened 2026-09-18
fix: process rules not always working.

### LINKED ISSUE #2750: Protect mode intermittently allows an unlisted process, with no denial or event
We run a synthetic check that execs an unlisted process into a pod whose group is in Protect mode and asserts it gets killed. On one cluster it comes back allowed roughly 0.6% of the time: 21 out of ~3,390 checks over five days. Two other clusters running the same configuration haven't produced a single one.

The process is `sleep`, invoked as `kubectl exec <pod> -- sleep 4.815162342`. It appears in no process profile on any group, and every group reads `process_profile.mode=Protect` with the intended baseline (zero-drift on four, basic on one). When it works, kubectl returns 137 and the audit stream carries a Process.Denied. When it doesn't, kubectl returns 0, `sleep` runs the full duration, and there is no event at all: no denial, no violation, nothing in the API to indicate the group stopped enforcing.

What I've ruled out:

- **Enforcer memory pressure.** We had a separate fault where the enforcer crossed its 75%-of-cgroup-limit threshold, logged `CGroupMemoryStatReset`, re-initialised without restarting the container, and fell open while it rebuilt. Raising the limit fixed that and resets are now zero, but this residual continued. Only two resets occurred in the window and neither is near a failure.
- **Policy re-application.** The NvSecurityRules for these groups haven't changed in eight days.
- **Node scoping.** The memory fault was node-scoped, with every Protect group on one node failing together. These span at least two nodes, and different groups fail in the same run.
- **Pod churn or startup races.** The same pod has failed hours apart with no restart in between.

The closest thing I have to a mechanism is a race in process lineage resolution, the enforcer allowing when it can't resolve the parent of a newly exec'd process in time. I can't confirm that from outside the enforcer.

Separately from the race itself: is there any counter or log line that distinguishes "allowed because lineage was unresolved" from "allowed because the profile permits it"? Right now an enforcement gap and a correct allow look identical from outside, which is what made this take five days to characterise and is why it can't be alerted on precisely.

Versions: enforcer 5.6.0, chart 2.11.0, k3s v1.36.1.

holyspectral:
## Description

Fix #2750 

In commit 4bcaf6b20, the commit added an allow-list to define prefixes to add into file monitoring.  This makes sense to reduce the amount of file traversal.  However, this leads to a side-effect where the process-enforcement caller enumExecutables() still sends an empty Dirs to path traverser.

When req.Dirs is empty, it skips all executables except those under the /, e.g., /pause.  This prevents fanotify from being applied to all executables except those in root partition.

When an executable is left without fanotify attached, it will fallback to netlink and it will not always take effect.

This commit fixes the issue.

## Test

### Checklist

- [ ] includes documentation
- [ ] adds unit tests


holyspectral:
sorry this should address #2750 , not #1800

kyledong-suse (approved):
LGTM, thanks!

holyspectral:

PR #2866, opened 2026-09-21
automated federation join failed with controller ha setup

### LINKED ISSUE #2804: automated federation join failed with controller ha setup
**Environment**
- Platform: Kubernetes / Ranche
- Kubernetes/Platform Version(s): Kubernetes 1.34/1.35 / Rancher 2.14.1
- NeuVector Version(s): 5.6.0

**Describe the bug**
We are deploying neuvector on our rancher clusters. The neuvector instance on the rancher server clusters acts as a federation primary, the downstream clusters are joined as managed clusters. The setup for the primary consists of three controller instances for high availability. The federation master service is exposed via a service of type load balancer.

As we did the rollout of neuvector on our test stage, we saw that only a small portion of our clusters are joined. As we looked into the controller logs on the downstream clusters we found an error while executing the api call for joining the federation primary with the error: http 417 - failed to decrypt data: base64 illegal data at byte 8

The error is caused by this line: https://github.com/neuvector/neuvector/blob/v5.6.0/controller/rest/federation.go#L2432

The "fixed join token" is only promoted on the leader of the primary cluster. If the rest api call gets forwarded to a follower, the controller does not use the join token from the configmap and always treat the join ticket as a user generated one (e.g. via the browser in the federation overview). So it tries to decode the uuid because it thinks it's base64 data which fails at the 8th byte which is the dash (-) because a dash is not a valid baes64 character.

As we scaled down the controllers on the federation primary, all clusters joined successfully after we restarted all controllers on the downstream clusters.
The roles are not promoted as labels to the pods so we cannot create a service which only exposes the controller leader.

**To Reproduce**
1. create a join token (e.g. with uuidgen)
2. deploy neuvector with three controllers and a [fedinit config](https://open-docs.neuvector.com/deploying/production/configmap/) including the join token on a rancher server cluster (the instance should be automatically promoted to a federation primary)
3. expose the fed master service with a load balancer
4. deploy a second neuvector on a downstream cluster, also with a fed init config including the join token
5. look into the logs for the error

**Expected behavior**

The automated join via configmap/secret should work with all healthy controller instances and not only with the leader.
The join token should be promoted on all instances. If a join via a user generated token is possible on all controllers a join via a static token should also work.

**Container Logs**

i will upload them later

williamlin-suse:
## Description

Fix https://github.com/neuvector/neuvector/issues/2804

## Test

Steps to reproduce the issue:
1. Create  `neuvector_init configmap` from `fedinitcfg.yaml` & `sysinitcfg.yaml` on k8s cluster-1(as NV primary cluster) so that
```
# in fedinitcfg.yaml
Join_Token: [xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx]   
Primary_Rest_Info: [cluster-1 info]

# in sysinitcfg.yaml
Enable_Tls_Verification: false
```
2. Create  `neuvector_init configmap` from `fedinitcfg.yaml` & `sysinitcfg.yaml` on k8s cluster-2(as NV managed cluster) so that
```
# in fedinitcfg.yaml
Join_Token: [xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx]   
Primary_Rest_Info: [cluster-1 info]
Managed_Rest_Info: [cluster-2 info]

# in sysinitcfg.yaml
Enable_Tls_Verification: false
```

3. Deploy NV with replicas=3 on k8s cluster-1. This is the primary NV cluster.
4. Log in NV cluster-1 UI -> `Assets` -> `System Components` -> `Controllers` & check which controller is the lead
5. On k8s cluster-1 delete the lead NV controller found in step 4 and wait until the new controller pod is created.
6. Deploy NV on k8s cluster-2. This is the managed NV cluster.
7. Check NV cluster-1 UI and notice that NV cluster-2 doesn't join NV cluster-1's federation successfully
There is a controller error log found in cluster-2:
`ERRO|CTL|rest.handlerJoinFedInternal: Request error - error=failed to decrypt data: illegal base64 data at input byte 8`

Steps to verify the fix:
1~6: same as `Steps to reproduce the issue`
7. Check NV cluster-1 UI and notice that NV cluster-2 joins NV cluster-1's federation successfully

### Checklist

- [ ] includes documentation
- [ ] adds unit tests


controller/rest/rest.go
```diff
@@ -2067,6 +2067,9 @@ func startFedRestServer(fedPingInterval uint32) {
 		return
 	} else {
 		_masterClusterIP = m.MasterCluster.RestInfo.Server
+		_fixedJoinToken = m.MasterCluster.FixedJoinToken
+		_allowSameK8sUidRejoin = m.MasterCluster.AllowSameK8sUidRejoin
+		log.WithFields(log.Fields{"_fixedJoinToken": _fixedJoinToken, "_allowSameK8sUidRejoin": _allowSameK8sUidRejoin}).Info("=> test : 1")
```
kyledong-suse: is `=> test : 1` leftover from your test?

williamlin-suse: removed the testing log in fb2ca667dbe2fd3745688f6840ad90a82ccb01b7

kyledong-suse (approved):
LGTM, thanks!

williamlin-suse:

williamlin-suse:

PR #2875, opened 2026-09-22
pathWalker: walk-level test for an empty Dirs request

brokenlander:
## Related Issue

#2750, follow-up to #2862 as discussed there. @holyspectral

## What happens

`TestShouldSkipDir` covers the helper. Nothing exercises `WalkPathTask` itself with an empty `Dirs`, which is the level the bug in #2750 was visible at.

## The test

`TestWalkPathEmptyDirsWalksWholeTree` runs `WalkPathTask` over a temp directory reached through `/proc/self/root` (the task always resolves its root as `/proc/<Pid>/root` + `Path`, so passing the test's own pid makes an ordinary temp dir reachable without root or a container) and checks a nested file is reported. `TestWalkPathDirsLimitsScope` checks a non-empty `Dirs` still prunes.

## Test

Fails with `shouldSkipDir` returning `true` on an empty list (the pre-#2862 walk), passes on main. `go vet`, gofmt clean.

## Notes for reviewer

Test only, no code change. No root needed.


holyspectral (approved):
LGTM.  Thanks!

